PR #3, opened 2026-05-14
Add PAI pack importer

jcfischer:
## Summary
- add a dry-run/apply PAI Pack importer that turns PAI packs into Soma skills
- classify pack files as portable, template, source-doc, or substrate-specific
- preserve PAI source docs as references and write a soma-pack.json manifest
- document the importer and add tests with a Telos-shaped fixture

## Verification
- bun test
- bun run typecheck
- bun run lint
- git diff --check
- real Telos dry-run: 7 portable, 63 template, 3 source-doc files

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,20 @@ bun run soma import algorithm --dry-run
 bun run soma import algorithm --apply
 ```
 
+The PAI Pack importer ports PAI skill packs into Soma skills. Dry-run is the
+default so the file classification can be reviewed before writing:
+
+```bash
+bun run soma import pai-pack --pai-pack-dir /Users/fischer/work/PAI/Packs/Telos
+bun run soma import pai-pack --apply --pai-pack-dir /Users/fischer/work/PAI/Packs/Telos
+```
+
+It preserves PAI pack source docs as references, copies portable workflows and
+tools under `~/.soma/skills/<skill>/`, and records a `soma-pack.json` manifest.
+Existing skills require `--overwrite`; substrate-specific files require
+`--include-substrate-specific`; likely secret files are denied. See
+[docs/pai-pack-importer.md](docs/pai-pack-importer.md).
+
 Soma also exposes a deterministic Algorithm harness. The harness wraps ISA work
 in a one-way phase machine:
 
--- a/docs/pai-pack-importer.md
+++ b/docs/pai-pack-importer.md
@@ -1,0 +1,91 @@
+# PAI Pack Importer
+
+PAI packs are the migration unit for the PAI skill layer. Soma imports them as
+portable Soma skills instead of installing them into a substrate home directly.
+
+## V0 Contract
+
+`soma import pai-pack` reads a PAI pack directory containing:
+
+- `README.md`
+- `INSTALL.md`
+- `VERIFY.md`
+- `src/SKILL.md`
+- optional `src/Workflows/`, `src/Tools/`, and template directories
+
+Dry-run is the default:
+
+```bash
+bun run soma import pai-pack --pai-pack-dir /Users/fischer/work/PAI/Packs/Telos
+```
+
+Apply requires an explicit flag:
+
+```bash
+bun run soma import pai-pack --apply --pai-pack-dir /Users/fischer/work/PAI/Packs/Telos
+```
+
+`--pai-pack-dir` is required. Existing Soma skills are not overwritten unless
+`--overwrite` is supplied. Files classified as `substrate-specific` are refused
+unless `--include-substrate-specific` is supplied. Likely secret files such as
+`.env`, private keys, credentials, tokens, and local settings files are denied
+by default.
+
+## Target Layout
+
+Imported packs land under `~/.soma/skills/<skill-name>/`.
+
+```text
+skills/telos/
+  SKILL.md
+  Workflows/
+  Tools/
+  DashboardTemplate/
+  ReportTemplate/
+  references/
+    PAI-PACK-README.md
+    PAI-PACK-INSTALL.md
+    PAI-PACK-VERIFY.md
+  soma-pack.json
+```
+
+The importer rewrites `SKILL.md` frontmatter to use a substrate-portable,
+lowercase Soma skill name. Source docs are preserved under `references/` so
+Claude-specific installation guidance remains available without becoming the
+Soma install procedure.
+
+## Classification
+
+The import plan classifies each file:
+
+- `portable`: skill entrypoint, workflows, tools, and the generated manifest.
+- `template`: dashboard/report application templates.
+- `source-doc`: original PAI pack README, INSTALL, and VERIFY docs.
+- `substrate-specific`: anything outside the known portable/template/doc paths.
+
+Each planned file also has an `origin`: `source` files include a PAI pack source
+path, while `generated` files carry a generator name for importer metadata. The
+generated `soma-pack.json` follows the exported `PaiPackManifest` contract.
+
+The classification is advisory in V0. It gives later adapters and review tools a
+stable place to decide what should project into Codex, Pi.dev, Claude Code, or a
+Cortex daemon.
+
+When substrate-specific files are explicitly included, Soma stores them outside
+the skill payload under `~/.soma/imports/pai-packs/<skill>/source/`. They are
+kept as migration source material, not projected skill content.
+
+## Boundaries
+
+The importer does not:
+
+- execute the PAI pack installer,
+- modify `~/.claude`,
+- overwrite existing Soma skills without `--overwrite`,
+- copy likely secret files,
+- run template dependency installs,
+- infer personal TELOS content,
+- publish imported skills to substrates by itself.
+
+Projection remains the adapter's job. Import makes the capability available in
+Soma; install/projection makes it available in a substrate.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -8,11 +8,13 @@ import {
   createAlgorithmRun,
   importAlgorithm,
   importPaiIdentity,
+  importPaiPack,
   installSomaForCodex,
   installSomaForPiDev,
   listAlgorithmRunSummaries,
   planAlgorithmImport,
   planPaiImport,
+  planPaiPackImport,
   planSomaForCodexInstall,
   planSomaForPiDevInstall,
   promoteAlgorithmRunMemory,
@@ -42,6 +44,9 @@ import type {
   PaiImportOptions,
   PaiImportPlan,
   PaiImportResult,
+  PaiPackImportOptions,
+  PaiPackImportPlan,
+  PaiPackImportResult,
   SomaInstallOptions,
   SomaInstallPlan,
   SomaInstallResult,
@@ -67,9 +72,9 @@ interface ParsedInstallArgs {
 
 interface ParsedImportArgs {
   command: "import";
-  source: "pai" | "algorithm";
+  source: "pai" | "algorithm" | "pai-pack";
   apply: boolean;
-  options: PaiImportOptions | AlgorithmImportOptions;
+  options: PaiImportOptions | AlgorithmImportOptions | PaiPackImportOptions;
 }
 
 interface ParsedAlgorithmArgs {
@@ -196,17 +201,18 @@ function parseInstallArgs(args: string[]): ParsedInstallArgs {
 function parseImportArgs(args: string[]): ParsedImportArgs {
   const [command, source, ...rest] = args;
 
-  if (command !== "import" || (source !== "pai" && source !== "algorithm")) {
+  if (command !== "import" || (source !== "pai" && source !== "algorithm" && source !== "pai-pack")) {
     throw new Error(
       [
         "Usage:",
         "  soma import pai [--dry-run] [--apply] [--home-dir <dir>] [--claude-home <dir>] [--soma-home <dir>]",
         "  soma import algorithm [--dry-run] [--apply] [--home-dir <dir>] [--pai-algorithm-dir <dir>] [--soma-home <dir>]",
+        "  soma import pai-pack [--dry-run] [--apply] [--home-dir <dir>] --pai-pack-dir <dir> [--soma-home <dir>] [--skill-name <name>] [--overwrite] [--include-substrate-specific]",
       ].join("\n"),
     );
   }
 
-  const options: PaiImportOptions & AlgorithmImportOptions = {};
+  const options: PaiImportOptions & AlgorithmImportOptions & PaiPackImportOptions = {};
   let apply = false;
 
   for (let index = 0; index < rest.length; index += 1) {
@@ -236,6 +242,32 @@ function parseImportArgs(args: string[]): ParsedImportArgs {
         }
         options.paiAlgorithmDir = readOption(rest, index, arg);
         index += 1;
+        break;
+      case "--pai-pack-dir":
+        if (source !== "pai-pack") {
+          throw new Error("--pai-pack-dir is only valid for soma import pai-pack.");
+        }
+        options.paiPackDir = readOption(rest, index, arg);
+        index += 1;
+        break;
+      case "--skill-name":
+        if (source !== "pai-pack") {
+          throw new Error("--skill-name is only valid for soma import pai-pack.");
+        }
+        options.skillName = readOption(rest, index, arg);
+        index += 1;
+        break;
+      case "--overwrite":
+        if (source !== "pai-pack") {
+          throw new Error("--overwrite is only valid for soma import pai-pack.");
+        }
+        options.overwrite = true;
+        break;
+      case "--include-substrate-specific":
+        if (source !== "pai-pack") {
+          throw new Error("--include-substrate-specific is only valid for soma import pai-pack.");
+        }
+        options.includeSubstrateSpecific = true;
         break;
       case "--soma-home":
         options.somaHome = readOption(rest, index, arg);
@@ -951,6 +983,48 @@ function formatAlgorithmImportResult(result: AlgorithmImportResult): string {
   ].join("\n");
 }
 
+function formatPaiPackImportPlan(plan: PaiPackImportPlan): string {
+  const counts = plan.files.reduce<Partial<Record<string, number>>>((acc, file) => {
+    acc[file.classification] = (acc[file.classification] ?? 0) + 1;
+    return acc;
+  }, {});
+
+  return [
+    "Soma PAI Pack import plan",
+    "source: pai-pack",
+    `mode: ${plan.apply ? "apply" : "dry-run"}`,
+    `paiPackDir: ${plan.paiPackDir}`,
+    `somaHome: ${plan.somaHome}`,
+    `skillName: ${plan.skillName}`,
+    `packName: ${plan.packName}`,
+    `description: ${plan.description}`,
+    "",
+    "Classification:",
+    `- portable: ${counts.portable ?? 0}`,
+    `- template: ${counts.template ?? 0}`,
+    `- source-doc: ${counts["source-doc"] ?? 0}`,
+    `- substrate-specific: ${counts["substrate-specific"] ?? 0}`,
+    "",
+    "Files:",
+    ...plan.files.map((file) => {
+      const source = file.origin === "source" ? file.source : `generated:${file.generator}`;
+      return `- [${file.classification}] ${source} -> ${file.target}`;
+    }),
+  ].join("\n");
+}
+
+function formatPaiPackImportResult(result: PaiPackImportResult): string {
+  return [
+    "Soma PAI Pack import applied",
+    `paiPackDir: ${result.paiPackDir}`,
+    `somaHome: ${result.somaHome}`,
+    `skillName: ${result.skillName}`,
+    "",
+    "Files:",
+    ...result.files.map((path) => `- ${path}`),
+  ].join("\n");
+}
+
 function formatAlgorithmRunResult(result: { path: string; run: { id: string; phase: string; effort: string } }): string {
   return [
     "Soma Algorithm run created",
@@ -1262,6 +1336,16 @@ export async function runSomaCli(args: string[]): Promise<string> {
       }
 
       return formatAlgorithmImportResult(await importAlgorithm(options));
+    }
+
+    if (parsed.source === "pai-pack") {
+      const options = parsed.options as PaiPackImportOptions;
+
+      if (!parsed.apply) {
+        return formatPaiPackImportPlan(await planPaiPackImport(options));
+      }
+
+      return formatPaiPackImportResult(await importPaiPack(options));
     }
 
     const options = parsed.options as PaiImportOptions;
--- a/src/index.ts
+++ b/src/index.ts
@@ -56,6 +56,12 @@ export type {
   PaiImportOptions,
   PaiImportPlan,
   PaiImportResult,
+  PaiPackImportFile,
+  PaiPackImportOptions,
+  PaiPackImportPlan,
+  PaiPackImportResult,
+  PaiPackManifest,
+  PaiPackManifestFile,
   WrittenContextBundle,
 } from "./types";
 
@@ -118,6 +124,7 @@ export {
 export { appendSomaMemoryEvent, searchSomaMemory, somaMemoryEventsPath } from "./memory";
 export { promoteAlgorithmRunMemory } from "./memory-promotion";
 export { importPaiIdentity, planPaiImport } from "./pai-importer";
+export { importPaiPack, planPaiPackImport } from "./pai-pack-importer";
 export { checkSomaPolicy, checkSomaPolicyBatch } from "./policy-audit";
 export { evaluateSomaPolicy } from "./policy";
 export { bootstrapSomaHome, loadSomaHome } from "./soma-home";
--- a/src/pai-pack-importer.ts
+++ b/src/pai-pack-importer.ts
@@ -1,0 +1,573 @@
+import { access, copyFile, lstat, mkdir, readdir, readFile, realpath, rename, rm, writeFile } from "node:fs/promises";
+import { homedir } from "node:os";
+import { dirname, isAbsolute, join, relative, resolve, sep } from "node:path";
+import { routePaiPackSourceFile, type PaiPackRenderMode } from "./pai-pack-routing";
+import type {
+  PaiPackGeneratedImportFile,
+  PaiPackImportFile,
+  PaiPackImportOptions,
+  PaiPackImportPlan,
+  PaiPackImportResult,
+  PaiPackManifest,
+  PaiPackSourceImportFile,
+} from "./types";
+
+interface PackMetadata {
+  name: string;
+  description: string;
+}
+
+const RESERVED_SKILL_NAMES = new Set(["soma", "the-algorithm"]);
+const REQUIRED_PACK_FILES = ["README.md", "INSTALL.md", "VERIFY.md", "src/SKILL.md"];
+const NORMALIZED_SKILL_NAME_PATTERN = /^[a-z0-9]+(?:-[a-z0-9]+)*$/;
+
+const SECRET_FILE_PATTERNS = [
+  /(^|\/)\.env$/,
+  /(^|\/)\.env\.(?!example$)[^/]+$/,
+  /(^|\/)id_[a-z0-9_-]+$/,
+  /(^|\/)[^/]+\.(pem|key|p12|pfx)$/,
+  /(^|\/)(credentials|secrets|tokens?)\.(json|yaml|yml|toml|ini|txt)$/,
+  /(^|\/)(settings|config)\.json$/,
+  /(^|\/)settings\.local\.json$/,
+  /(^|\/)local\.settings\.json$/,
+  /(^|\/)\.npmrc$/,
+  /(^|\/)\.pypirc$/,
+  /(^|\/)\.netrc$/,
+  /(^|\/)\.gem\/credentials$/,
+  /(^|\/)\.cargo\/credentials(?:\.toml)?$/,
+  /(^|\/)(\.aws|aws)\/credentials$/,
+  /(^|\/)(\.kube|kube)\/config$/,
+  /(^|\/)pip\.conf$/,
+];
+
+type RoutedPaiPackImportFile =
+  | (PaiPackSourceImportFile & { renderMode: Extract<PaiPackRenderMode, "copy" | "skill"> })
+  | (PaiPackGeneratedImportFile & { renderMode: Extract<PaiPackRenderMode, "manifest" | "archive-manifest"> });
+
+interface InternalPaiPackImportPlan extends PaiPackImportPlan {
+  routedFiles: RoutedPaiPackImportFile[];
+}
+
+function isRoutedSourceFile(file: RoutedPaiPackImportFile): file is Extract<RoutedPaiPackImportFile, { origin: "source" }> {
+  return file.origin === "source";
+}
+
+function resolvePackHomes(options: PaiPackImportOptions = {}): { paiPackDir: string; somaHome: string } {
+  if (!options.paiPackDir) {
+    throw new Error("soma import pai-pack requires --pai-pack-dir <dir>.");
+  }
+
+  const home = resolve(options.homeDir ?? homedir());
+
+  return {
+    paiPackDir: resolve(options.paiPackDir),
+    somaHome: resolve(options.somaHome ?? join(home, ".soma")),
+  };
+}
+
+function slugifySkillName(value: string): string {
+  return value
+    .trim()
+    .replace(/([a-z0-9])([A-Z])/g, "$1-$2")
+    .toLowerCase()
+    .replace(/[^a-z0-9]+/g, "-")
+    .replace(/^-+|-+$/g, "");
+}
+
+function parseFrontmatter(content: string): Partial<Record<string, string>> {
+  if (!content.startsWith("---\n")) {
+    return {};
+  }
+
+  const end = content.indexOf("\n---", 4);
+  if (end === -1) {
+    return {};
+  }
+
+  const frontmatter = content.slice(4, end);
+  const fields: Partial<Record<string, string>> = {};
+
+  for (const line of frontmatter.split("\n")) {
+    const separator = line.indexOf(":");
+    if (separator === -1) continue;
+    const key = line.slice(0, separator).trim();
+    const value = line.slice(separator + 1).trim();
+    if (key && value && !value.startsWith("[")) {
+      fields[key] = value.replace(/^["']|["']$/g, "");
+    }
+  }
+
+  return fields;
+}
+
+async function readPackMetadata(paiPackDir: string): Promise<PackMetadata> {
+  const readme = await readFile(join(paiPackDir, "README.md"), "utf8");
+  const fields = parseFrontmatter(readme);
+  const heading = /^#\s+(.+)$/m.exec(readme)?.[1]?.trim();
+
+  return {
+    name: fields.name ?? heading ?? "pai-pack",
+    description: fields.description ?? "",
+  };
+}
+
+async function collectFiles(root: string): Promise<string[]> {
+  const files: string[] = [];
+  const realRoot = await realpath(root);
+
+  async function visit(dir: string): Promise<void> {
+    const entries = await readdir(dir, { withFileTypes: true });
+
+    for (const entry of entries) {
+      const fullPath = join(dir, entry.name);
+      if (entry.name.includes("\\")) {
+        throw new Error(`PAI pack import refused ambiguous path separator: ${relative(root, fullPath).split(sep).join("/")}`);
+      }
+
+      if (entry.isSymbolicLink()) {
+        throw new Error(`PAI pack import refused symlink path: ${relative(root, fullPath).split(sep).join("/")}`);
+      }
+
+      if (entry.isDirectory()) {
+        if (entry.name === "node_modules" || entry.name === ".next" || entry.name === "dist") {
+          continue;
+        }
+        if (entry.name === ".git" || entry.name === ".hg" || entry.name === ".svn") {
+          throw new Error(`PAI pack import refused VCS metadata directory: ${relative(root, fullPath).split(sep).join("/")}`);
+        }
+        await visit(fullPath);
+      } else if (entry.isFile()) {
+        const realFile = await realpath(fullPath);
+        if (!isWithinPath(realRoot, realFile)) {
+          throw new Error(`PAI pack import refused path outside pack root: ${relative(root, fullPath).split(sep).join("/")}`);
+        }
+        const path = relative(root, fullPath).split(sep).join("/");
+        if (!path.startsWith("..") && !path.includes("../")) {
+          files.push(path);
+        }
+      }
+    }
+  }
+
+  await visit(root);
+  return files.sort();
+}
+
+async function pathExists(path: string): Promise<boolean> {
+  return access(path)
+    .then(() => true)
+    .catch((error: unknown) => {
+      if (error instanceof Error && "code" in error && error.code === "ENOENT") return false;
+      throw error;
+    });
+}
+
+function isLikelySecretPath(path: string): boolean {
+  const normalizedPath = path.replace(/\\/g, "/");
+  const segments = normalizedPath.split("/").map((segment) => segment.toLowerCase());
+  if (segments.some((segment) => segment === ".git" || segment === ".hg" || segment === ".svn")) {
+    return true;
+  }
+  if (segments.some((segment) => /^(secrets?|credentials?|tokens?)$/.test(segment))) {
+    return true;
+  }
+
+  const basename = segments.at(-1) ?? "";
+  if (/\b(token|secret|credential|key)s?\b/i.test(basename.replace(/[_\-.]+/g, " "))) {
+    return true;
+  }
+
+  return SECRET_FILE_PATTERNS.some((pattern) => pattern.test(normalizedPath.toLowerCase()));
+}
+
+function assertRequiredPackFiles(sourceFiles: string[]): void {
+  const sourceFileSet = new Set(sourceFiles);
+  const missing = REQUIRED_PACK_FILES.filter((path) => !sourceFileSet.has(path));
+  if (missing.length > 0) {
+    throw new Error(`PAI pack import requires V0 pack file(s): ${missing.join(", ")}`);
+  }
+}
+
+function assertSafePackPaths(sourceFiles: string[]): void {
+  const unsafePaths = sourceFiles.filter((path) => path.replace(/\\/g, "/").split("/").includes(".."));
+  if (unsafePaths.length > 0) {
+    throw new Error(`PAI pack import refused unsafe path segment(s): ${unsafePaths.join(", ")}`);
+  }
+}
+
+function assertUniqueTargets(files: Pick<PaiPackSourceImportFile | PaiPackGeneratedImportFile, "target">[]): void {
+  const seen = new Map<string, string>();
+  const duplicates: string[] = [];
+
+  for (const file of files) {
+    const key = file.target.toLowerCase();
+    const previous = seen.get(key);
+    if (previous) {
+      duplicates.push(`${previous} <-> ${file.target}`);
+    } else {
+      seen.set(key, file.target);
+    }
+  }
+
+  if (duplicates.length > 0) {
+    throw new Error(`PAI pack import refused duplicate target path(s): ${duplicates.join(", ")}`);
+  }
+}
+
+function isWithinPath(root: string, target: string): boolean {
+  const path = relative(root, target);
+  return path === "" || (!path.startsWith(`..${sep}`) && path !== ".." && !isAbsolute(path));
+}
+
+async function nearestExistingAncestor(path: string): Promise<string> {
+  let candidate = path;
+  while (!(await pathExists(candidate))) {
+    const parent = dirname(candidate);
+    if (parent === candidate) {
+      return candidate;
+    }
+    candidate = parent;
+  }
+  return candidate;
+}
+
+async function assertParentWithinRealRoot(root: string, target: string): Promise<void> {
+  const realRoot = await realpath(root);
+  const ancestor = await nearestExistingAncestor(dirname(target));
+  const realAncestor = await realpath(ancestor);
+  if (!isWithinPath(realRoot, realAncestor)) {
+    throw new Error(`PAI pack import refused target parent outside Soma home: ${target}`);
+  }
+}
+
+async function resolveSafeSourceFile(realPackRoot: string, source: string): Promise<string> {
+  const sourceStat = await lstat(source);
+  if (sourceStat.isSymbolicLink()) {
+    throw new Error(`PAI pack import refused symlink source during apply: ${source}`);
+  }
+
+  const realSource = await realpath(source);
+  if (!isWithinPath(realPackRoot, realSource)) {
+    throw new Error(`PAI pack import refused source outside pack root during apply: ${source}`);
+  }
+
+  return realSource;
+}
+
+async function mapWithConcurrency<T, R>(items: T[], concurrency: number, worker: (item: T) => Promise<R>): Promise<R[]> {
+  const results: R[] = [];
+  let nextIndex = 0;
+
+  async function runWorker(): Promise<void> {
+    while (nextIndex < items.length) {
+      const index = nextIndex;
+      nextIndex += 1;
+      results[index] = await worker(items[index]);
+    }
+  }
+
+  const workerCount = Math.min(concurrency, items.length);
+  await Promise.all(Array.from({ length: workerCount }, runWorker));
+  return results;
+}
+
+function renderManifest(plan: PaiPackImportPlan): string {
+  const skillRoot = join(plan.somaHome, "skills", plan.skillName);
+  const skillFiles = plan.files.filter((file) => isWithinPath(skillRoot, file.target));
+  return renderManifestForRoot(plan, skillRoot, skillFiles);
+}
+
+function renderArchiveManifest(plan: PaiPackImportPlan): string {
+  const archiveRoot = join(plan.somaHome, "imports", "pai-packs", plan.skillName);
+  const archiveFiles = plan.files.filter((file) => isWithinPath(archiveRoot, file.target) && file.origin === "source");
+  return renderManifestForRoot(plan, archiveRoot, archiveFiles);
+}
+
+function renderManifestForRoot(plan: PaiPackImportPlan, root: string, files: PaiPackImportFile[]): string {
+  const manifest: PaiPackManifest = {
+    schema: "soma.pai-pack-import.v1",
+    skillName: plan.skillName,
+    packName: plan.packName,
+    description: plan.description,
+    files: files.map((file) => {
+      const manifestFile = {
+        target: relative(root, file.target).split(sep).join("/"),
+        classification: file.classification,
+        origin: file.origin,
+      };
+
+      if (file.origin === "source") {
+        return {
+          ...manifestFile,
+          origin: "source",
+          source: relative(plan.paiPackDir, file.source).split(sep).join("/"),
+        };
+      }
+
+      return {
+        ...manifestFile,
+        origin: "generated",
+        generator: file.generator,
+      };
+    }),
+  };
+
+  return `${JSON.stringify(manifest, null, 2)}\n`;
+}
+
+function rewriteSkillFrontmatter(content: string, skillName: string, description: string): string {
+  if (!content.startsWith("---\n")) {
+    return content;
+  }
+
+  const end = content.indexOf("\n---", 4);
+  if (end === -1) {
+    return content;
+  }
+
+  const body = content.slice(end + "\n---".length);
+  return ["---", `name: ${JSON.stringify(skillName)}`, `description: ${JSON.stringify(description || `Imported PAI pack: ${skillName}`)}`, "metadata:", "  source: pai-pack", "---", body.trimStart()].join("\n");
+}
+
+async function buildPaiPackImportPlan(options: PaiPackImportOptions = {}): Promise<InternalPaiPackImportPlan> {
+  const homes = resolvePackHomes(options);
+  const sourceFiles = await collectFiles(homes.paiPackDir);
+  assertSafePackPaths(sourceFiles);
+  assertRequiredPackFiles(sourceFiles);
+
+  const metadata = await readPackMetadata(homes.paiPackDir);
+  const skillName = options.skillName ? slugifySkillName(options.skillName) : slugifySkillName(metadata.name);
+  if (!skillName) {
+    throw new Error("soma import pai-pack requires a non-empty skill name after normalization.");
+  }
+  if (!NORMALIZED_SKILL_NAME_PATTERN.test(skillName)) {
+    throw new Error(`soma import pai-pack produced invalid normalized skill name '${skillName}'.`);
+  }
+  if (RESERVED_SKILL_NAMES.has(skillName)) {
+    throw new Error(`soma import pai-pack cannot overwrite reserved Soma skill '${skillName}'.`);
+  }
+  if (!options.overwrite) {
+    await access(join(homes.somaHome, "skills", skillName))
+      .then(() => {
+        throw new Error(`Soma skill '${skillName}' already exists. Re-run with --overwrite to replace it.`);
+      })
+      .catch((error: unknown) => {
+        if (error instanceof Error && "code" in error && error.code === "ENOENT") return;
+        throw error;
+      });
+    await access(join(homes.somaHome, "imports", "pai-packs", skillName))
+      .then(() => {
+        throw new Error(`Soma PAI pack archive '${skillName}' already exists. Re-run with --overwrite to replace it.`);
+      })
+      .catch((error: unknown) => {
+        if (error instanceof Error && "code" in error && error.code === "ENOENT") return;
+        throw error;
+      });
+  }
+
+  const secretFiles = sourceFiles.filter(isLikelySecretPath);
+  if (secretFiles.length > 0) {
+    throw new Error(`PAI pack import refused likely secret file(s): ${secretFiles.join(", ")}`);
+  }
+
+  const routes = sourceFiles.map((path) => ({ path, route: routePaiPackSourceFile(path) }));
+  const substrateSpecific = routes.filter(({ route }) => route.classification === "substrate-specific");
+  if (substrateSpecific.length > 0 && !options.includeSubstrateSpecific) {
+    throw new Error(
+      `PAI pack import refused substrate-specific file(s) without --include-substrate-specific: ${substrateSpecific.map(({ path }) => path).join(", ")}`,
+    );
+  }
+
+  const routedFiles: RoutedPaiPackImportFile[] = routes.map(({ path, route }) => ({
+    source: join(homes.paiPackDir, path),
+    target: join(homes.somaHome, route.root === "skill" ? "skills" : "imports/pai-packs", skillName, route.relativePath),
+    classification: route.classification,
+    renderMode: route.renderMode,
+    origin: "source",
+  }));
+
+  routedFiles.push({
+    target: join(homes.somaHome, `skills/${skillName}/soma-pack.json`),
+    classification: "portable",
+    renderMode: "manifest",
+    origin: "generated",
+    generator: "pai-pack-importer",
+  });
+  if (substrateSpecific.length > 0) {
+    routedFiles.push({
+      target: join(homes.somaHome, `imports/pai-packs/${skillName}/soma-pack-archive.json`),
+      classification: "substrate-specific",
+      renderMode: "archive-manifest",
+      origin: "generated",
+      generator: "pai-pack-importer",
+    });
+  }
+
+  const escapedTargets = routedFiles.filter((file) => !isWithinPath(homes.somaHome, file.target));
+  if (escapedTargets.length > 0) {
+    throw new Error(`PAI pack import refused target path outside Soma home: ${escapedTargets.map((file) => file.target).join(", ")}`);
+  }
+  assertUniqueTargets(routedFiles);
+
+  const files = routedFiles.map(({ renderMode: _renderMode, ...file }) => file);
+
+  return {
+    apply: false,
+    paiPackDir: homes.paiPackDir,
+    somaHome: homes.somaHome,
+    skillName,
+    packName: metadata.name,
+    description: metadata.description,
+    files,
+    routedFiles,
+  };
+}
+
+export async function planPaiPackImport(options: PaiPackImportOptions = {}): Promise<PaiPackImportPlan> {
+  const { routedFiles: _routedFiles, ...plan } = await buildPaiPackImportPlan(options);
+  return plan;
+}
+
+async function stagePaiPackFiles(plan: InternalPaiPackImportPlan, stageRoot: string): Promise<void> {
+  const realPackRoot = await realpath(plan.paiPackDir);
+
+  await mapWithConcurrency(plan.routedFiles, 8, async (file) => {
+    const target = file.target;
+    const stagedTarget = join(stageRoot, relative(plan.somaHome, target));
+    if (!isWithinPath(stageRoot, stagedTarget)) {
+      throw new Error(`PAI pack import refused staged target path outside stage root: ${target}`);
+    }
+
+    await mkdir(dirname(stagedTarget), { recursive: true });
+
+    if (file.renderMode === "manifest") {
+      await writeFile(stagedTarget, renderManifest(plan), "utf8");
+    } else if (file.renderMode === "archive-manifest") {
+      await writeFile(stagedTarget, renderArchiveManifest(plan), "utf8");
+    } else if (file.renderMode === "skill") {
+      const source = await resolveSafeSourceFile(realPackRoot, file.source);
+      const content = await readFile(source, "utf8");
+      await writeFile(stagedTarget, `${rewriteSkillFrontmatter(content, plan.skillName, plan.description).trimEnd()}\n`, "utf8");
+    } else {
+      if (!isRoutedSourceFile(file)) {
+        throw new Error(`PAI pack import cannot copy generated file: ${target}`);
+      }
+      const source = await resolveSafeSourceFile(realPackRoot, file.source);
+      await copyFile(source, stagedTarget);
+    }
+  });
+}
+
+interface PromotionContext {
+  plan: InternalPaiPackImportPlan;
+  stageRoot: string;
+  backupRoot: string;
+  skillRoot: string;
+  sourceArchiveRoot: string;
+  overwrite: boolean;
+  cleanupRoots: Set<string>;
+}
+
+async function promoteStagedImportWithRollback(context: PromotionContext): Promise<void> {
+  const { plan, stageRoot, backupRoot, skillRoot, sourceArchiveRoot, overwrite, cleanupRoots } = context;
+  const backupSkillRoot = join(backupRoot, "skills", plan.skillName);
+  const backupArchiveRoot = join(backupRoot, "imports", "pai-packs", plan.skillName);
+  const hadSkillRoot = await pathExists(skillRoot);
+  const hadArchiveRoot = await pathExists(sourceArchiveRoot);
+
+  const stagedSkillRoot = join(stageRoot, "skills", plan.skillName);
+  const stagedArchiveRoot = join(stageRoot, "imports", "pai-packs", plan.skillName);
+  let promotedSkillRoot = false;
+  let promotedArchiveRoot = false;
+
+  try {
+    if (overwrite) {
+      if (hadSkillRoot) {
+        await mkdir(dirname(backupSkillRoot), { recursive: true });
+        await rename(skillRoot, backupSkillRoot);
+      }
+      if (hadArchiveRoot) {
+        await mkdir(dirname(backupArchiveRoot), { recursive: true });
+        await rename(sourceArchiveRoot, backupArchiveRoot);
+      }
+    }
+
+    if (await pathExists(stagedSkillRoot)) {
+      await mkdir(dirname(skillRoot), { recursive: true });
+      await rename(stagedSkillRoot, skillRoot);
+      promotedSkillRoot = true;
+    }
+
+    if (await pathExists(stagedArchiveRoot)) {
+      await mkdir(dirname(sourceArchiveRoot), { recursive: true });
+      await rename(stagedArchiveRoot, sourceArchiveRoot);
+      promotedArchiveRoot = true;
+    }
+  } catch (error) {
+    try {
+      if (promotedSkillRoot) {
+        await rm(skillRoot, { recursive: true, force: true });
+      }
+      if (promotedArchiveRoot) {
+        await rm(sourceArchiveRoot, { recursive: true, force: true });
+      }
+      if (hadSkillRoot && (await pathExists(backupSkillRoot))) {
+        await mkdir(dirname(skillRoot), { recursive: true });
+        await rename(backupSkillRoot, skillRoot);
+      }
+      if (hadArchiveRoot && (await pathExists(backupArchiveRoot))) {
+        await mkdir(dirname(sourceArchiveRoot), { recursive: true });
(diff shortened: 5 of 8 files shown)


jcfischer (commented):

6 finding(s): 5 important, 1 suggestion.


### CodeQuality
The importer is mostly straightforward, but it does not faithfully copy arbitrary pack files. Two edge cases can produce incorrect output paths or corrupted copied content.

- **[important]** `src/pai-pack-importer.ts:189` — **Binary files are copied as UTF-8**
  Every imported file is read as UTF-8 text and rewritten with normalized trailing newlines. Pack templates can contain images or other binary assets, which this will corrupt during apply.
  
  Suggestion:

  ```
  Copy non-SKILL files as buffers, and only text-transform SKILL.md.
  ```

- **[important]** `src/pai-pack-importer.ts:138` — **Empty skill names are accepted**
  slugifySkillName can return an empty string for invalid metadata or --skill-name input. That writes into skills/ directly instead of a named skill directory.
  
  Suggestion:

  ```
  Reject empty slugified skill names with a clear error before planning.
  ```

### Security
The importer adds a new trust boundary: local PAI pack content can write into the Soma skill store. Two issues meet the security threshold: unguarded skill overwrites and wholesale copying of likely-secret substrate files.

- **[important]** `src/pai-pack-importer.ts:185` — **Pack import can overwrite trusted skills**
  The importer derives the destination skill from pack metadata or CLI input and writes files with no collision or reserved-name guard. A malicious pack named like an existing trusted skill can replace its SKILL.md, workflows, or tools under ~/.soma/skills.
  
  Suggestion:

  ```
  Reject existing skill directories unless an explicit --overwrite flag is supplied, and reserve built-in skill names.
  ```

- **[important]** `src/pai-pack-importer.ts:110` — **Importer copies likely secrets**
  Unknown files are classified as substrate-specific but still copied into references/source. A pack containing .env, keys, tokens, or substrate config would propagate secrets into the portable Soma skill store.
  
  Suggestion:

  ```
  Deny known secret filenames/patterns by default and require explicit opt-in for substrate-specific files.
  ```

### Architecture
The importer mostly preserves the intended boundary between import planning, Soma skill materialization, and later substrate projection. The main architecture issue is that the generic pack importer bakes in a specific Telos workspace default, coupling the reusable module to one pack.

- **[important]** `src/pai-pack-importer.ts:15` — **Generic importer defaults to Telos**
  A reusable PAI pack importer should not encode one personal pack path as its default source. That couples the core module to Telos and makes future pack imports inherit a misleading, non-generic behavior.
  
  Suggestion:

  ```
  Require `paiPackDir`, or move any Telos convenience default to CLI/docs.
  ```

### Performance
The importer runs as an explicit CLI migration path, not a hot request path, and the changed code does not introduce a concrete performance regression. File traversal and copy work are bounded by the pack directory and are acceptable for this apply/dry-run workflow.

_No findings._

### Maintainability
The importer is small and readable overall, with most branching kept shallow. The main maintainability cost is duplicated path-routing knowledge across classification, targeting, and write behavior.

- **[suggestion]** `src/pai-pack-importer.ts:84` — **Consolidate pack path routing**
  `classifySourceFile`, `targetPathFor`, and the write loop each encode related path rules, so adding a new source-doc or portable path requires coordinated edits across multiple branches.
  
  Suggestion:

  ```
  Create one helper that returns `{ classification, target, renderMode }` for each source path.
  ```


---

jcfischer (commented):

### CodeQuality
The importer is generally clear and covered by focused tests, but one path-handling edge case can break classification on non-POSIX platforms. Findings are limited to CodeQuality concerns around correctness and edge cases.

- **[important]** `src/pai-pack-importer.ts:96` — **Normalize collected relative paths**
  `relative()` returns platform-native separators, but routing and secret-pattern checks compare against forward-slash paths. On Windows, expected files like `src\SKILL.md` would be misclassified instead of imported as portable/template/source docs.
  
  Suggestion:

  ```
  Normalize collected paths with `split(sep).join("/")` before routing.
  ```

### Security
The importer adds explicit dry-run behavior, collision protection, reserved-name checks, substrate-specific gating, and path-based secret denial. I found no concrete security issue in the changed surfaces from the provided diff.

_No findings._

### Architecture
The importer fits the existing import-module pattern, but the import plan does not carry enough operation metadata for the apply phase. That makes future routing changes depend on target path conventions instead of the plan contract.

- **[important]** `src/pai-pack-importer.ts:211` — **Plan drops render operation metadata**
  `routeSourceFile` decides `renderMode`, but `plan.files` exposes only source, target, and classification; `importPaiPack` later re-infers behavior from target suffixes. This couples write semantics to path names and makes future pack layouts or generated files harder to add without hidden conventions.
  
  Suggestion:

  ```
  Add an internal plan field for render mode, or keep routed files with operation metadata through apply.
  ```

### Performance
The importer does filesystem traversal and file copying in a CLI path, with work proportional to pack size and no obvious hot-path impact. I did not find concrete performance regressions such as N+1 I/O, sync work in async paths, or unbounded retained memory.

_No findings._

### Maintainability
The importer is mostly readable, with classification and apply paths kept compact. The main maintainability cost is repeated command-specific CLI parsing and a few bundled test cases that will get harder to extend as importer options grow.

- **[suggestion]** `src/cli.ts:243` — **Repeated pai-pack option guards**
  Each pai-pack-only option repeats the same source check and error shape, so adding more importer flags will copy this branch pattern again.
  
  Suggestion:

  ```
  Extract a small `assertPaiPackOption(source, arg)` helper or parse pai-pack options in a command-specific function.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:80` — **Split bundled rejection scenarios**
  This test covers invalid names, reserved names, collisions, secrets, and substrate-specific files in one flow with shared mutable fixture state. A failure later in the test will obscure which policy changed.
  
  Suggestion:

  ```
  Split into focused tests by rejection category.
  ```


---

jcfischer (commented):

6 finding(s): 2 important, 4 suggestion.


### CodeQuality
The importer is generally clear and covered by focused tests, but overwrite behavior leaves stale files behind. That can make an applied import differ from the current pack contents after re-importing an existing skill.

- **[important]** `src/pai-pack-importer.ts:238` — **Overwrite leaves stale files**
  When `overwrite` is enabled, existing files under the skill directory are not removed before writing the new pack. Re-importing a changed pack can leave deleted or renamed files in place, so the resulting skill no longer matches the source pack.
  
  Suggestion:

  ```
  Before writing with `overwrite`, remove `skills/<skillName>` or explicitly prune files not present in the new plan.
  ```

### Security
The importer has reasonable guardrails around dry-run defaults, reserved skill names, and path-based refusal, but its secret-file denylist misses common credential-bearing config files. That creates a concrete leak path when a PAI pack contains registry or tool auth files that are copied into Soma skills.

- **[important]** `src/pai-pack-importer.ts:13` — **Secret denylist misses auth config files**
  Files such as `.npmrc`, `.pypirc`, and `.netrc` commonly contain tokens but are not matched by `SECRET_FILE_PATTERNS`, so a pack can import them into `~/.soma/skills/<skill>/`. This violates the documented invariant that likely secret files are denied by default.
  
  Suggestion:

  ```
  Add common credential config filenames such as `.npmrc`, `.pypirc`, and `.netrc` to `SECRET_FILE_PATTERNS`.
  ```

### Architecture
The importer keeps dependencies one-way and the new code mostly lives at the right layer. The only architectural concern is that an internal rendering strategy becomes part of the exported API, which makes later importer refactors harder.

- **[suggestion]** `src/types.ts:303` — **Internal render mode leaks publicly**
  PaiPackImportFile is exported from src/index.ts, so renderMode becomes part of the public importer contract even though it describes how this implementation writes files. That couples future callers to copy/skill/manifest mechanics rather than the stable import plan shape.
  
  Suggestion:

  ```
  Keep renderMode module-private or expose a domain field callers need, such as classification plus target.
  ```

### Performance
The importer is a one-shot CLI path, so there are no hot-path or render-loop concerns. One concrete file-I/O latency issue appears in apply mode: copies are serialized across every imported file.

- **[suggestion]** `src/pai-pack-importer.ts:247` — **Serial file copies slow large imports**
  Apply mode awaits mkdir, readFile, and writeFile for each file before starting the next one, so template-heavy packs pay cumulative filesystem latency. The documented Telos run already has 73 files, and larger dashboard templates will scale poorly.
  
  Suggestion:

  ```
  Copy files with a bounded Promise pool, reusing mkdir per target directory.
  ```

### Maintainability
The importer is readable overall, but the CLI parser and file routing policy are starting to accumulate command-specific branching and path-string coupling. These are maintainability suggestions, not merge blockers.

- **[suggestion]** `src/cli.ts:201` — **Extract pai-pack argument parsing**
  `parseImportArgs` now mixes shared import flags with four pai-pack-only branches and repeated source guards, increasing future-change cost as import sources grow.
  
  Suggestion:

  ```
  Move pai-pack-specific option handling into a small `parsePaiPackImportOption` or source-specific parser.
  ```

- **[suggestion]** `src/pai-pack-importer.ts:122` — **Centralize source path routing**
  `routeSourceFile` couples classification and target rewriting through repeated path literals plus `path.slice("src/".length)`, making new template/workflow routes easy to update inconsistently.
  
  Suggestion:

  ```
  Define route prefixes once and use a helper like `stripSrcPrefix(path)` for target construction.
  ```


---

jcfischer (commented):

5 finding(s): 2 important, 3 suggestion.


### CodeQuality
The importer is mostly straightforward and well-covered for the fixture path. I found one real edge-case correctness issue in template routing.

- **[important]** `src/pai-pack-importer.ts:142` — **Template routing can corrupt targets**
  The condition matches any path containing `DashboardTemplate/` or `ReportTemplate/`, but the target always strips the first four characters as if the path starts with `src/`. A top-level or nested template directory would be classified as a template and written to a truncated target path.
  
  Suggestion:

  ```
  Require `path.startsWith("src/DashboardTemplate/") || path.startsWith("src/ReportTemplate/")` before slicing.
  ```

### Security
The importer adds file-copying from an untrusted pack boundary and includes basic guards for skill names, overwrite, substrate-specific files, and likely secret filenames. I found no concrete security issue in the diff that meets this lens threshold.

_No findings._

### Architecture
The importer mostly keeps the PAI-to-Soma boundary explicit: planning, apply, CLI wiring, and public exports follow the existing importer shape. One routing choice weakens that boundary by allowing substrate-specific source trees to become part of the imported skill payload.

- **[important]** `src/pai-pack-importer.ts:148` — **Substrate files enter skill payload**
  When `--include-substrate-specific` is used, substrate-specific files are copied under `skills/<skill>/references/source`, making them part of the Soma skill tree rather than a separate migration archive. That blurs the documented boundary that import makes portable Soma capability available while projection remains adapter-owned.
  
  Suggestion:

  ```
  Keep substrate-specific files outside the skill payload, or store only manifest references to them.
  ```

### Performance
The importer does no hot-path work and avoids sync I/O. The main performance cost is apply mode serializing every filesystem copy, which scales linearly with pack size and avoidably increases import latency.

- **[suggestion]** `src/pai-pack-importer.ts:275` — **File copies run serially**
  Apply mode awaits mkdir/read/write for each routed file one at a time, so a pack with many template assets pays cumulative filesystem latency. The described Telos pack already has 73 files, making this a concrete import-time slowdown.
  
  Suggestion:

  ```
  Copy independent files with bounded concurrency, e.g. process plan.routedFiles through a small worker pool.
  ```

### Maintainability
The importer is reasonably structured, with most file-routing and rendering behavior isolated in small helpers. The main maintainability drag is clustered source-specific CLI parsing and an overpacked rejection test that will be harder to update as importer policy grows.

- **[suggestion]** `src/cli.ts:243` — **Repeated pai-pack option guards**
  Each new pai-pack-only flag repeats the same source check and error shape, adding branch noise to an already broad parser. Future import sources will likely copy this pattern again.
  
  Suggestion:

  ```
  Extract a small assertImportSource(source, flag, expected) helper or dispatch source-specific option parsing.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:88` — **Rejection scenarios share one test**
  This test combines invalid names, reserved names, collisions, secret files, and substrate-specific files with shared mutable fixture state. A future policy change or failure will be harder to localize.
  
  Suggestion:

  ```
  Split into focused tests or a small table for independent rejection cases.
  ```


---

jcfischer (commented):

### CodeQuality
The importer is generally clear and covered by focused tests, but one overwrite path leaves old imported source material behind. That can make subsequent imports report success while retaining files from a prior pack shape.

- **[important]** `src/pai-pack-importer.ts:278` — **Overwrite leaves stale source files**
  `--overwrite` only removes the skill directory, but substrate-specific files are written under `imports/pai-packs/<skill>/source`. Re-importing with `--include-substrate-specific` can leave deleted or renamed source files from the previous import in place.
  
  Suggestion:

  ```
  Also remove `join(plan.somaHome, "imports", "pai-packs", plan.skillName)` when overwriting.
  ```

### Security
The importer adds explicit dry-run/apply boundaries, denies likely secret paths, and keeps substrate-specific files opt-in. I found no concrete security issue in the changed paths under this lens.

_No findings._

### Architecture
The importer keeps the PAI pack migration contained behind a small public API and does not introduce reversed dependencies. The main architectural tradeoff is a single module owning discovery, classification, rendering, and writes, but for a V0 importer this does not yet create a clear rework risk.

_No findings._

### Performance
The importer is filesystem-heavy but runs as an explicit CLI import path, not a hot path. I did not find a concrete performance regression that warrants blocking this PR.

_No findings._

### Maintainability
The importer is compact enough for V0, but one planning function is accumulating validation, routing, and manifest setup in a way that will raise future change cost. No duplication or dead-code issues stand out in the diff.

- **[suggestion]** `src/pai-pack-importer.ts:188` — **Split importer plan construction**
  buildPaiPackImportPlan mixes target validation, collision checks, source-file policy checks, routing, and manifest injection in one function. Adding another classification or import policy will require editing the same long branch-heavy flow.
  
  Suggestion:

  ```
  Extract validateSkillTarget(), validateSourceFiles(), and buildRoutedFiles().
  ```


---

jcfischer (commented):

5 finding(s): 2 important, 3 suggestion.


### CodeQuality
The importer is broadly straightforward and covered by focused tests. I found one correctness issue in manifest path normalization that can misclassify targets by string prefix rather than path boundary.

- **[important]** `src/pai-pack-importer.ts:160` — **Manifest uses unsafe prefix check**
  `file.target.startsWith(skillRoot)` treats sibling paths with the same string prefix as inside the skill root. A target like `.../skills/telos-extra/...` would be rendered relative to `telos`, producing an incorrect manifest path.
  
  Suggestion:

  ```
  Use `relative(skillRoot, file.target)` and check that it is not absolute and does not start with `..`.
  ```

### Security
The importer has reasonable path normalization, reserved-name checks, default dry-run behavior, and secret-path denial for the documented high-risk filenames. I found no concrete security issue in the diff under this lens.

_No findings._

### Architecture
The importer introduces a coherent V0 boundary between PAI pack source material, Soma skill payloads, and substrate-specific archives. The main architectural risk is that the importer hard-codes pack layout and classification policy inside the filesystem importer, which will make future pack formats or adapters harder to evolve cleanly.

- **[important]** `src/pai-pack-importer.ts:147` — **Classification policy is embedded in importer**
  `routeSourceFile` mixes PAI pack schema knowledge, Soma target layout, and adapter-facing classification policy in one private function. As additional pack templates or substrate adapters arrive, this central importer becomes the place every projection rule must change.
  
  Suggestion:

  ```
  Extract pack routing/classification into a small policy module or table exported separately from the filesystem import implementation.
  ```

### Performance
The importer is mostly one-shot CLI work, so there are no hot-path or memory-growth concerns in the diff. One apply path serializes independent filesystem copies, which can add avoidable wall-clock time for larger packs.

- **[suggestion]** `src/pai-pack-importer.ts:280` — **Parallelize independent file copies**
  `importPaiPack` processes every routed file serially, so a pack with dozens of files pays read/write/mkdir latency one file at a time. The PR verification already cites 73 files for a real dry run, making this an inferable apply-time cost.
  
  Suggestion:

  ```
  Build per-file async tasks and `await Promise.all(...)`, while keeping `written` derived from the task results.
  ```

### Maintainability
The importer is generally organized around small helpers, but the main plan builder now concentrates several independent responsibilities. The CLI option parsing also picks up repeated source-specific guards that will get harder to extend as imports grow.

- **[suggestion]** `src/pai-pack-importer.ts:194` — **Split plan builder responsibilities**
  buildPaiPackImportPlan mixes home resolution, skill-name validation, collision checks, secret checks, substrate filtering, route expansion, and manifest insertion in one flow. Future importer changes will need to touch the same function for unrelated policies.
  
  Suggestion:

  ```
  Extract validation helpers such as assertSkillNameAllowed, assertDestinationWritable, and assertFilesImportable.
  ```

- **[suggestion]** `src/cli.ts:247` — **Deduplicate pai-pack flag guards**
  Each pai-pack-only flag repeats the same source check and error shape inline. Adding another import type or flag will keep expanding parseImportArgs with near-copied branches.
  
  Suggestion:

  ```
  Use a small requireImportSource(source, "pai-pack", arg) helper.
  ```


---

jcfischer (commented):

5 finding(s): 2 important, 3 suggestion.


### CodeQuality
The importer is generally clear and has focused coverage for routing, dry-run, apply, collisions, and denied files. One correctness gap remains around required pack structure validation, which can produce an incomplete Soma skill without failing early.

- **[important]** `src/pai-pack-importer.ts:184` — **Required pack files are not validated**
  The importer only requires README.md implicitly, then imports whatever collectFiles returns. A pack missing src/SKILL.md, INSTALL.md, or VERIFY.md can still plan or apply, producing an incomplete skill despite the documented V0 contract.
  
  Suggestion:

  ```
  After collectFiles, assert required paths include src/SKILL.md, INSTALL.md, and VERIFY.md before routing.
  ```

### Security
The importer adds filesystem-copy behavior with dry-run defaults, overwrite gating, reserved-name checks, path normalization, and explicit denials for likely secret files. I found no concrete security issue in the changed trust boundaries for this PR.

_No findings._

### Architecture
The importer is mostly separated into CLI, routing, and application layers, but the public import plan conflates generated artifacts with source-file routes. That weakens the plan as a stable contract for future adapters and review tooling.

- **[important]** `src/pai-pack-importer.ts:211` — **Generated manifest masquerades as README route**
  The plan exposes `soma-pack.json` as if it were copied from `README.md`, so consumers cannot distinguish original pack files from generated importer metadata. That makes the public plan contract brittle for later adapters that rely on `files` for provenance, review, or projection decisions.
  
  Suggestion:

  ```
  Add an explicit `origin` or `renderMode` field to the public plan file type, or represent generated manifest entries separately.
  ```

### Performance
The importer is a one-shot CLI path, so there are no hot-path concerns. The only concrete performance cost is apply-time file copying being fully serialized across independent files.

- **[suggestion]** `src/pai-pack-importer.ts:241` — **Serializes independent file copies**
  Each imported file waits for the previous mkdir/read/write to finish, so apply time scales with the sum of all file I/O. The PR notes real packs with 73 files, making this visible for template-heavy packs.
  
  Suggestion:

  ```
  Copy files with bounded concurrency, preserving per-file mkdir before write.
  ```

### Maintainability
The importer is mostly readable, with routing split out cleanly. The main maintainability cost is a couple of broad functions/tests that bundle independent concerns and will be harder to extend as importer rules grow.

- **[suggestion]** `src/pai-pack-importer.ts:145` — **Split plan construction phases**
  buildPaiPackImportPlan mixes option resolution, skill-name validation, overwrite checks, secret checks, substrate checks, routing, and manifest insertion in one sequential block. Adding new import classifications or guards will require editing the same function in multiple places.
  
  Suggestion:

  ```
  Extract validation and routing into focused helpers, e.g. validateSkillTarget(...) and buildRoutedFiles(...).
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:81` — **Split overloaded rejection test**
  This single test covers invalid names, reserved names, collisions, secret files, and substrate-specific files while mutating shared fixture state. A future failure will be harder to localize, and adding one more rejection rule will make the test broader still.
  
  Suggestion:

  ```
  Split into focused tests by rejection category.
  ```


---

jcfischer (commented):

### CodeQuality
The importer is generally clear and defensive for the main file-routing path. I found one CodeQuality issue around validation/error handling for incomplete packs.

- **[suggestion]** `src/pai-pack-importer.ts:87` — **Missing README gets raw filesystem error**
  README.md is part of the documented V0 pack shape, but a missing README fails as a low-level readFile ENOENT before pack validation runs. That makes incomplete-pack errors inconsistent and harder to act on from the CLI.
  
  Suggestion:

  ```
  Add README.md to REQUIRED_PACK_FILES and validate required files before readPackMetadata.
  ```

### Security
The importer handles obvious path and flag boundaries, but it can still copy likely secrets hidden under common service config filenames. That violates the stated secret-file deny boundary before writing pack contents into Soma-controlled locations.

- **[important]** `src/pai-pack-importer.ts:16` — **Secret denylist misses common config files**
  The documented invariant says likely secret files are denied, but the path denylist misses common credential-bearing files such as `settings.json`, `config.json`, `aws/credentials`, and `kube/config`. A pack can place real tokens in `src/Tools/config.json` or, with `--include-substrate-specific`, `claude/settings.json`, and the importer will copy them into `~/.soma`.
  
  Suggestion:

  ```
  Expand the denylist for known credential config paths, or scan copied text files for high-confidence secret keys before writing.
  ```

### Architecture
The importer is placed behind a cohesive public API with routing split into a small helper module, and the docs define a clear boundary between import and projection. I did not find architecture-lens issues that warrant changes before merge.

_No findings._

### Performance
The importer does filesystem traversal and copy work in straightforward batch CLI paths, with no sync I/O, network/database access, or hot-loop render behavior. No performance issue is concrete enough to warrant a finding from this diff.

_No findings._

### Maintainability
The importer is mostly straightforward, with routing split out cleanly. The main maintainability cost is concentrated in one planning function and one broad test that both bundle several independent concerns.

- **[suggestion]** `src/pai-pack-importer.ts:158` — **Split the import planner**
  buildPaiPackImportPlan now combines home resolution, metadata normalization, overwrite checks, required-file validation, secret filtering, substrate filtering, route construction, and manifest injection. That makes future changes to one policy harder to review without re-reading the whole planner.
  
  Suggestion:

  ```
  Extract focused helpers for skill validation, source-file policy checks, and routed-file construction.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:84` — **Split the bundled rejection test**
  This test covers invalid names, reserved names, collisions, secrets, and substrate-specific files using shared mutable fixture state. A failure in one branch can obscure which importer policy regressed.
  
  Suggestion:

  ```
  Split into separate tests by rejection policy.
  ```


---

jcfischer (commented):

6 finding(s): 2 important, 4 suggestion.


### CodeQuality
The importer is generally coherent for dry-run/apply flows, but one path handling edge case can misclassify or allow denied files on Windows-style paths supplied in tests or callers. Findings are limited to correctness and edge-case handling.

- **[important]** `src/pai-pack-importer.ts:24` — **Secret matching assumes POSIX separators**
  `isLikelySecretPath` tests slash-based patterns, but collected paths are normalized while callers or future tests may pass Windows-style separators into routing or checks. A path like `src\DashboardTemplate\.env` would not match the denylist, making the guard separator-dependent.
  
  Suggestion:

  ```
  Normalize candidate paths with `path.split(sep).join("/")` or replace backslashes before applying `SECRET_FILE_PATTERNS`.
  ```

### Security
The importer has explicit path-based guards for reserved names, overwrite, substrate-specific files, and many secret filenames. One documented secret-handling invariant is too narrow and can allow common token file names to be imported.

- **[important]** `src/pai-pack-importer.ts:21` — **Token file denylist is too narrow**
  The docs promise token files are denied, but the pattern only blocks exact basenames like token.txt or tokens.json. Common names such as github-token.txt or api_tokens.yml would be copied into the Soma skill or source archive.
  
  Suggestion:

  ```
  Reject basenames containing token, secret, credential, or key, except explicit examples.
  ```

### Architecture
The importer is mostly cohesive: routing, planning, application, and CLI exposure follow the existing import-command shape. One boundary concern is that the importer owns source-document archive layout and skill layout details in multiple modules, which may make future adapter routing harder to evolve.

- **[suggestion]** `src/pai-pack-routing.ts:23` — **Routing hard-codes target storage layout**
  The routing helper both classifies PAI pack files and decides concrete Soma storage paths, including skill payload paths and source-archive paths. That couples classification policy to current filesystem layout, so future adapter or manifest-only routing changes will need to edit this low-level classifier.
  
  Suggestion:

  ```
  Return route intent plus relative payload path, and let the importer map intents to Soma storage roots.
  ```

### Performance
The importer is CLI-oriented, so most added work is outside hot request paths. One copy path creates avoidable memory pressure for large template assets by buffering whole files before writing them.

- **[suggestion]** `src/pai-pack-importer.ts:269` — **Stream or copy files without buffering**
  The generic copy branch reads each source file fully into memory, then writes it back out. PAI pack templates can include binary assets, so large files create avoidable peak memory usage proportional to file size.
  
  Suggestion:

  ```
  Use fs.copyFile for renderMode "copy" instead of writeFile(await readFile(...)).
  ```

### Maintainability
The importer is mostly decomposed into small helpers, and the routing logic stays readable. The main maintainability cost is in broad CLI parsing branches and an overpacked rejection test that will be harder to extend cleanly.

- **[suggestion]** `src/cli.ts:243` — **Repeated source guard blocks**
  Each new pai-pack flag repeats the same source check and error shape inside parseImportArgs, making the parser harder to extend as import subcommands grow.
  
  Suggestion:

  ```
  Extract pai-pack option parsing or use a per-source option table.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:82` — **Broad rejection test mixes cases**
  One test covers invalid names, reserved names, collisions, secret detection, and substrate-specific routing with shared fixture mutation. That raises future-change cost because unrelated rejection behavior is coupled in one long test body.
  
  Suggestion:

  ```
  Split into focused tests by rejection category.
  ```


---

jcfischer (commented):

5 finding(s): 2 important, 3 suggestion.


### CodeQuality
The importer is mostly clear and covered by focused tests. I found one correctness issue in source archiving when substrate-specific paths contain reserved path segments.

- **[important]** `src/pai-pack-routing.ts:50` — **Substrate targets can escape archive root**
  Substrate-specific file paths are interpolated directly into the archive target. A pack entry whose relative path contains `..` segments can resolve outside `imports/pai-packs/<skill>/source` during `join`, so included substrate files may be written to the wrong location.
  
  Suggestion:

  ```
  Reject relative pack paths containing `..` path segments before routing or normalize and assert the final target remains within the archive root.
  ```

### Security
No concrete Security-lens issues were found in the provided diff. The importer defaults to dry-run, rejects likely secret paths, requires explicit substrate-specific inclusion, and routes writes through normalized skill names.

_No findings._

### Architecture
The importer keeps the main boundaries clean: routing is isolated, CLI integration stays mostly orchestration-only, and substrate-specific material is kept outside the skill payload. The one architectural concern is the public plan shape conflating copied files and generated artifacts.

- **[important]** `src/types.ts:300` — **Generated files need distinct provenance**
  `PaiPackImportFile.source` is required even when `origin` is `generated`, so the manifest is forced to point at `README.md` as a synthetic source. That makes the public plan contract ambiguous for future adapters that need to distinguish copied files from generated outputs.
  
  Suggestion:

  ```
  Use a discriminated union: source files require `source`; generated files carry `generator` or omit `source`.
  ```

### Performance
The importer is not on a hot runtime path, so most added work is acceptable for a CLI migration command. One apply-path loop serializes independent filesystem operations and can add visible latency for template-heavy packs.

- **[suggestion]** `src/pai-pack-importer.ts:273` — **File writes are fully serialized**
  Each imported file waits for the previous mkdir/copy/write to finish, so apply time becomes the sum of independent filesystem latencies. The verified Telos pack already has dozens of template files, and larger app templates will pay this directly.
  
  Suggestion:

  ```
  Process files with bounded concurrency, e.g. a small worker pool around the per-file write block.
  ```

### Maintainability
The importer is mostly readable, with routing split out cleanly. The main maintainability cost is that planning and rejection tests concentrate several independent concerns in single blocks, which will make future importer rules harder to change safely.

- **[suggestion]** `src/pai-pack-importer.ts:171` — **Split importer planning responsibilities**
  buildPaiPackImportPlan now handles home resolution, pack validation, collision checks, secret rejection, substrate-specific policy, route construction, and manifest insertion in one function. Future classification or policy changes will require editing a dense multi-concern block.
  
  Suggestion:

  ```
  Extract focused helpers such as validateSkillTarget, assertImportableSourceFiles, and buildRoutedImportFiles.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:84` — **Split broad rejection test**
  One test covers invalid names, reserved names, collisions, multiple secret patterns, Windows-style paths, and substrate-specific imports. A failure in this block will be harder to diagnose and discourages adding new rejection cases.
  
  Suggestion:

  ```
  Split into focused tests for name validation, collision handling, secret detection, and substrate-specific routing.
  ```


---

jcfischer (commented):

### CodeQuality
The importer is generally clear and covered by focused tests, but one path-safety check is ineffective for Windows-style separator input. That leaves an edge case where a synthetic backslash path can bypass the unsafe-path guard and be treated as a normal filename on POSIX.

- **[important]** `src/pai-pack-importer.ts:127` — **Normalize before unsafe path check**
  `assertSafePackPaths` splits on `/` before normalizing backslashes, so a path like `src\..\escape.md` is not rejected by the guard. The test currently expects this to throw as unsafe, but the implementation only detects it later because it is substrate-specific, making the guard incorrect for portable/template backslash paths.
  
  Suggestion:

  ```
  Normalize first: `const unsafePaths = sourceFiles.filter((path) => path.replace(/\\/g, "/").split("/").includes(".."));`
  ```

### Security
The importer has reasonable default denial for substrate-specific and likely secret paths, but one trust-boundary gap allows untrusted pack files to influence writes outside the intended pack root. This should be fixed before relying on the importer for reviewed PAI packs.

- **[blocker]** `src/pai-pack-importer.ts:96` — **Symlinked pack directories can escape import root**
  collectFiles trusts Dirent traversal and later copyFile reads the discovered source paths without resolving and proving they remain inside paiPackDir. A malicious pack can include a symlinked directory under an allowed prefix, causing files outside the pack root to be imported and potentially bypassing the documented secret-file denial by naming the symlink path innocuously.
  
  Suggestion:

  ```
  Resolve each file with realpath and reject it unless it remains within realpath(paiPackDir); skip or reject symlinked directories.
  ```

### Architecture
The importer is mostly shaped as a cohesive migration boundary, with routing split from apply-time filesystem work. One structural concern is that pack safety policy is embedded in the importer instead of living behind a reusable boundary for future importers/adapters.

- **[suggestion]** `src/pai-pack-importer.ts:23` — **Importer owns reusable file policy**
  Secret-path denial and unsafe path policy are embedded directly in the PAI pack importer, so future pack importers or projection adapters will need to duplicate or reach into this module. That weakens the importer boundary as Soma gains more migration sources.
  
  Suggestion:

  ```
  Move path safety classification into a shared import-policy module.
  ```

### Performance
The importer is a CLI path, so there are no hot render/server loops, but apply mode serializes every filesystem operation. For packs with dozens of template/source files, that makes wall-clock import time scale with the sum of all per-file mkdir/copy/read/write latencies.

- **[important]** `src/pai-pack-importer.ts:273` — **Apply copies files serially**
  Each routed file waits for mkdir plus copy/read/write before the next starts; the documented Telos dry-run already has 73 files, so apply latency accumulates one filesystem round trip at a time.
  
  Suggestion:

  ```
  Use bounded concurrency for per-file writes, and create unique target dirs once.
  ```

### Maintainability
The importer is generally structured around small helpers, but one planning function now owns several independent concerns. The test suite has one broad validation test that will be harder to extend or diagnose as importer rules grow.

- **[suggestion]** `src/pai-pack-importer.ts:190` — **Split plan construction responsibilities**
  buildPaiPackImportPlan mixes home resolution, metadata parsing, skill-name validation, collision checks, secret checks, routing, substrate checks, and file assembly in one function. Future importer rules will have to thread through the whole function instead of isolated validation and routing helpers.
  
  Suggestion:

  ```
  Extract validateSkillNameAndDestination(), validateSourceFiles(), and buildRoutedFiles().
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:92` — **Split broad validation test**
  This test covers invalid names, collisions, secrets, unsafe paths, and substrate-specific routing in one fixture-mutating flow. A failure early in the sequence hides later validation coverage and makes additions harder to place.
  
  Suggestion:

  ```
  Split into focused tests or a small table per validation category.
  ```


---

jcfischer (commented):

7 finding(s): 2 important, 5 suggestion.


### CodeQuality
The importer is generally coherent for the CodeQuality lens, but one edge case can crash valid imports with an unhelpful runtime error. The issue is in concurrent copy handling for colliding target paths after routing.

- **[important]** `src/pai-pack-importer.ts:293` — **Target collisions can race writes**
  Different source paths can normalize to the same target on case-insensitive filesystems or through generated routes, and the concurrent writer will race without a deterministic error. This can silently produce the wrong imported file or fail with a low-level filesystem error.
  
  Suggestion:

  ```
  Detect duplicate `file.target` values in the plan before writing and throw a clear importer error.
  ```

### Security
The importer handles several filesystem trust-boundary checks, including symlink refusal, target containment, overwrite gating, and path-based secret denial. I found one concrete gap where common secret directory layouts can still be imported.

- **[important]** `src/pai-pack-importer.ts:113` — **Secret directories are not denied**
  `isLikelySecretPath` checks suspicious basenames and a few exact credential filenames, but it does not reject files inside directories named `secrets`, `credentials`, or `tokens`. A PAI pack containing `src/Tools/secrets/prod.env` or `src/DashboardTemplate/credentials/aws` would be copied into the Soma skill despite the documented invariant that likely secret files are denied.
  
  Suggestion:

  ```
  Reject suspicious path segments, not only basenames, e.g. deny any segment matching `secrets?`, `credentials?`, or `tokens?`.
  ```

### Architecture
The importer fits the existing CLI/importer shape, but it concentrates several architectural responsibilities in one module. The main forward-compatibility risk is that pack safety, routing, rendering, and write orchestration will be harder to evolve independently as more pack formats or substrates arrive.

- **[suggestion]** `src/pai-pack-importer.ts:1` — **Importer owns too many concerns**
  This module combines path safety policy, secret-path policy, pack metadata parsing, SKILL frontmatter rewriting, manifest rendering, planning, and filesystem writes. That makes future importer variants or adapter-specific routing harder to add without editing the same high-churn file.
  
  Suggestion:

  ```
  Extract safety checks, metadata/frontmatter rendering, and manifest rendering into focused helpers.
  ```

- **[suggestion]** `src/cli.ts:201` — **Import CLI is becoming monolithic**
  Adding a third import source expands the shared parser with source-specific flags and widening option unions. More importers will keep increasing casts and conditional flag ownership in one function.
  
  Suggestion:

  ```
  Dispatch to source-specific import parsers after validating `source`.
  ```

### Performance
The importer avoids sync I/O and caps apply-time copy concurrency, so the main performance risk is planning-time filesystem traversal. One avoidable per-entry stat pattern makes dry-run slower on larger template-heavy packs.

- **[suggestion]** `src/pai-pack-importer.ts:105` — **Avoid per-entry lstat during traversal**
  `readdir(..., { withFileTypes: true })` already exposes directory, file, and symlink checks, but the walker still calls `lstat` for every entry. Large template trees pay an extra O(n) filesystem syscall cost in dry-run and apply planning.
  
  Suggestion:

  ```
  Use `entry.isSymbolicLink()`, `entry.isDirectory()`, and `entry.isFile()` directly; keep `realpath` only where still needed.
  ```

### Maintainability
The importer is readable overall, but the main planning path and one rejection test now concentrate many independent cases in single blocks. Splitting those paths would lower future change cost without changing behavior.

- **[suggestion]** `src/pai-pack-importer.ts:202` — **Split the import planner**
  buildPaiPackImportPlan mixes home resolution, required-file checks, skill-name validation, collision checks, secret checks, route filtering, manifest insertion, and target escape validation in one long flow. Future routing or validation changes will force readers to keep the whole pipeline in mind.
  
  Suggestion:

  ```
  Extract validation and route assembly helpers, e.g. validateSkillName, assertNoExistingSkill, rejectDisallowedFiles, and buildRoutedFiles.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:86` — **Split the rejection matrix**
  This test covers invalid names, reserved names, collisions, multiple secret patterns, ambiguous paths, symlinks, and substrate-specific files in one large body. A failure here will be harder to localize, and adding another rejection case will make the test harder to scan.
  
  Suggestion:

  ```
  Split into focused tests or a table for secret-path cases.
  ```


---

jcfischer (commented):

5 finding(s): 2 important, 3 suggestion.


### CodeQuality
The importer is generally well-covered for happy paths and refusals, but one correctness gap lets case-variant secret filenames bypass the denylist. That can lead to importing files the feature contract says are refused.

- **[important]** `src/pai-pack-importer.ts:159` — **Secret path matching is case-sensitive**
  `SECRET_FILE_PATTERNS` are tested against the original path, so case variants like `.ENV`, `Credentials.JSON`, or `.NPMRC` can bypass the denylist. The importer promises to refuse likely secret files, and this leaves common filesystem case variants unhandled.
  
  Suggestion:

  ```
  Lowercase `normalizedPath` before applying `SECRET_FILE_PATTERNS`, and add a case-variant test.
  ```

### Security
The importer handles several filesystem trust-boundary risks, including symlinks, path traversal, reserved names, overwrite gating, and likely secret filenames. I found no concrete security issue in this diff that meets the attack-path threshold for this lens.

_No findings._

### Architecture
The importer mostly preserves the Soma boundary by importing packs into skills and archiving substrate-specific material outside the skill payload. The main architecture gap is that the generated pack manifest is treated as private formatter output even though docs position it as a future adapter contract.

- **[important]** `src/pai-pack-importer.ts:212` — **Manifest contract is private**
  `soma-pack.json` is documented as the stable handoff for later adapters and review tools, but its schema is only an inline object inside `renderManifest`. That makes the adapter-facing contract hard to reuse, validate, or evolve without coupling future code to importer internals.
  
  Suggestion:

  ```
  Add an exported `PaiPackManifest` type/schema and render the manifest from that contract.
  ```

### Performance
The importer is mostly bounded CLI filesystem work and avoids sync I/O, unbounded caches, and sequential file copies on apply. One collision path does unnecessary full-pack traversal before it can fail.

- **[suggestion]** `src/pai-pack-importer.ts:230` — **Check collisions before full traversal**
  `buildPaiPackImportPlan` walks and realpaths the entire pack before checking whether the target skill already exists, so rerunning against an existing large pack does O(files) filesystem work only to throw.
  
  Suggestion:

  ```
  Read metadata first, check the target skill path, then call `collectFiles`.
  ```

### Maintainability
The importer is reasonably decomposed around routing and small validators, but the central plan builder and one rejection test are carrying too much change surface. The maintainability issues are refactorable without changing behavior.

- **[suggestion]** `src/pai-pack-importer.ts:231` — **Split the plan builder**
  buildPaiPackImportPlan combines option resolution, existence checks, metadata normalization, secret/substrate validation, route construction, and target validation in one high-state function. Future changes to classification or validation will be harder to localize.
  
  Suggestion:

  ```
  Extract validation and routed-file assembly helpers.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:86` — **Split the rejection mega-test**
  This test covers invalid names, reserved names, collisions, multiple secret patterns, unsafe paths, symlinks, and substrate-specific files in one long mutable fixture flow. A failure will be harder to localize, and adding cases requires threading more state through the same body.
  
  Suggestion:

  ```
  Split by rejection category; table-drive the secret-path cases.
  ```


---

jcfischer (commented):

5 finding(s): 1 blocker, 2 important, 2 suggestion.


### CodeQuality
The importer is generally clear and covered by focused tests. I found one correctness issue in overwrite/apply behavior that can lose an existing skill before the import is guaranteed to succeed.

- **[blocker]** `src/pai-pack-importer.ts:337` — **Overwrite can delete before failing**
  With --overwrite, the existing skill and source archive are removed before any source file is read or written. If a later copy/read/write fails, the previous skill is already lost and the import is left partial.
  
  Suggestion:

  ```
  Stage into a temporary directory first, then atomically replace the existing skill after all writes succeed.
  ```

### Security
The importer has solid path and symlink defenses for pack traversal, but its secret-file denylist misses common VCS metadata inside portable/template paths. That can copy credential-bearing repository config into Soma skills despite the documented secret-handling boundary.

- **[important]** `src/pai-pack-importer.ts:19` — **Git metadata can leak credentials**
  Files under `src/DashboardTemplate/.git/` or portable paths are classified as template/portable and copied, but `.git/config` can contain credentialed remote URLs. The documented invariant says likely secret files are denied by default, and this misses a common credential-bearing path.
  
  Suggestion:

  ```
  Deny VCS metadata directories such as `.git`, `.hg`, and `.svn`, or at least deny `.git/config` and related credential stores.
  ```

### Architecture
The importer is mostly cleanly isolated behind a new routing module and public plan/result types. The main architectural concern is that optional substrate archives leak into the skill manifest, making the skill artifact depend on state outside the skill root.

- **[important]** `src/pai-pack-importer.ts:216` — **Manifest includes external archive files**
  When `--include-substrate-specific` is used, `plan.files` includes targets under `imports/pai-packs/...`, so `soma-pack.json` inside the skill records files outside `skills/<skill>`. That weakens the skill as a cohesive portable artifact and couples future projection/publishing code to importer archive layout.
  
  Suggestion:

  ```
  Keep `soma-pack.json` limited to skill-root files, or write archive metadata to a separate import manifest under `imports/pai-packs/<skill>/`.
  ```

### Performance
The importer avoids sync I/O in async paths and uses bounded concurrency for writes. No concrete performance regression is visible in the provided diff.

_No findings._

### Maintainability
The importer is mostly decomposed into focused helpers, but a few new blocks make future changes harder by concentrating unrelated cases. The main maintainability cost is the oversized negative-path test, with one secondary extraction opportunity in the planner.

- **[suggestion]** `test/pai-pack-importer.test.ts:89` — **Split the mixed rejection test**
  This single test covers invalid names, collisions, secret patterns, unsafe paths, symlinks, and substrate-specific routing, so failures require scanning a long setup chain to find the broken concern. The repeated write/expect/rm pattern also makes adding a new rejection case easy to misplace.
  
  Suggestion:

  ```
  Split into focused tests, or table-drive the secret-path cases with a small helper.
  ```

- **[suggestion]** `src/pai-pack-importer.ts:233` — **Extract planner validation stages**
  buildPaiPackImportPlan mixes home resolution, pack validation, collision checks, secret checks, substrate checks, route construction, generated manifest insertion, and target validation in one function. The readable helper set already exists around it, so extracting the validation and routing stages would reduce state a maintainer must keep in mind.
  
  Suggestion:

  ```
  Extract validatePackInput(...) and buildRoutedFiles(...) from buildPaiPackImportPlan.
  ```


---

jcfischer (commented):

### CodeQuality
The importer is mostly coherent, but apply can leave a partially removed existing import when filesystem promotion fails. That makes one correctness issue worth fixing before merge.

- **[important]** `src/pai-pack-importer.ts:354` — **Overwrite can destroy existing import**
  When `overwrite` is true, the existing skill and archive are deleted before the staged replacement is promoted. If either `rename` fails after the deletes, the previous working import is lost and the new one may not be installed.
  
  Suggestion:

  ```
  Promote with a backup/rollback path, or rename existing roots aside and restore them if promotion fails.
  ```

### Security
The importer adds useful path and secret guards, but two file-system trust-boundary checks are raceable or lexical only. Both issues can let a malicious local pack/layout bypass the documented outside-home or symlink protections during apply.

- **[important]** `src/pai-pack-importer.ts:337` — **Symlink check is raceable**
  The plan rejects symlinks during collection, but apply later copies by path with copyFile, which follows a source path that can be swapped after validation. A writable malicious pack can race the import and copy files outside the pack into the Soma skill.
  
  Suggestion:

  ```
  Open/copy source files with no-follow semantics or re-lstat and realpath each source immediately before copying.
  ```

- **[important]** `src/pai-pack-importer.ts:348` — **Target containment ignores symlink parents**
  Target checks use lexical paths under somaHome, but apply renames into paths whose parent components may be symlinks. A crafted Soma home with skills or imports as a symlink can write imported content outside the configured Soma home despite the outside-home guard.
  
  Suggestion:

  ```
  Resolve and reject symlinked target ancestors before rm/rename, or verify real target parents stay under realpath(somaHome).
  ```

### Architecture
The importer is structurally coherent overall: routing is separated from application, and the public API follows existing import-plan/import-apply shapes. The main architectural risk is that CLI import parsing is becoming a central switchboard for importer-specific concerns.

- **[suggestion]** `src/cli.ts:204` — **Import parser is becoming importer-specific**
  `parseImportArgs` now owns option validation for three importers and returns a widened union that later requires casts. Each new importer will add more source-specific flags and branching to the CLI layer instead of keeping importer option parsing near the importer boundary.
  
  Suggestion:

  ```
  Split `pai`, `algorithm`, and `pai-pack` option parsing into source-specific parser functions.
  ```

### Performance
No performance issues found with concrete cost in this PR. The importer uses asynchronous filesystem APIs and bounded concurrency for writes, and the remaining full-tree planning work is proportional to the explicit import operation.

_No findings._

### Maintainability
The importer is mostly straightforward, but two new flows concentrate unrelated cases into long bodies that will be harder to extend. The issues are maintainability-only and are refactor suggestions rather than merge blockers.

- **[suggestion]** `src/pai-pack-importer.ts:251` — **Split import plan assembly**
  buildPaiPackImportPlan mixes path discovery, pack validation, collision checks, routing, substrate gating, and generated-file assembly in one long function. Each new route or validation rule will make the function harder to scan and test in isolation.
  
  Suggestion:

  ```
  Extract route assembly and validation into small helpers such as buildRoutedPaiPackFiles and assertAllowedPaiPackRoutes.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:82` — **Split broad rejection test**
  This test mutates several fixtures and verifies invalid names, reserved names, collisions, secret detection, path safety, symlinks, and substrate-specific behavior in one large body. A failure midway obscures which rule regressed and makes future cases easy to add in the wrong place.
  
  Suggestion:

  ```
  Split by concern or use table-driven helpers for secret/path rejection cases.
  ```


---

jcfischer (commented):

4 finding(s): 1 blocker, 1 important, 2 suggestion.


### CodeQuality
The importer is generally well covered, but one apply path can delete pre-existing archive data after a failed import. This is a correctness and error-handling issue in the new PAI pack importer.

- **[blocker]** `src/pai-pack-importer.ts:435` — **Failed import can delete archive data**
  When `--include-substrate-specific` is used without `--overwrite`, an existing `imports/pai-packs/<skill>` archive is not backed up before the staged archive rename. If that rename fails, the catch block unconditionally removes `sourceArchiveRoot`, deleting pre-existing data that the caller did not authorize overwriting.
  
  Suggestion:

  ```
  Reject an existing `sourceArchiveRoot` unless `--overwrite` is set, or only remove it in rollback when this import created it.
  ```

### Security
The importer mostly validates paths and rejects symlinks, traversal, substrate-specific files, and several secret-like paths before copying. One documented secret-handling invariant is incomplete enough to allow common private key files into imported skills.

- **[important]** `src/pai-pack-importer.ts:16` — **Common SSH private keys are not denied**
  The denylist blocks id_rsa and id_ed25519 but misses common private key filenames such as id_ecdsa and id_dsa. A pack containing src/Tools/id_ecdsa or a template copy of that file would pass validation and be copied into ~/.soma/skills despite the documented private-key refusal invariant.
  
  Suggestion:

  ```
  Deny id_ecdsa, id_dsa, and id_* private key names, or detect OpenSSH private key content before copy.
  ```

### Architecture
The PR keeps the importer mostly cohesive: CLI wiring stays thin, routing is isolated, and public types describe the planned/imported surface. I do not see an architecture-lens issue that warrants blocking or requesting changes before merge.

_No findings._

### Performance
The importer does filesystem-heavy work, but it is scoped to a CLI import path and uses bounded concurrency for apply-time writes. I did not find concrete performance regressions that warrant blocking this PR.

_No findings._

### Maintainability
The importer is mostly organized around clear routing and planning primitives, but the apply path concentrates too many responsibilities in one function. Test coverage is broad, though one rejection test packs many unrelated fixtures into a long scenario that will be costly to extend.

- **[suggestion]** `src/pai-pack-importer.ts:379` — **Split importer apply workflow**
  `importPaiPack` combines staging writes, render-mode dispatch, overwrite backup, rollback, cleanup, and result assembly in one nested function. Future changes to archive handling or rollback semantics will require holding too much state in one place.
  
  Suggestion:

  ```
  Extract staged file writing and commit/rollback into helpers.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:91` — **Split rejection matrix test**
  This test combines invalid names, collisions, secret detection, path validation, symlinks, and substrate-specific behavior in one long flow. A failure midway obscures which policy regressed and makes adding new cases more expensive.
  
  Suggestion:

  ```
  Split by rejection category or use table-driven subcases.
  ```


---

jcfischer (commented):

5 finding(s): 1 blocker, 1 important, 3 suggestion.


### CodeQuality
The importer is generally clear and well-covered, but the overwrite rollback path has a correctness issue that can delete the user's previous skill if recovery fails. That should be fixed before merge because it affects failure handling during apply.

- **[blocker]** `src/pai-pack-importer.ts:472` — **Rollback can delete the backup**
  The `finally` block removes `backupRoot` even if the rollback path failed before restoring the original skill or archive. In an overwrite apply, a promotion failure plus a restore failure can leave the original content only in `backupRoot`, then this line deletes it.
  
  Suggestion:

  ```
  Only remove `backupRoot` after a successful promotion or successful rollback; preserve it when rollback restoration throws.
  ```

### Security
The importer adds meaningful trust-boundary checks for pack paths, symlinks, target containment, overwrite handling, and likely secret filenames. I found no concrete security issue in the diff for this lens.

_No findings._

### Architecture
The importer keeps the external surface small and introduces a separate routing module, which fits the dry-run/apply boundary. One policy concern is still embedded in the filesystem importer, making future pack analysis or adapters depend on import execution internals.

- **[suggestion]** `src/pai-pack-importer.ts:30` — **Extract pack policy from importer**
  Secret-path denial and pack acceptance policy live inside the filesystem importer, while the docs describe classification as a stable input for later adapters and review tools. Those consumers would need to depend on importer internals instead of a pure policy/classification module.
  
  Suggestion:

  ```
  Move deny patterns and pack validation into a pure `pai-pack-policy` module used by the importer.
  ```

### Performance
The importer is mostly bounded to filesystem traversal and copy work, with concurrency applied during writes. One traversal path can still spend substantial time walking files that will be refused anyway.

- **[important]** `src/pai-pack-importer.ts:124` — **Refuse VCS directories before traversal**
  `collectFiles` descends into `.git`, `.hg`, and `.svn` directories before `isLikelySecretPath` later rejects them. A copied template repository can contain thousands of VCS object files, making dry-run and apply pay a large avoidable traversal cost before failing.
  
  Suggestion:

  ```
  Skip or reject VCS directories in the `entry.isDirectory()` branch before recursing.
  ```

### Maintainability
The importer is reasonably decomposed for planning and routing, but the apply path concentrates staging, promotion, backup, and rollback state in one long function. Tests cover the behavior well, though one rejection test is carrying repeated setup/assert/remove logic that will be costly to extend.

- **[suggestion]** `src/pai-pack-importer.ts:377` — **Split apply promotion flow**
  importPaiPack mixes rendering staged files, backup setup, promotion state, rollback, and cleanup in one large function. The promotedSkillRoot/promotedArchiveRoot and backup restoration branches make future overwrite changes hard to reason about locally.
  
  Suggestion:

  ```
  Extract stagePaiPackFiles and promotePaiPackImport helpers, with rollback contained in the promotion helper.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:88` — **Table-drive rejection cases**
  The rejection test repeats the same write, expect rejection, and cleanup pattern for many secret-path variants. Adding another denied path requires editing a long procedural block instead of appending a case.
  
  Suggestion:

  ```
  Use a cases array plus a helper that creates the path, asserts the error, and removes it.
  ```


---

jcfischer (commented):

5 finding(s): 2 important, 3 suggestion.


### CodeQuality
The importer is generally well-covered for expected paths and failure cases. I found one correctness issue in rollback handling that can mask the original failure and leave the destination absent after a failed restore.

- **[important]** `src/pai-pack-importer.ts:467` — **Rollback failure masks original error**
  If restoring the backup fails, the code throws only the rollback error after already removing any promoted destination. Callers lose the original import failure, and the skill/archive may be left missing with no result context.
  
  Suggestion:

  ```
  Throw an error that includes both the original error and rollback error, or preserve backups and report the failed restore explicitly.
  ```

### Security
The importer has solid path containment checks and refuses symlinks, VCS metadata, and many secret-looking files. One documented local-settings secret pattern is missed, allowing a common credential file to be imported.

- **[important]** `src/pai-pack-importer.ts:21` — **local.settings.json is not denied**
  Docs say local settings files are denied, but Azure-style local.settings.json does not match the current patterns. A pack containing local database or service credentials in that file would be copied into the Soma skill/archive.
  
  Suggestion:

  ```
  Add a local.settings.json deny pattern and a regression test.
  ```

### Architecture
The importer is structurally well-contained: routing, public types, CLI integration, and filesystem application have clear boundaries for a V0 migration tool. I did not find architecture-lens issues that would force rework before merge.

_No findings._

### Performance
The importer is a one-shot filesystem workflow with bounded concurrency, so there are no hot-path or unbounded-growth concerns. One repeated filesystem lookup scales directly with pack size and is avoidable.

- **[suggestion]** `src/pai-pack-importer.ts:233` — **Avoid repeated pack root realpath**
  resolveSafeSourceFile recomputes realpath(packRoot) for every copied or rewritten file, adding one redundant filesystem lookup per source file during apply. The reported Telos run has 73 files, so this is a measurable extra syscall fanout in the main import loop.
  
  Suggestion:

  ```
  Compute realPackRoot once before mapWithConcurrency and pass it into resolveSafeSourceFile.
  ```

### Maintainability
The importer is mostly organized around small helpers, but the apply path concentrates staging, backup, promotion, rollback, and cleanup in one function. The test suite also has one oversized rejection test that will be hard to extend cleanly.

- **[suggestion]** `src/pai-pack-importer.ts:374` — **Split import apply workflow**
  importPaiPack mixes staging, file rendering, backup, promotion, rollback, cleanup, and result assembly in one long nested flow. Future changes to overwrite or rollback behavior will require keeping too much state in one function.
  
  Suggestion:

  ```
  Extract helpers such as stageImportFiles, backupExistingImport, promoteStagedImport, and rollbackImport.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:96` — **Split oversized rejection test**
  This single test covers invalid names, collisions, secrets, VCS paths, symlinks, unsafe names, and substrate-specific routing. Adding another rejection case requires editing a long setup-heavy block instead of a focused scenario.
  
  Suggestion:

  ```
  Split into focused tests or a small table for secret-path cases.
  ```


---

jcfischer (commented):

5 finding(s): 1 blocker, 1 important, 3 suggestion.


### CodeQuality
The importer is generally clear and well-covered, but one correctness issue can make overwrite rollback fail and leave the import partially applied. No other CodeQuality-scope issues stood out from the diff.

- **[blocker]** `src/pai-pack-importer.ts:460` — **Cleanup deletes rollback backups**
  When rollback fails, `cleanupRoots.pop()` removes `backupRoot` from cleanup, not `stageRoot`, so the finally block can delete the only backup of the previous skill/archive. This makes a failed overwrite harder to recover and can leave the user without the original data.
  
  Suggestion:

  ```
  Remove `stageRoot` from cleanup instead, and keep `backupRoot` for recovery.
  ```

### Security
The importer has solid path traversal, symlink, overwrite, and filename-based secret guards for local pack ingestion. I found no concrete security attack path in the diff that violates the documented importer boundaries.

_No findings._

### Architecture
The importer has a reasonable first boundary between CLI, routing, and apply logic, and it keeps substrate projection out of the skill payload. The main architectural risk is that the routing module mixes file classification with Soma storage layout, which will make future adapter or layout changes ripple through classification code.

- **[important]** `src/pai-pack-routing.ts:22` — **Routing owns storage layout**
  `routePaiPackSourceFile` both classifies pack files and chooses concrete Soma target paths. That couples the portable pack taxonomy to today's `skills/` and `imports/` layout, making future projection/layout changes require edits in classification logic.
  
  Suggestion:

  ```
  Return classification plus relative payload/archive path, and let the importer map routes to Soma roots.
  ```

### Performance
The importer is not on a request hot path, and most file work is bounded to explicit import runs. One avoidable per-file filesystem call adds measurable overhead for larger packs during apply.

- **[suggestion]** `src/pai-pack-importer.ts:216` — **Avoid repeated pack-root realpath**
  resolveSafeSourceFile recomputes realpath(packRoot) for every copied file, adding one extra filesystem syscall per source file during apply. The PR's own Telos run has 73 files, so this is already dozens of redundant syscalls.
  
  Suggestion:

  ```
  Compute realPackRoot once in importPaiPack and pass it into resolveSafeSourceFile.
  ```

### Maintainability
The importer is generally decomposed into focused helpers, but the apply path and one rejection test have accumulated several responsibilities in single blocks. These are maintainable now, yet future changes to staging, rollback, or denylist coverage will be harder than necessary.

- **[suggestion]** `src/pai-pack-importer.ts:390` — **Split import apply orchestration**
  `importPaiPack` combines staging, rendering, backup, promotion, rollback, cleanup, and result construction in one deeply nested function. Changes to overwrite or rollback behavior will require keeping too much state in view at once.
  
  Suggestion:

  ```
  Extract staged file writing, promotion, and rollback into named helpers.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:89` — **Table-drive rejection cases**
  The rejection test repeats write/expect/remove setup for many secret and path cases, making future denylist additions noisy and easy to mis-edit. The substrate-specific and collision cases are distinct, but the secret-path cases share one shape.
  
  Suggestion:

  ```
  Use a small helper or table for `{ path, content, expected }` rejection cases.
  ```


---

jcfischer (commented):

### CodeQuality
The importer is generally careful about path validation and error handling. I found one correctness issue where apply-mode plans are reported as dry-run in generated metadata.

- **[important]** `src/pai-pack-importer.ts:343` — **Applied manifest records dry-run mode**
  buildPaiPackImportPlan always returns apply: false, and importPaiPack reuses that plan when rendering soma-pack.json. Any metadata or downstream code reading the manifest cannot distinguish an applied import from a dry-run plan.
  
  Suggestion:

  ```
  Set apply from the caller, or override the internal plan to apply: true inside importPaiPack before rendering.
  ```

### Security
The importer adds explicit path, symlink, overwrite, and likely-secret guards around a filesystem trust boundary. I did not find a concrete security issue in the diff under this lens.

_No findings._

### Architecture
The importer has a coherent top-level boundary: routing is split out, the public API exposes plan/apply, and substrate-specific payloads are kept outside skill content. One durable manifest contract loses provenance that the plan already models, which will make later migration and projection tooling harder to build without a schema bump.

- **[important]** `src/types.ts:318` — **Manifest drops file provenance**
  `PaiPackManifestFile` keeps only `origin`, while the import plan distinguishes source paths and generated file generators. Since `soma-pack.json` is a durable v1 contract, omitting provenance forces future adapters or re-import tools to infer source mapping or introduce a breaking schema change.
  
  Suggestion:

  ```
  Include `source` for source files and `generator` for generated files in `PaiPackManifestFile`.
  ```

### Performance
No concrete performance regression stands out in this PR. The importer does bounded filesystem traversal and uses limited concurrency for apply-time writes without introducing hot-path sync I/O, N+1 network/database access, or unbounded in-memory growth.

_No findings._

### Maintainability
The importer is mostly decomposed into focused helpers, but the apply path concentrates staging, promotion, rollback, and cleanup in one nested function. Tests cover the surface well, with one broad rejection test that is becoming hard to extend or diagnose.

- **[important]** `src/pai-pack-importer.ts:386` — **Split apply transaction logic**
  importPaiPack mixes file rendering, staging, backup creation, promotion, rollback, and cleanup in one large nested flow. Future changes to overwrite or archive behavior will require holding too much state across the function.
  
  Suggestion:

  ```
  Extract helpers such as stagePaiPackFiles() and promoteStagedImportWithRollback().
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:91` — **Split broad rejection test**
  This single test covers invalid names, collisions, secret patterns, VCS metadata, unsafe paths, symlinks, and substrate-specific routing. A failure in one case can obscure the intended behavior being protected.
  
  Suggestion:

  ```
  Split into focused tests by rejection category.
  ```


---

jcfischer (commented):

### CodeQuality
The importer is broadly covered and the core dry-run/apply flow is coherent. I found one correctness issue where custom `--soma-home` values are ignored when deriving the default PAI pack source path.

- **[important]** `src/pai-pack-importer.ts:43` — **Default pack path ignores somaHome**
  `resolvePackHomes` derives the default pack directory from `homeDir` even when callers provide only `somaHome`. This can make `soma import pai-pack --soma-home <dir>` look in the real user home for packs instead of the selected Soma home context.
  
  Suggestion:

  ```
  Derive defaults consistently from `somaHome` or require `--pai-pack-dir` before resolving unrelated homes.
  ```

### Security
The importer has solid baseline path, symlink, overwrite, and secret-path checks for normal pack contents. I found no concrete security issue in the diff that meets the lens threshold.

_No findings._

### Architecture
The importer keeps the main Soma skill boundary intact by routing PAI source docs and substrate-specific material into distinct locations. The main architectural concern is that the durable manifest schema captures command execution state rather than only stable pack metadata.

- **[suggestion]** `src/types.ts:334` — **Manifest stores invocation state**
  `PaiPackManifest.apply` makes the persisted skill manifest depend on how the importer was invoked, not on the imported pack's stable structure. That couples future projection or refresh tooling to CLI execution mode and makes the public manifest contract less cohesive.
  
  Suggestion:

  ```
  Remove `apply` from `PaiPackManifest`; keep dry-run/apply state only on `PaiPackImportPlan`.
  ```

### Performance
The importer avoids sync I/O and applies file writes with bounded concurrency, so there are no broad hot-path concerns. One apply-path inefficiency adds redundant filesystem resolution per copied file and is worth tightening if packs grow.

- **[suggestion]** `src/pai-pack-importer.ts:233` — **Cache pack root realpath**
  resolveSafeSourceFile recomputes realpath(packRoot) for every copied source file during apply. A real Telos import already has 73 files, so this adds dozens of identical filesystem calls on the apply path.
  
  Suggestion:

  ```
  Compute realPackRoot once in stagePaiPackFiles and pass it into resolveSafeSourceFile.
  ```

### Maintainability
The importer is generally organized around small helpers, but the apply path and rejection tests now carry most of the maintainability risk. The clearest issues are a high-arity rollback helper and a large copy-pasted test case that will be costly to extend.

- **[suggestion]** `src/pai-pack-importer.ts:426` — **Group rollback parameters**
  promoteStagedImportWithRollback takes seven positional parameters, several of them related root paths with similar meanings. Future changes can easily pass the wrong root or cleanup set without type help.
  
  Suggestion:

  ```
  Pass a single context object, e.g. { plan, stageRoot, backupRoot, skillRoot, sourceArchiveRoot, overwrite, cleanupRoots }.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:87` — **Extract repeated rejection fixtures**
  The rejection test repeats the same write, expect rejection, and cleanup pattern for many secret-path cases. Adding a new denied path now requires editing a long scrolling test body instead of adding a table row.
  
  Suggestion:

  ```
  Use a table-driven helper for rejected fixture mutations and keep substrate, symlink, and collision cases separate.
  ```


---

jcfischer (commented):

4 finding(s): 1 important, 2 suggestion, 1 nit.


### CodeQuality
The importer is broadly clear and defensively implemented for the reviewed paths. I found one error-handling issue where successful imports can still be reported as failures during temporary cleanup.

- **[important]** `src/pai-pack-importer.ts:517` — **Cleanup failure masks successful import**
  The `finally` block awaits removal of staging and backup directories after promotion. If cleanup fails after files were promoted, `importPaiPack` rejects even though the import already succeeded, leaving callers with an incorrect failure result.
  
  Suggestion:

  ```
  Catch cleanup errors after promotion, or report them separately without overriding a successful import.
  ```

### Security
The importer adds explicit dry-run/apply behavior, path containment checks, symlink refusal, overwrite gating, and likely-secret filename denial at the main filesystem trust boundary. I found no concrete security issue in this diff under the Security lens.

_No findings._

### Architecture
The importer adds a coherent Soma-side migration boundary: routing is separated, public types are exported, and CLI integration stays at the command edge. I do not see an architecture-lens issue that warrants blocking or changing this PR before merge.

_No findings._

### Performance
No concrete performance regressions are evident in the diff. The importer uses async filesystem APIs, bounded copy concurrency, and skips large generated directories during traversal.

_No findings._

### Maintainability
The importer is mostly factored into named validation, routing, staging, and rendering helpers, so the new surface is readable despite its size. The main maintainability costs are a complex rollback function signature, one dead branch, and an overly broad table-worthy test.

- **[suggestion]** `src/pai-pack-importer.ts:431` — **Group promotion context parameters**
  promoteStagedImportWithRollback takes seven positional parameters, including related roots, overwrite state, and cleanup bookkeeping. Future changes to promotion or rollback will be easier to review if those travel as one named context.
  
  Suggestion:

  ```
  Pass a PromotionContext object with roots, overwrite, and cleanupRoots.
  ```

- **[nit]** `src/pai-pack-importer.ts:277` — **Remove unreachable manifest branch**
  renderManifest filters to skillFiles inside skillRoot, then checks isWithinPath(skillRoot, file.target) again before choosing the manifest target. The else branch is dead under the current filter and adds noise.
  
  Suggestion:

  ```
  Use relative(skillRoot, file.target) directly.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:92` — **Split repeated rejection cases**
  The invalid/reserved/colliding/secret/substrate-specific test repeats write/expect/remove blocks for many secret path cases. A table-driven helper would make adding new denied paths cheaper and reduce scan time.
  
  Suggestion:

  ```
  Extract assertRejectedPackFile(relativePath, contents, message) and loop over cases.
  ```


---

jcfischer (commented):

### CodeQuality
The importer is generally careful about dry-run/apply behavior, path validation, and rollback. I found one correctness issue where the generated SKILL.md can become invalid when pack metadata contains newline-bearing frontmatter values.

- **[important]** `src/pai-pack-importer.ts:310` — **Skill name is not YAML-escaped**
  The rewritten frontmatter writes `name: ${skillName}` directly. A CLI-provided `--skill-name` containing a newline can normalize to a string with an embedded line break, producing malformed frontmatter and potentially extra YAML keys.
  
  Suggestion:

  ```
  Validate normalized skill names with `/^[a-z0-9]+(?:-[a-z0-9]+)*$/` or JSON-quote the name field.
  ```

### Security
The importer adds several explicit trust-boundary checks for paths, symlinks, overwrite behavior, and likely secret files. I found no concrete security issue in the provided diff.

_No findings._

### Architecture
The importer keeps the CLI integration and routing surface mostly cohesive, but the manifest boundary drops part of the import plan. That weakens the documented adapter/review-tool contract for future projections.

- **[important]** `src/pai-pack-importer.ts:269` — **Manifest omits archived source files**
  `renderManifest` filters to files under the skill root, so included substrate-specific files are copied but never represented in `soma-pack.json`. The docs describe classification as the stable place for later adapters and review tools to decide projection behavior, and this loses that inventory at the manifest boundary.
  
  Suggestion:

  ```
  Include archive entries in the manifest with relative archive targets, or write a separate archive manifest.
  ```

### Performance
The importer uses async filesystem APIs and bounded copy concurrency, with no sync I/O, network/database pattern, or unbounded cache growth visible in the diff. I did not find a concrete performance regression for this lens.

_No findings._

### Maintainability
The importer is readable overall, with routing separated from planning and staging. The main maintainability cost is concentrated in rollback promotion and a large duplicated rejection test.

- **[suggestion]** `src/pai-pack-importer.ts:409` — **Rollback promotion repeats root handling**
  promoteStagedImportWithRollback duplicates the same backup, promote, cleanup, and restore flow for skill and archive roots, making future changes easy to apply to one path but miss the other.
  
  Suggestion:

  ```
  Model skill/archive as entries and loop through backup, promote, and restore helpers.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:89` — **Rejection test repeats secret cases**
  The large rejection test repeats the same write-expect-remove pattern for each secret path, so adding or changing cases requires editing copy-pasted blocks.
  
  Suggestion:

  ```
  Use a table of secret file cases with a shared helper.
  ```


---

jcfischer (commented):

### CodeQuality
The importer is mostly coherent for the CodeQuality lens, but one error path can leave successful overwrite backups behind as leaked state. That makes repeated applies harder to reason about after a partial failure.

- **[important]** `src/pai-pack-importer.ts:523` — **Rollback leaves restored backups undeleted**
  When promotion fails after backing up existing roots, rollback restores the backups but still keeps `backupRoot` in `cleanupRoots`. The `finally` block then tries to remove the now-empty backup root, but if restored backup parents remain or rollback partially moves only one root, backup residue can be left under `.tmp`.
  
  Suggestion:

  ```
  After each successful restore, remove the corresponding backup parent or keep cleanup targeted to the whole backup root after all restores complete.
  ```

### Security
The importer adds explicit trust-boundary checks for path traversal, symlinks, substrate-specific files, overwrite behavior, and likely secret files. I found no concrete security attack path in the diff for this lens.

_No findings._

### Architecture
The importer adds a coherent PAI-pack-to-Soma-skill boundary with routing separated from staging and promotion. The public surface is narrow and forward-compatible enough for V0, with no architecture findings in scope.

_No findings._

### Performance
The importer does bounded filesystem traversal and uses capped parallelism for apply-time copies. I did not find a concrete performance regression in this diff.

_No findings._

### Maintainability
The importer is reasonably decomposed at the routing/type level, but the apply path concentrates rollback state in one complex function. Tests also add a long repeated rejection matrix that will be costly to extend.

- **[suggestion]** `src/pai-pack-importer.ts:466` — **Split rollback promotion steps**
  promoteStagedImportWithRollback mixes backup, promotion, rollback, and cleanup-root mutation in one nested function with several state flags. Future overwrite or archive changes will require editing the whole transaction flow at once.
  
  Suggestion:

  ```
  Extract backupExistingRoots, promoteStagedRoots, and rollbackPromotedRoots helpers around a small transaction context.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:83` — **Table-drive rejection cases**
  The invalid/secret/substrate test repeats write, expect rejection, and cleanup blocks across many cases. Adding new denied paths now means copying another multi-line sequence into an already large test body.
  
  Suggestion:

  ```
  Use a small array of cases plus a helper that mutates the fixture, asserts the error text, and cleans up.
  ```


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### CodeQuality
The importer is generally careful about path handling and has focused coverage for dry-run/apply behavior, overwrite, secrets, symlinks, and substrate-specific routing. I found no CodeQuality-lens issues that warrant blocking or changing this PR.

_No findings._

### Security
The importer has meaningful path, symlink, overwrite, and likely-secret guards at the local filesystem boundary. I found no concrete security attack path in this diff under the documented importer contract.

_No findings._

### Architecture
The PR keeps the new PAI pack importer behind a narrow public API and separates path routing into its own module. I did not find an architecture-lens issue that warrants blocking or requesting changes before merge.

_No findings._

### Performance
The PR adds a filesystem-heavy importer, but the expensive work is bounded to explicit CLI import/dry-run operations and uses async I/O with capped concurrency during staging. I do not see a concrete performance regression that warrants blocking this PR under this lens.

_No findings._

### Maintainability
The importer is mostly decomposed, but two newly introduced areas concentrate many independent cases in one place. Splitting the rollback flow and rejection tests would reduce future-change cost without changing behavior.

- **[suggestion]** `src/pai-pack-importer.ts:483` — **Split rollback promotion stages**
  promoteStagedImportWithRollback mixes backup, promotion, rollback, cleanup bookkeeping, and error wrapping in one stateful function. Future changes to one phase require keeping several mutable flags and paths in mind at once.
  
  Suggestion:

  ```
  Extract backupExistingRoots, promoteStagedRoots, and rollbackPromotedRoots helpers around a small context object.
  ```

- **[suggestion]** `test/pai-pack-importer.test.ts:99` — **Split broad rejection test**
  This single test covers invalid names, collisions, secrets, VCS metadata, path separators, symlinks, and substrate files. A failure in one setup step can hide unrelated rejection coverage and makes additions hard to scan.
  
  Suggestion:

  ```
  Split by rejection class or use a small table for secret-path cases.
  ```


---

PR #4, opened 2026-05-15
Quiet Soma hook classification output

jcfischer:
## Summary
- Collapse visible Soma prompt classification hook output to one concise line
- Apply the same quieter summary to the generated Pi.dev extension
- Update install/projection tests to ensure the old multi-line Operating requirement block is gone

## Verification
- bun test test/install.test.ts test/home-projection.test.ts
- bun run lint
- bun run typecheck
- bun test
- git diff --check


jcfischer (commented):

1 finding(s): 1 suggestion.


### CodeQuality
The change consistently collapses prompt classification output while preserving the key Algorithm/non-Algorithm guidance. I found no CodeQuality issues in the diff.

_No findings._

### Maintainability
The diff is small and mostly reduces complexity by replacing the multi-line builder with direct returns. The only maintainability cost is that the new classification wording is duplicated between the live Codex hook and the generated Pi.dev extension.

- **[suggestion]** `src/adapters/codex-hook-entry.mjs:87` — **Duplicated classification formatter**
  The concise Soma output format now exists both here and inside the generated Pi.dev extension string, so future wording changes must be kept in sync manually.
  
  Suggestion:

  ```
  Extract the formatter text into a shared template/helper used by both adapters, or add a single source-of-truth fixture for the generated extension.
  ```


---

PR #19, opened 2026-05-15
feat(soma): add path protection policy

jcfischer:
## Summary

Adds Soma portable policy for preventing destructive operations on protected directories.

### What it does

The policy engine now supports three action types: **write**, **delete**, and **modify**.

- **write** (unchanged): checks for private Soma marker leaks to public files
- **delete**: blocks `rm -rf`, `rmdir`, `trash`, `mv` on protected paths  
- **modify**: blocks writes/edits targeting protected paths

### Protected paths (default)

`~/.soma`, `~/.claude`, `~/.pi`, `~/.config/cortex`, `~/.config/metafactory`, `~/.config/k`

### New modules

- **`src/policy-path-guard.ts`** — Bash command parser (handles `rm -rf`, `rmdir`, `trash`, `mv`, quoted paths, chain operators, glob patterns) + protected path checker
- **`src/adapters/pi-dev-path-guard.ts`** — Generates a pi.dev TypeScript extension that intercepts `tool_call` events and blocks destructive operations on protected paths

### Substrate integration

The pi.dev adapter now includes `soma-path-guard.ts` in both `buildPiDevContext` and `buildPiDevHomeContext` bundles. When loaded, it:

1. Intercepts `tool_call` for `bash` commands — parses for destructive operations
2. Blocks with `{ block: true, reason: "..." }` when targets hit protected paths
3. Also guards `write`/\`edit\` tools from writing to protected directories

### CLI

```bash
soma policy check --action delete --destination ~/.claude/memory
# → decision: deny (protected-path: ~/.claude)

soma policy check --action write --destination /tmp/public.md --content ok
# → decision: allow (unchanged behavior)
```

### Verification

- Typecheck: ✅
- Lint: ✅  
- Tests: ✅ (169 pass, 0 fail — 33 new)


jcfischer (commented):

5 finding(s): 5 important; 5 lens(es) failed to run: CodeQuality, Security, Architecture, Performance, Maintainability.


### CodeQuality — DID NOT RUN
> ⚠ Lens failed to execute. Verdict cannot rely on this lens's coverage; re-run before merging.

  substrate exited with code 2: error: unexpected argument '--ask-for-approval' found

  tip: to pass '--ask-for-approval' as a value, use '-- --ask-for-approval'

Usage: codex exec [OPTIONS] [PROMPT]
       codex exec [OPTIONS] <COMMAND> [ARGS]

For more information, try '--help'.


### Security — DID NOT RUN
> ⚠ Lens failed to execute. Verdict cannot rely on this lens's coverage; re-run before merging.

- **[important]** `(lens output):0` — **Security: model deviated from JSON contract**
  substrate exited with code 2: error: unexpected argument '--ask-for-approval' found

  tip: to pass '--ask-for-approval' as a value, use '-- --ask-for-approval'

Usage: codex exec [OPTIONS] [PROMPT]
       codex exec [OPTIONS] <COMMAND> [ARGS]

For more information, try '--help'.


> ⚠ Lens failed to execute. Verdict cannot rely on this lens's coverage; re-run before merging.

  substrate exited with code 2: error: unexpected argument '--ask-for-approval' found

  tip: to pass '--ask-for-approval' as a value, use '-- --ask-for-approval'

Usage: codex exec [OPTIONS] [PROMPT]
       codex exec [OPTIONS] <COMMAND> [ARGS]

For more information, try '--help'.


### Performance — DID NOT RUN
> ⚠ Lens failed to execute. Verdict cannot rely on this lens's coverage; re-run before merging.

- **[important]** `(lens output):0` — **Performance: model deviated from JSON contract**
  substrate exited with code 2: error: unexpected argument '--ask-for-approval' found

  tip: to pass '--ask-for-approval' as a value, use '-- --ask-for-approval'

Usage: codex exec [OPTIONS] [PROMPT]
       codex exec [OPTIONS] <COMMAND> [ARGS]

For more information, try '--help'.


### Maintainability — DID NOT RUN
> ⚠ Lens failed to execute. Verdict cannot rely on this lens's coverage; re-run before merging.

- **[important]** `(lens output):0` — **Maintainability: model deviated from JSON contract**
  substrate exited with code 2: error: unexpected argument '--ask-for-approval' found

  tip: to pass '--ask-for-approval' as a value, use '-- --ask-for-approval'

Usage: codex exec [OPTIONS] [PROMPT]
       codex exec [OPTIONS] <COMMAND> [ARGS]

For more information, try '--help'.



---

jcfischer:

11 finding(s): 1 blocker, 5 important, 4 suggestion, 1 nit.


### CodeQuality
Path guard policy is well-tested for happy paths but the bash parser only inspects the leading command, so trivial compound shells (`cd && rm -rf ~/.soma`) bypass the pi.dev guard. Generated extension also mixes ESM `import type` with `require()` calls, which is fragile across runtimes.

- **[important]** `src/policy-path-guard.ts:118` — **Chain operators bypass destructive-command detection**
  `if (isChainOperator(token)) break;` stops collecting targets but the parser never recurses into the right-hand side of `&&`/`||`/`;`/`|`. A command like `cd /tmp && rm -rf ~/.soma` has `cd` as `mainCommand`, returns empty, and the guard allows it — directly defeating the stated protection on `~/.soma`. The mirror code in `src/adapters/pi-dev-path-guard.ts` (`extractDestructiveTargets`) has the same hole at runtime in pi.dev.
  
  Suggestion:

  ```
  Split tokens on chain operators first, then run destructive-command detection on each segment independently.
  ```

- **[suggestion]** `src/adapters/pi-dev-path-guard.ts:40` — **Generated extension mixes `import type` with `require()`**
  The rendered file begins with `import type { ExtensionAPI }` (ESM) yet repeatedly calls `require("node:path")` inside `isInside`, `isProtected`, `extractDestructiveTargets`. In a strict ESM runtime `require` is undefined and the guard throws on first invocation, silently failing open. Only tested under `bun --eval` (`generated pi.dev guard extension is valid TypeScript`).
  
  Suggestion:

  ```
  Emit a top-level `import { relative, resolve, isAbsolute } from "node:path";` and drop the per-call `require()`s.
  ```

- **[suggestion]** `src/policy-path-guard.ts:29` — **cleanToken redirection strip is asymmetric**
  `cleaned.replace(/^[<>]+/, "").replace(/>$/, "")` strips both `<` and `>` at the start but only `>` at the end, and leaves `<<` here-doc markers untouched. Inputs like `rm -rf <target` lose the leading `<` but `target<` keeps it — token shapes become surprising.
  
  Suggestion:

  ```
  Either strip both characters symmetrically (`/[<>]+$/`) or stop trying to clean redirections inside path tokens — redirections are normally separate shell tokens.
  ```

- **[nit]** `src/policy-audit.ts:29` — **Dummy destinationPath in batch enhancement**
  `withPathGuardForDelete({ destinationPath: process.cwd(), ...checkOptions })` injects a placeholder `destinationPath` only to satisfy the `SomaPolicyCheckOptions` type, then the spread overrides it. Confusing for readers — the value is meaningless yet looks intentional.
  
  Suggestion:

  ```
  Refactor `withPathGuardForDelete` to accept the narrower shape it actually uses (`Pick<SomaPolicyCheckOptions, "action"|"protectedPaths"|"privateRoots">`) so callers don't pass dummy fields.
  ```

### Security
Path-guard parser easy to bypass: shell wrappers (`bash -c`, `eval`, `xargs`, `find -delete`), chained commands (`echo x; rm -rf ~/.soma`), and unexpanded env vars (`$HOME`) all evade detection. Documented default protected paths (~/.claude, ~/.pi, ~/.config/*) are not actually enforced at the `soma policy check` CLI — only the pi-dev runtime extension hardcodes them.

- **[blocker]** `src/policy-path-guard.ts:85` — **Bash parser bypassable via wrappers, chains, env vars**
  Parser only inspects `const mainCommand = tokens[cmdIndex];` after skipping `NON_DESTRUCTIVE_PREFIXES`. `bash -c "rm -rf ~/.soma"`, `sh -c ...`, `eval rm ...`, `xargs rm -rf`, `find ~/.soma -delete`, and chained forms like `echo ok; rm -rf ~/.soma` all return mainCommand != rm/mv and yield zero targets — the guard allows them. `rm -rf $HOME/.soma` also bypasses since tokens are never shell-expanded before matching against `~/.soma`. This breaks the core security invariant the PR advertises.
  
  Suggestion:

  ```
  Recurse into `bash -c`/`sh -c` quoted payloads, split on `;`/`&&`/`||`/`|` and scan every sub-command, perform env-var expansion ($HOME, ${HOME}) before path resolution, and add `find`/`xargs`/`eval` to a wrapper list that re-parses their args.
  ```

- **[important]** `src/policy-audit.ts:5` — **Default protected paths never reach policy CLI**
  `withPathGuardForDelete` builds `protectedPaths: [...(options.protectedPaths ?? []), ...privateRootPaths]` and `evaluateResolvedSomaPathGuard` likewise always supplies a non-empty array, so the `options.protectedPaths ?? SOMA_DEFAULT_PROTECTED_PATHS` fallback in `evaluatePathGuard` is unreachable. `soma policy check --action delete --destination ~/.claude/memory` therefore returns allow, contradicting the PR's documented default set (~/.claude, ~/.pi, ~/.config/cortex, …).
  
  Suggestion:

  ```
  Merge `SOMA_DEFAULT_PROTECTED_PATHS` into the options before calling `evaluatePathGuard`, or have policy.ts seed defaults when no caller-supplied list exists.
  ```

- **[important]** `src/adapters/pi-dev-path-guard.ts:40` — **Generated guard mixes ESM `import type` with `require()` — fail-open risk**
  The emitted module begins with `import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";` yet every path operation calls `require("node:path")`. In a pure-ESM pi.dev runtime, `require` is undefined and the extension throws at first tool_call — silently disabling the guard while the agent keeps running, defeating the protection at exactly the moment it is needed.
  
  Suggestion:

  ```
  Emit a top-level `import { relative, resolve, isAbsolute } from "node:path";` instead of inline `require` calls, or wrap loading errors so a guard-load failure hard-fails the agent rather than fails open.
  ```

### Architecture
Path-guard logic is duplicated between the core `policy-path-guard.ts` module and the pi.dev extension renderer with no shared source of truth for the default protected paths. The core/audit split also double-applies private-root projection.

- **[important]** `src/adapters/pi-dev-path-guard.ts:11` — **Default protected paths hardcoded, diverge from core**
  `renderPathGuardExtension` re-declares the protected-path list as a string literal (`"// Protected paths inherited from Soma path guard policy.", "const PROTECTED_PATHS = [", ... "~/.soma" ... "~/.config/k"`) instead of materializing it from `SOMA_DEFAULT_PROTECTED_PATHS` in `policy-path-guard.ts`. Any future addition to defaults silently skips the pi.dev substrate, defeating the policy's purpose.
  
  Suggestion:

  ```
  Pass `SOMA_DEFAULT_PROTECTED_PATHS` (or the resolved list) into `renderPathGuardExtension` and emit it via `JSON.stringify(paths)` so core remains the single source of truth.
  ```

- **[suggestion]** `src/policy-audit.ts:5` — **Private-root→protectedPath conversion duplicated across layers**
  `withPathGuardForDelete` maps `somaProjectionPrivateRoots(options)` into `protectedPaths`, then `evaluateResolvedSomaPathGuard` in `policy.ts` maps `scope.roots` into protected paths again (`rootProtectedPaths: SomaProtectedPath[] = roots.map(...)`). Same responsibility lives in two modules; future changes to private-root semantics must be kept in sync.
  
  Suggestion:

  ```
  Pick one layer (policy core) to own root→protectedPath projection and drop the audit-side wrapper.
  ```

### Performance
Path guard runs per tool_call but operates on small fixed list (~6 protected paths) with cached require() — sub-millisecond per check. No hot-path, allocation, I/O, or memory growth concerns concrete enough to flag.

_No findings._

### Maintainability
Path-guard logic exists twice: once as a TypeScript module in policy-path-guard.ts and again as hand-built string source inside renderPathGuardExtension. Both copies must be kept in lock-step on every future change.

- **[important]** `src/adapters/pi-dev-path-guard.ts:11` — **Renderer reimplements path-guard runtime logic**
  renderPathGuardExtension hand-emits PROTECTED_PATHS, DESTRUCTIVE_DELETE, DESTRUCTIVE_MOVE, prefix set, expandTilde, isInside, cleanToken, tokenize, and extractDestructiveTargets — all already implemented in src/policy-path-guard.ts (SOMA_DEFAULT_PROTECTED_PATHS, DESTRUCTIVE_COMMANDS, DESTRUCTIVE_MOVE_COMMANDS, NON_DESTRUCTIVE_PREFIXES, expandTilde, findProtectedPath, cleanToken, tokenize, parseBashDestructivePaths). Two copies, no shared source of truth — a future fix to tokenize or the protected-paths list will silently desync the pi.dev extension from the library.
  
  Suggestion:

  ```
  Ship a single runtime file (e.g. src/runtime/path-guard.runtime.ts) read once at build/load time with fs.readFileSync, and have renderPathGuardExtension just wrap it with the pi.on("tool_call", …) boilerplate. SOMA_DEFAULT_PROTECTED_PATHS in particular should be JSON.stringify-injected from the existing constant, not retyped.
  ```

- **[suggestion]** `src/adapters/pi-dev-path-guard.ts:12` — **String-concatenation render is hard to read**
  The 161-line body is built via `"  …" + JSON.stringify(x) + "…"` chains for every literal (e.g. `"if (path === " + JSON.stringify("~") + ") return process.env.HOME || " + JSON.stringify("/") + ";"`). Quoting is fragile and the resulting TS is hard to scan when reviewing logic changes.
  
  Suggestion:

  ```
  Replace the array+join with one template literal and `${JSON.stringify(somaHome)}` placeholders, or — better — inline a runtime asset file as suggested above so the source reads as normal TypeScript.
  ```


---

Issue #112, opened 2026-05-18
soma migrate pai: plan mode should exit 0 on refused-other; apply mode keeps exit 1


jcfischer:
## Symptom

Plan-mode (default, no `--apply`) currently exits non-zero when ANY pack outcomes as `refused-other`:

```
❯ bun soma migrate pai --pai-repo ~/work/PAI
...
soma migrate pai — 1 pack(s) failed with genuine errors:
  - context-search: PAI pack import requires V0 pack file(s): src/SKILL.md (or at least one nested src/<Name>/SKILL.md)

error: script "soma" exited with code 1
```

77 packs imported successfully. 22 collisions handled. 3 reserved-name refusals handled. 1 pack (`context-search`) is malformed in the PAI source (ships `src/commands/` only, no `SKILL.md` anywhere). The plan as a whole is informative; treating it as an error on dry-run is over-loud.

## Per #97 AC-4 design

> Migration exits non-zero if any pack failed, but only AFTER attempting all packs.

This was scoped to the apply path — CI scripts running `migrate pai --apply` need a hard signal that a write phase had a genuine error to triage. Dry-run was not the target of that AC.

## Proposal

Split exit semantics by mode:

| Mode | `refused-other` present | Exit code |
|---|---|---|
| **Plan** (default, no `--apply`) | yes | **0** (informational, footer still highlights the failures) |
| **Plan** | no | 0 |
| **Apply** (`--apply`) | yes | 1 (unchanged — write phase had a genuine error) |
| **Apply** | no | 0 |

Status quo for `--apply` and for the success paths. Only plan-mode-with-failures changes (currently 1 → 0).

The footer line stays — it's still the principal signal that a pack failed. Just not paired with a non-zero exit on dry-run.

## Rationale

- **Interactive use**: `bun soma migrate pai --pai-repo ~/work/PAI` (no apply) is "show me what would happen". A non-zero exit on a known-malformed upstream PAI pack adds friction without informing the human.
- **CI use**: scripts use `--apply` to actually write. That's where exit code matters. Dry-run-in-CI scripts already filter on stdout content, not exit code.
- **No regression**: anything currently checking the dry-run exit code as a gate is implicitly treating "PAI has 1 malformed pack" as a deploy blocker. That's a doctrine miss — the migration is content-preserving and idempotent; a malformed upstream pack is a known-data condition, not an unrecoverable migration failure.

## Acceptance criteria

[...]

Issue #102, opened 2026-05-18
soma migrate pai plan-mode still aborts on substrate-specific packs (#97 fix was apply-only)


jcfischer:
## Symptom

After #97 (PR #99) landed, `soma migrate pai --apply` correctly logs-and-continues on substrate-specific packs. But the **plan path** (no `--apply`) still aborts on the first such pack:

```
❯ bun soma migrate pai --pai-repo ~/work/PAI
$ bun src/cli.ts migrate pai --pai-repo /Users/fischer/work/PAI
PAI pack import refused substrate-specific file(s) without --include-substrate-specific: src/Foundation.md, src/MethodSelection.md
error: script "soma" exited with code 1
```

Plan mode is the user's *first* contact with the orchestrator. If it explodes here, they never see the apply path.

## Root cause

`src/pai-migration.ts:653-657` (`planPaiMigration`) calls `planPaiPackImport` directly via `runBoundedConcurrent`:

```ts
await runBoundedConcurrent(
  packPaths,
  (paiPackDir) => planPaiPackImport({ homeDir: options.homeDir, paiPackDir, somaHome }),
  4,
);
```

No per-pack try/catch. `planPaiPackImport` shares the same `runPaiPackImport` body that throws `PaiPackSubstrateSpecificRefusal` at `src/pai-pack-importer.ts:431` whenever a pack has substrate-specific files and `includeSubstrateSpecific` is false. The throw escapes the concurrency primitive and aborts the whole plan.

The `importOnePackWithOutcome` wrapper added in #97 (apply path) catches this refusal and converts it to a structured `PaiPackOutcome`. Plan path was not similarly wrapped.

## Fix

Add a `planOnePackWithOutcome` (or extend the existing wrapper) for the plan path:

- Call `planPaiPackImport` inside a try/catch.
- Convert `PaiPackSubstrateSpecificRefusal` → `refused-substrate-specific` outcome.
- Convert `Reserved-Soma-skill` errors → `refused-reserved` outcome (plan-side variant; currently the reserved check fires in `runPaiPackImport` too).
- Convert anything else → `refused-other` outcome.
- Propagate `includeSubstrateSpecific` from `options` so that the user's `--include-substrate-specific` flag works in plan mode the same way it works in apply mode.

The plan output should include the same per-pack outcome table that apply emits (per #97 AC-3), so the user sees up-front which packs will be skipped and why.

## Acceptance criteria

[...]

PR #119, opened 2026-05-18
fix(migrate): handle user-owned symlinks gracefully in claude-skills migrate (#118)

### LINKED ISSUE #118: soma migrate claude-skills: abort on user-owned symlink inside skill subtree (#115 follow-up)
## Symptom

\`\`\`
❯ bun soma migrate claude-skills --from ~/.claude/skills
$ bun src/cli.ts migrate claude-skills --from /Users/fischer/.claude/skills
soma migrate claude-skills refused symlink path: Accounting
error: script "soma" exited with code 1
\`\`\`

The migration aborts on the first non-editor-config symlink encountered inside any skill subtree. Real-world user has \`~/.claude/skills/Business/Accounting\` → \`~/work/invisible/Accounting\` (a legitimate symlink to a private skill they're developing in another worktree).

## Root cause

\`src/claude-skills-migrator.ts:273\` throws on any symlink that isn't matched by the editor-config denylist (\`.cursor/\`, \`.vscode/\`, etc.). The throw escapes \`collectSkillFiles\` and aborts the entire migrate.

\`\`\`ts
if (entry.isSymbolicLink()) {
  if (matchEditorConfigSymlinkDir(rel)) {
    continue;
  }
  throw new Error(\`soma migrate claude-skills refused symlink path: \${rel}\`);
}
\`\`\`

Two problems:

1. **Hard abort** — same first-pack-poisons-all pattern that #102 / #109 fixed for pai-migration. One symlinked subpath aborts every other skill's import.
2. **Misleading error path** — \`\${rel}\` is relative to the SKILL containing the symlink, not the migrate root. The error says \`refused symlink path: Accounting\` when the actual path is \`Business/Accounting\`. Hard to diagnose.

## Proposal

Two options (pick one or combine):

### A — Skip user-owned symlinks with audit kind (mirrors #104)

If the symlink target resolves within the user's HOME (e.g. \`~/work/...\`, \`~/.claude/...\`, \`~/.soma/...\`), skip with new audit kind \`skipped-user-owned-symlink\` rather than aborting. Other symlinks (e.g. \`/etc/passwd\`, paths outside HOME) still refuse — security boundary preserved.

### B — Per-skill error isolation

Even when a symlink legitimately refuses, the surrounding skill should be classified as \`refused-other\` rather than aborting the whole migrate. Other skills continue. Mirrors #102's log-and-continue pattern.

### Bonus — fix the error path

\`\`\`ts
throw new Error(\`soma migrate claude-skills refused symlink path: \${sourceName}/\${rel}\`);
\`\`\`

Include the source skill name so the principal can locate the symlink without grep.

## Acceptance criteria

- [ ] AC-1: A symlink target resolving within \`$HOME\` is skipped with audit kind \`skipped-user-owned-symlink\` (per-skill audit). The containing skill imports cleanly.
- [ ] AC-2: A symlink target outside \`$HOME\` (e.g. \`/etc/\`, \`/var/\`, \`/tmp/\`) still refuses — security boundary preserved.
- [ ] AC-3: A genuine refusal classifies the skill as \`refused-other\` with the full path in the reason; other skills continue (log-and-continue per #102).
- [ ] AC-4: Error messages include the source skill name (\`<SourceName>/<rel>\`).
- [ ] AC-5: Fixture tests: user-owned symlink (skipped), out-of-home symlink (refused), mixed tree (one symlink-refused + others imported).

## Source

User-reported on first run of \`soma migrate claude-skills --from ~/.claude/skills\` post-#115. \`Business/Accounting\` is a symlink to user's \`~/work/invisible/Accounting\` private skill.

jcfischer:
## Summary

Closes #118.

`soma migrate claude-skills --from ~/.claude/skills` aborted on the first non-editor-config symlink encountered inside any skill subtree. Real-world repro: user's `Business/Accounting` symlinks to `~/work/invisible/Accounting`, a legitimate private skill in another worktree. One symlink poisoned every other skill's import.

The rescope: **symlinks pointing to user-owned content are a normal pattern** (private/work-repo skills symlinked into `~/.claude/skills`). The fix follows them and imports the target bytes, instead of refusing.

## What changed

- **`resolveSymlinkSafely(absPath, homeRealPath, visitedRealPaths)`** — new helper that resolves a symlink via `realpath`, validates the target is inside `$HOME` (and outside a credential denylist: `.ssh/`, `.aws/`, `.gnupg/`, `.kube/`, `.docker/`), and detects cycles. Multi-hop chains can't escape via symlink-to-symlink hops because `realpath` resolves the full chain.
- **File symlinks** contribute their target bytes at the symlink's relpath. **Directory symlinks** are walked recursively; children land under `<symlinkRel>/<childName>`.
- **Out-of-home / broken / cyclic / denylisted targets** throw, which the orchestrator catches and classifies as `refused-other` (per-skill log-and-continue, mirror of #102/#112). Other skills continue.
- **Top-level `<Name>/SKILL.md` symlinks** are now resolved by the walker via the same helper, instead of being refused at listing time.
- **`.git/` inside a followed symlink branch** is silently skipped (worktrees commonly carry `.git/`); inline `.git/` at the skill root still refuses.
- **Error messages include `<SourceName>/<rel>`** so principals can locate the offending path (AC-4).
- **CLI exit-code gating**: apply mode exits 1 when any skill is `refused-other`; plan mode exits 0 (mirror of #112).
- **Portability report** gains a "Followed user-owned symlinks" table when any symlinks were resolved.

## ACs

- [x] **AC-1**: Symlink whose realpath is inside `$HOME` is **followed**; target bytes imported; per-skill `followed-user-owned-symlink` audit recorded.
- [x] **AC-2**: Out-of-home target (`/etc/`, `/var/`, …) refuses.
- [x] **AC-3**: Refusal → `refused-other` per skill; other skills continue.
- [x] **AC-4**: Error messages include `<SourceName>/<rel>`.
- [x] **AC-5**: Fixture coverage — file-symlink follow, dir-symlink recursive walk, top-level SKILL.md symlink, cycle, symlink-then-escape attack, out-of-home + sibling continue, real-world `Business/Accounting` repro, followed-worktree `.git/` silent skip, inline `.git/` still refuses, portability report audit surface.

## Before / after

```
$ bun soma migrate claude-skills --from ~/.claude/skills    # BEFORE
soma migrate claude-skills refused symlink path: Accounting
error: script "soma" exited with code 1
```

```
$ bun soma migrate claude-skills --from ~/.claude/skills    # AFTER
...
  - business [portable] → imported (clean)
  - design-explorer [portable] → refused-other (... refused VCS metadata directory: design-explorer/.git)
...
Totals: 78 imported, 0 skipped-idempotent, 18 skipped-claude-specific, 1 refused-other.
```

## Test plan

- [x] `bun run typecheck` — clean.
- [x] `bun test` — 814 pass / 0 fail (baseline 804, +10).
- [x] Smoke: `bun src/cli.ts migrate claude-skills --from ~/.claude/skills` no longer aborts. User's `Business/Accounting` imports cleanly via followed-symlink.
- [ ] Holly review (pilot+Holly path; Sage daemon offline).

### DIFF
--- a/src/claude-skills-migrator.ts
+++ b/src/claude-skills-migrator.ts
@@ -57,6 +57,7 @@ import { EDITOR_CONFIG_DIRS } from "./pai-pack-noise";
 import { runBoundedConcurrent } from "./internal-concurrency";
 import { verifySubstrateProjection } from "./claude-skills-substrate-verify";
 import type {
+  ClaudeSkillAuditEntry,
   ClaudeSkillOutcome,
   ClaudeSkillPortabilityTag,
   ClaudeSkillSubstrateVerifyResult,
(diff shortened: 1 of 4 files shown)


jcfischer (commented):
## R1 — Holly code review: soma#119

**PR**: fix(migrate): handle user-owned symlinks gracefully in claude-skills migrate (#118)
**Scope**: 4 files, +824/−54, 1 commit

### Verdict: LGTM ✅

The security model is sound: `realpath`-based resolution prevents multi-hop escape, the `$HOME` boundary + denylist anchor is correctly implemented (separator-aware prefix check at line 82 prevents `/home/user-evil` matching `/home/user`), and per-walk cycle detection via `visitedRealPaths` catches dir-level loops that the kernel's `ELOOP` misses. The per-skill log-and-continue pattern correctly mirrors #102/#112 precedent.

### Findings

**S1 — Missing denylist test** (Suggestion)
No test verifies that a symlink targeting e.g. `$HOME/.ssh/id_rsa` is refused via the denylist path. The logic in `resolveSymlinkSafely` is straightforward (first segment check against `HOME_SUBPATH_DENYLIST`), but a regression test would lock it down. One test with a `.ssh` target and a `refused: "denylist"` assertion would cover it.

**S2 — `refused: "outside-home"` overloaded for non-regular files** (Suggestion, low priority)
Line 109: a device file or socket inside `$HOME` gets `refused: "outside-home"` with reason `"not a regular file or directory"`. The reason string disambiguates, but the kind is misleading — it's not actually outside home. A `refused: "unsupported-type"` kind would be more precise. Low priority since the consumer only checks `"refused" in resolved`, not the specific kind.

**N1 — `tag: "portable"` on refused-other outcomes** (Nit)
Line 409: skills that failed to read get `tag: "portable"` because classification never ran. The portability report already swaps in `refusalReason` for refused-other rows (line 477-479), so this doesn't surface to users. But it's a latent inaccuracy if anything ever keys off `tag` for refused outcomes. Documenting this as intentional (or using a sentinel like `"unclassified"`) would prevent future confusion — though adding a tag value is a bigger change.

### What I verified

- `resolveSymlinkSafely` — separator-aware boundary check, denylist anchoring, ENOENT/ELOOP error mapping, per-walk cycle detection seeding, stat-after-resolve correctness (`lstat` on a `realpath`-resolved target is equivalent to `stat`)
- `collectSkillFiles` — `withinFollowedSymlink` flag correctly relaxes within-root check only inside followed branches; VCS metadata dirs silently skipped inside followed branches but still refused at skill root; plain-dir cycle tracking
- `readSourceSkill` — regex-based error message splicing for AC-4 with `(in ${sourceName})` fallback
- `buildPlanCore` — `ReadResult` tagged union for per-skill error isolation; refused-other outcomes get empty `sourceSha` (no manifest entry)
- `cli.ts` — plan mode exits 0 with refused-other; apply mode throws `SomaCliError(1)` on any refused-other count
- `types.ts` — `ClaudeSkillAuditEntry` type, `disposition` union extended, `refusalReason` optional field
- Test coverage — 10 new tests covering file/dir symlink follow, top-level SKILL.md symlink, out-of-home refusal + sibling continues, cycle detection, escape chain, real-world repro, .git handling, portability report audit

Clean PR. S1 (denylist test) is the only item I'd want addressed before merge; S2 and N1 are fine to defer.

jcfischer (commented):
## R2 — Holly code review: soma#119

**Delta**: commit `e059ceb` — adds denylist regression test (S1 from R1)

### Verdict: LGTM ✅ — no new findings

**S1 resolved**: New test `#118 migrateClaudeSkills refuses denylisted home subpath (.ssh) as refused-other` correctly covers the denylist path — plants a fake `~/.ssh/id_rsa`, symlinks into skill tree, asserts `refused-other` disposition + reason contains both source path and "denylist". Locks down the security-critical path against regression.

**S2 + N1**: Deferred per R1 guidance. No concerns.

**R2 summary**: 815/815 tests, all ACs covered, security model verified across both rounds. Ship it.

Issue #139, opened 2026-05-19
soma migrate claude-skills: progress emitter spams append-only lines during concurrent read+classify phase


jcfischer:
## Symptom

After #125 (PR #138 / \`0cfa9ef\`) shipped the progress indicator, the actual output during \`soma migrate claude-skills --from ~/.claude/skills\` is 97 append-only lines, NOT a single overwriting line as the AC contract implied:

\`\`\`
[discovering 97 skill(s) ...]
[2/97] AlexHormoziPitch  [reading + classifying ... read] (2ms)
[3/97] ApertureOscillation  [reading + classifying ... read] (3ms)
[6/97] AutoResearch  [reading + classifying ... read] (2ms)
[7/97] BeCreative  [reading + classifying ... read] (6ms)
[8/97] Blackboard  [reading + classifying ... read] (2ms)
[1/97] Agents  [reading + classifying ... read] (14ms)
...
\`\`\`

Note the out-of-order completion ([2], [3], [6], [7], [8], [1], [4]...) — that's the 4-wide bounded concurrency in \`runBoundedConcurrent\` at \`claude-skills-migrator.ts:814\`.

## Root cause

\`progress.stepComplete\` is called from concurrent workers. With 4 in-flight skills, \`\\r\` overwrite would clobber 3 of 4 lines. The emitter currently falls back to append-only \`\\n\` for the concurrent read+classify phase — which is technically correct but produces 97 lines of noise the user has to scroll past.

The sequential phases (description rewrite, apply write, smoke verify) work fine — those are one-at-a-time, \`\\r\` overwrite is clean. Only the concurrent fast-phase is noisy.

## Proposal

Three-tier progress display per phase:

| Phase | Concurrency | Display |
|---|---|---|
| **Discovery** | 1 | One line: \`[discovering N skill(s) ...]\` |
| **Read + classify** | 4-wide concurrent | **Single rolling line** with \`\\r\`: \`[N/total] processing <latest> + K others...\`. Mutex-protected: only one writer holds the line at a time. On phase complete: replace with summary \`read + classified N skills in <elapsed> (avg <avg>ms, max <max>ms)\` |
| **Description rewrite** | 1 sequential | One line per skill with \`\\r\` overwrite (current behavior — already works) |
| **Apply write** | 4-wide concurrent | Same as read+classify: rolling line + final summary |
| **Smoke verify** | per-substrate sequential | One line per skill per substrate with \`\\r\` overwrite |

### Implementation

[...]

Issue #125, opened 2026-05-18
soma migrate claude-skills: progress indicator for --rewrite-descriptions (LLM calls block silently for minutes)


jcfischer:
## Symptom

\`soma migrate claude-skills --apply --rewrite-descriptions claude\` on the user's real \`~/.claude/skills/\` tree (97 skills, ~10 needing rewrite) blocks silently for minutes. No output until the entire run completes. Principal has no way to tell:

- Is it still progressing?
- Which skill is currently being rewritten?
- How many remain?
- Did the LLM dispatcher hang?

The LLM dispatcher (subscription-billed \`claude\` subprocess, \`codex exec\`, or Pi.dev) takes 5-30s per call. With 10 oversize descriptions on a real tree, that's a 50s-5min black box.

## Proposal

Stream per-skill progress to stderr (so stdout stays parseable for scripts) as each phase fires. Pattern matches \`gh pr\`, \`brew install\`, \`cargo build\` — progress on stderr, results on stdout.

Concrete output shape:

\`\`\`
[1/97]  agent-tool-bootstrap  [classifying ... claude-specific (skipped)]
[2/97]  agents                [classifying ... claude-specific (skipped)]
[3/97]  alex-hormozi-pitch    [classifying ... portable]
[4/97]  apify                 [classifying ... needs-adapt]
[4/97]  apify                 [rewriting description via claude (1318 chars → target 900)... 1.8s → 836 chars]
[5/97]  ar-xiv                [classifying ... needs-adapt]
[6/97]  art                   [classifying ... claude-specific (skipped)]
...
[97/97] zoom-out              [classifying ... portable]

Apply complete: 78 imported (10 description-rewrites took 47.3s total), 18 skipped-claude-specific, 1 refused-other.
\`\`\`

Lines go to **stderr**. The existing \`stdout\` summary stays unchanged so scripts that parse it (or pipe to a file) aren't broken.

### Verbosity levels

| Mode | Behavior |
|---|---|
| **Default** (TTY stderr) | Stream per-skill progress lines (the shape above). Replace the previous line on the same TTY using \`\\r\` for skills that complete in <100ms (idempotent skips, classification-only). Show full new line for slow operations (rewrite, smoke). |
| **\`--quiet\`** | No progress stream. Only the final stdout summary. CI-friendly. |
| **\`--verbose\`** (already exists per #106) | Per-skill progress + full file lists in final stdout. |
[...]

PR #140, opened 2026-05-19
fix(migrate): suppress per-skill progress during concurrent phases (#139 quick patch)

jcfischer:
## Summary

Quick patch for #139: concurrent read+classify phase emits 97 interleaved per-skill lines on a real PAI tree (one per skill). 4-wide fan-out via `runBoundedConcurrent` defeats the #125 `\r`-overwrite (only one of 4 workers can hold the row).

This patch adds explicit concurrent-phase boundaries to the emitter — `beginConcurrentPhase(name, total, concurrency)` and `endConcurrentPhase(name, elapsedMs)`. While a concurrent phase is active, per-skill `step` / `stepComplete` are output no-ops (but still record elapsed times for the end summary's avg/max). The migrator wraps the read+classify loop with begin/end.

**Scope: quick patch only.** The full three-tier rolling-line proposal in #139's body (mutex-protected single rolling line with in-flight count) is deferred as Phase 2 follow-up — Option A (explicit phase boundaries with suppression) is the minimum change that closes the principal-reported UX gap.

Sequential phases (description rewrite, write-after-rewrite, smoke verify per-substrate) are untouched — `step`/`stepComplete` outside a concurrent bracket keeps the #125 `\r`-overwrite behavior.

## Quick patch ACs satisfied

- [x] Concurrent read+classify emits one banner + one summary, NOT 97 lines
- [x] Sequential phases preserve per-skill `\r`-overwrite from #125
- [x] Phase-end summary includes count + total elapsed + avg + max per-step ms
- [x] `--quiet` still suppresses both banner and summary
- [x] Non-TTY behavior preserved (banner/summary are `\n`-terminated)
- [x] stdout summary stays byte-stable (no banner strings leak into `result.outcomes`)

## Deferred (Phase 2 — not in this PR)

The original #139 proposal included a mutex-protected single rolling line with in-flight count display. That's a richer UX (`[N/total] processing <latest> + K others...` updating in place), but requires a render queue + render loop. The quick patch's "begin + suppressed steps + summary" is the minimum change that closes the spam symptom. Phase 2 can build on top of the begin/end boundaries this PR introduces.

## Before / after

Real PAI tree (97 skills):

```
# BEFORE (97 interleaved lines from 4-wide fan-out)
[discovering 97 skill(s) ...]
[2/97] AlexHormoziPitch  [reading + classifying ... read] (2ms)
[3/97] ApertureOscillation  [reading + classifying ... read] (3ms)
[6/97] AutoResearch  [reading + classifying ... read] (1ms)
[7/97] BeCreative  [reading + classifying ... read] (3ms)
[8/97] Blackboard  [reading + classifying ... read] (2ms)
[1/97] Agents  [reading + classifying ... read] (10ms)
... (91 more lines)

# AFTER (3 lines: discovery + begin banner + end summary)
[discovering 97 skill(s) ...]
[read + classify: 97 skills, 4-wide concurrent ...]
[read + classify: 97 skills in 12.7s (avg 18ms, max 522ms)]
```

## Test evidence

867 → 877 (+10 net):

- `test/claude-skills-progress.test.ts` (+8 unit tests): non-TTY banner shape, TTY banner shape (still `\n`-terminated), step/stepComplete no-op while active, end summary shape with elapsed + avg + max, zero-samples edge case (avg 0ms max 0ms not NaN), sequential resumption after `endConcurrentPhase`, `--quiet` suppresses both, TTY non-quiet still suppresses per-skill rows (the #139 regression guard)
- `test/claude-skills-migrator-progress.test.ts` (+2 integration tests): plan brackets read+classify with begin/end; all `stepComplete("reading + classifying", ...)` events fall strictly between begin and end indices

```
877 pass
0 fail
2818 expect() calls
Ran 877 tests across 58 files.
```

`bun run typecheck` clean.

## Files

- `src/claude-skills-progress.ts` — `beginConcurrentPhase` / `endConcurrentPhase` on `ProgressEmitter`; suppression flag inside `createProgressEmitter`; per-step elapsed accumulator for summary avg/max; no-op stubs on `createNoopProgressEmitter`
- `src/claude-skills-migrator.ts` — wrap the read+classify loop (including the sequential classification follow-on, which is part of the same logical phase) with begin/end inside `buildPlanCore`
- `test/claude-skills-progress.test.ts` — 8 new unit tests
- `test/claude-skills-migrator-progress.test.ts` — 2 new integration tests + emitter stub updated with phase methods

## Test plan

- [x] `bun test` — 877 pass
- [x] `bun run typecheck` — clean
- [x] Smoke: `bun src/cli.ts migrate claude-skills --from ~/.claude/skills 2>&1 | head -5` → 3 lines (was 97+)
- [x] Smoke: re-verified stdout summary byte-stable (no `[discovering` or `[read + classify` strings in `result.outcomes`)

Refs #139 (closes the principal-reported UX gap; full three-tier rolling-line proposal deferred as Phase 2).

### DIFF
--- a/src/claude-skills-migrator.ts
+++ b/src/claude-skills-migrator.ts
@@ -811,6 +811,29 @@ async function buildPlanCore(
   // emitter handles TTY-vs-non-TTY mechanics; we only feed it
   // skill + phase + outcome detail.
   const nameToIndex = new Map<string, number>(names.map((n, i) => [n, i + 1]));
+  // #139 — bracket the read+classify phase with
+  // beginConcurrentPhase / endConcurrentPhase. The emitter suppresses
+  // per-skill output during the bracket and emits one banner + one
+  // summary, replacing the #125 behavior of 97 interleaved append-
+  // only lines (4-wide concurrent fan-out produced one stepComplete
+  // per skill, which the #125 emitter rendered as 97 `\n`-terminated
+  // lines because `\r`-overwrite would clobber 3 of 4 workers).
+  //
+  // The bracket extends across BOTH the concurrent read fan-out AND
+  // the sequential classification loop below — the `classified`
+  // stepComplete is conceptually part of the same "read + classify"
+  // phase (the principal sees one phase, not two), and emitting
+  // those 97 lines append-only would re-introduce the same #139
+  // symptom we're fixing.
+  //
+  // The migrator's own per-phase wall-clock timing
+  // (`readClassifyMs`) is computed by the outer caller around
+  // `buildPlanCore` and remains the source of truth for the
+  // stdout Timing block. The `concurrentStart` delta below is only
+  // used to feed the emitter's banner/summary line.
+  const READ_CLASSIFY_CONCURRENCY = 4;
+  const concurrentStart = Date.now();
+  progress.beginConcurrentPhase("read + classify", names.length, READ_CLASSIFY_CONCURRENCY);
   const readResults: ReadResult[] = await runBoundedConcurrent<string, ReadResult>(
     names,
     async (name) => {
@@ -826,7 +849,7 @@ async function buildPlanCore(
         return { ok: false, sourceName: name, reason };
       }
     },
-    4,
+    READ_CLASSIFY_CONCURRENCY,
   );
 
   const prevBySource = new Map<string, ClaudeSkillsMigrationManifestEntry>();
@@ -901,6 +924,12 @@ async function buildPlanCore(
     });
   }
   outcomes.sort((a, b) => a.sourceName.localeCompare(b.sourceName));
+  // #139 — close the concurrent phase here, AFTER the sequential
+  // classification loop. See the matching beginConcurrentPhase above:
+  // the bracket spans the full logical read+classify phase, not
+  // just the runBoundedConcurrent fan-out, so the per-skill
+  // `classified` stepComplete events stay suppressed too.
+  progress.endConcurrentPhase("read + classify", Date.now() - concurrentStart);
 
   return { isFlatSkillsTree: true, outcomes, reads };
 }
--- a/src/claude-skills-progress.ts
+++ b/src/claude-skills-progress.ts
@@ -87,6 +87,51 @@ export interface ProgressEmitter {
   ): void;
 
   /**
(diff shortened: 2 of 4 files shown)


jcfischer (commented):
## Holly Review — soma#140

**Verdict: APPROVE** _(can't formally approve own-org PR — this is my sign-off)_

Clean, well-scoped quick patch. The explicit phase-boundary approach (Option A from #139) is the right minimum change — it replaces 97 interleaved lines with one banner + one summary while preserving sequential-phase behavior from #125.

### What I checked

**Correctness:**
- `activeConcurrentPhase` state is set before the `quiet` early-return in `beginConcurrentPhase`, so suppression logic is active even in quiet mode. Symmetric clearing in `endConcurrentPhase`. Correct.
- `stepComplete` records `elapsedSamples` while inside the bracket (for avg/max summary) but only when not quiet — consistent because `endConcurrentPhase` also skips the summary under quiet. No divergent state paths.
- `endConcurrentPhase` without matching `begin` is handled defensively (`!phase` guard after clearing state). Won't crash on misuse.
- Zero-samples edge case produces `avg 0ms, max 0ms`, not NaN. Tested.
- The bracket spans both the concurrent read fan-out AND the sequential classification loop, which is correct — both produce per-skill lines that would re-introduce the #139 symptom.

**Design:**
- `READ_CLASSIFY_CONCURRENCY` extraction eliminates the duplicated magic `4` between `runBoundedConcurrent` and the phase banner. Good.
- Banner uses `\n` termination on both TTY and non-TTY — correct since no subsequent step will overwrite it (steps are suppressed).
- Summary uses `phase.total` (from begin call) for the count, not `samples.length`. This means the summary always shows the full skill count regardless of how many hit `stepComplete`. Correct UX.

**Tests (10 new, 877 total):**
- 8 unit tests cover non-TTY banner, TTY banner, suppression, summary shape, zero-samples, sequential resumption, quiet, and TTY regression guard.
- 2 integration tests verify the migrator brackets read+classify with begin/end and that all stepComplete events fall strictly within the bracket indices.
- Coverage is thorough for the surface area.

**No issues found.** Phase 2 (mutex-protected rolling line with in-flight count) is a clean follow-up on top of the begin/end boundaries this PR introduces.

PR #167, opened 2026-05-20
Fix top-level symlinked Claude skill discovery

### LINKED ISSUE #166: migrate claude-skills skips top-level symlinked skill directories
## Problem

`pilot-review-loop` exists in the Claude skills home but is not imported into Soma because it is a top-level symlinked skill directory.

```text
/Users/fischer/.claude/skills/pilot-review-loop -> /Users/fischer/work/mf/pilot/skill
```

The target contains a valid `SKILL.md`, but `soma migrate claude-skills --from ~/.claude/skills --dry-run` does not discover or classify it at all. It is absent from the migration status and absent from the dry-run outcome groups, so it is not imported, skipped as Claude-specific, or refused with an actionable reason.

## Root cause

`src/claude-skills-migrator.ts` `listFlatSkillNames()` only includes direct children where `Dirent.isDirectory()` is true:

```ts
const entries = await readdir(fromDir, { withFileTypes: true });
for (const entry of entries) {
  if (!entry.isDirectory()) continue;
  if (entry.name.startsWith(".")) continue;
  const skillMdPath = join(fromDir, entry.name, "SKILL.md");
  if (await pathExists(skillMdPath)) {
    names.push(entry.name);
  }
}
```

For a top-level symlink to a directory, `entry.isDirectory()` is false, so the skill is silently skipped before `readSourceSkill()` / `collectSkillFiles()` can apply the existing safe symlink handling.

The migrator already has careful symlink support for symlinks *inside* a discovered skill tree (#118), but discovery itself does not support a symlinked top-level skill directory.

## Evidence

- `find ~/.claude/skills -maxdepth 3 -iname "*pilot*"` returns `~/.claude/skills/pilot-review-loop`.
- `ls -la ~/.claude/skills/pilot-review-loop` shows it is a symlink to `/Users/fischer/work/mf/pilot/skill`.
- `~/.claude/skills/pilot-review-loop/SKILL.md` exists and has valid frontmatter (`name: pilot-review-loop`, description present).
- `soma migrate claude-skills --status --from /Users/fischer/.claude/skills` reports 78 imported skills and does not list `pilot-review-loop`.
- `soma migrate claude-skills --from /Users/fischer/.claude/skills --dry-run` discovers 97 skills but does not mention `pilot-review-loop` in any group.

## Desired behavior

Top-level symlinked skill directories should be handled deliberately:

- If the symlink resolves inside the allowed user-owned boundary and the target contains `SKILL.md`, include it as a normal source skill using the symlink basename as `sourceName`.
- Reuse the existing safe symlink policy: broken, cyclic, out-of-home, or denylisted targets should produce a per-skill `refused-other` outcome instead of disappearing.
- The imported skill should preserve the top-level symlink name (`pilot-review-loop`) as the source skill identity unless the source frontmatter requires a more specific normalized name.

## Acceptance criteria

- `listFlatSkillNames()` or the discovery equivalent includes top-level symlinked directories when safe.
- A test fixture with `<from>/pilot-review-loop -> <home>/work/pilot/skill` and target `SKILL.md` appears in the dry-run plan.
- Unsafe top-level symlinks are surfaced as per-skill `refused-other` outcomes, not silently skipped.
- Existing in-tree symlink behavior from #118 remains unchanged.
- `soma migrate claude-skills --from ~/.claude/skills --dry-run` reports `pilot-review-loop` in an outcome group on this machine.

jcfischer:
## Summary
- include top-level symlinked Claude skill directories in flat skill discovery when they expose SKILL.md
- validate symlinked skill roots through the existing safe symlink resolver before walking files
- add #166 regression tests for pilot-review-loop style symlinked skill roots and unsafe out-of-home skill roots

## Verification
- bun test test/claude-skills-migrator.test.ts
- bun run soma migrate claude-skills --from /Users/fischer/.claude/skills --dry-run
- bun run typecheck
- bun test
- git diff --check

Closes #166

### DIFF
--- a/src/claude-skills-migrator.ts
+++ b/src/claude-skills-migrator.ts
@@ -678,6 +678,20 @@ async function readSourceSkill(
   if (!(await pathExists(skillMdPath))) {
     throw new Error(`soma migrate claude-skills: ${sourceName}/SKILL.md not found.`);
   }
+  const rootStat = await lstat(skillDir);
+  if (rootStat.isSymbolicLink()) {
+    const resolvedRoot = await resolveSymlinkSafely(skillDir, homeRealPath, new Set());
+    if ("refused" in resolvedRoot) {
+      throw new Error(
+        `soma migrate claude-skills refused symlink path: ${sourceName} — ${resolvedRoot.reason}`,
+      );
+    }
+    if (resolvedRoot.kind !== "dir") {
+      throw new Error(
+        `soma migrate claude-skills refused symlink path: ${sourceName} — top-level skill symlink target is not a directory: ${resolvedRoot.realPath}`,
+      );
+    }
+  }
   // #118 — wrap collect errors with the source skill name so the
   // refusal reason includes `<sourceName>/<rel>` per AC-4. The walker's
   // error messages all follow the shape `soma migrate claude-skills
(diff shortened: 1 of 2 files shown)


jcfischer (commented):

No findings. Sage approves.


### CodeQuality
The change is focused and the added tests cover both successful top-level symlinked skill discovery and per-skill refusal for unsafe targets. Through the CodeQuality lens, I do not see a correctness, clarity, error-handling, edge-case, or idiomatic-style issue that should block or change this PR.

_No findings._

### Architecture
The change keeps discovery in the Claude skills migrator and continues routing symlink validation through the existing safe resolver. From the architecture lens, it does not introduce a new boundary violation or irreversible public surface change.

_No findings._

### Performance
The PR adds one bounded `lstat` per imported skill and one `pathExists` check for top-level symlink entries during flat discovery. I do not see a concrete hot-path, allocation, I/O, or memory-growth regression in this diff.

_No findings._

### Maintainability
The PR keeps the maintainability impact small: the production change is localized, and the tests cover the new discovery shape without introducing broad helper churn. I did not find material duplication, excessive complexity, dead code, or inappropriate abstraction in this diff.

_No findings._


---

PR #209, opened 2026-05-26
Fix telemetry session-end counts

jcfischer:
## Summary
- count only canonical lifecycle.session_end events as session ends
- keep lifecycle.session_end.registry-write-failed as writeback/failure telemetry
- extend observability regression to model the actual failure + final session_end sequence

## Verification
- bun test test/observability.test.ts test/cli.test.ts --grep "telemetry|summarizes"
- bunx eslint src/observability.ts test/observability.test.ts
- bun run typecheck
- bun test

### DIFF
--- a/src/observability.ts
+++ b/src/observability.ts
@@ -233,7 +233,7 @@ function isSessionStartEvent(event: SomaMemoryEvent): boolean {
 }
 
 function isSessionEndEvent(event: SomaMemoryEvent): boolean {
-  return event.kind === "lifecycle.session_end" || event.kind.startsWith("lifecycle.session_end.");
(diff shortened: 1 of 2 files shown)


jcfischer (commented):

No findings. Sage approves.


### CodeQuality
The change is narrowly scoped and matches the described behavior: only exact `lifecycle.session_end` events count as session ends. The updated regression covers the failure telemetry plus canonical end sequence, and I found no CodeQuality issues in the diff.

_No findings._

### Architecture
The change stays inside the observability aggregation boundary and narrows an internal classifier without altering public surface or dependency direction. I do not see an architecture finding under the requested triggers.

_No findings._


---

PR #261, opened 2026-05-29
Fix #250: add inbound content security

### LINKED ISSUE #250: Integrate content-filter as Soma inbound-content security layer
> *This was generated by AI during triage.*

## Context

Soma is moving toward a Soma-native security system that eventually replaces PAI-specific hooks while preserving the useful security model. PAI security hooks should not be copied verbatim into Soma; the better direction is a portable Soma security core projected into each substrate.

`the-metafactory/content-filter` is a public repository and should be treated as a first-class upstream Meta Factory component. It will also be released through Meta Factory, so Soma can integrate it as a package/dependency rather than vendoring PAI-local hook code.

Repository: https://github.com/the-metafactory/content-filter

## Goal

Integrate `@metafactory/content-filter` as Soma's inbound-content security layer:

- scan externally sourced content before an assistant reads or consumes it
- route acquired external artifacts into a configured sandbox/quarantine root where the substrate supports enforcement
- write security/audit decisions into Soma memory instead of PAI-local memory
- project the strongest available enforcement surface into each supported substrate

## Proposed Design

Add a Soma-owned security layer with at least these conceptual modules:

- `SomaPolicy`: existing private/protected path and destructive-operation policy
- `SomaContentFilter`: wrapper around `@metafactory/content-filter` library APIs such as `filterContent` and `filterContentString`
- `SomaSecurityAudit`: append-only security events under `<soma-home>/memory/SECURITY` and/or `<soma-home>/memory/STATE/events.jsonl`
- `SomaSecurityProjection`: adapter-specific hook/extension/rule projection

Substrate projection targets:

- Claude Code: install `PreToolUse` hooks for acquisition commands and read tools, equivalent to content-filter's `SandboxEnforcer` and `ContentFilter` hooks but Soma-configured.
- Codex: extend the existing policy hook/target extraction path to scan reads from configured external-content roots and block/ask according to the filter decision.
- Pi.dev: enforce inside the existing extension/tool-call guard path where reads or external artifacts are routed through Soma.
- Cursor: project policy/rules first; enforce only where Cursor exposes a reliable hook/MCP/tool gate.
- Cortex/Myelin: enforce at artifact/task ingress before work is handed to an agent.

## Acceptance Criteria

- [ ] Soma has a documented inbound-content security model that distinguishes trusted workspace content from externally sourced content.
- [ ] Soma can configure one or more external-content sandbox roots under the Soma home or policy config.
- [ ] Soma exposes a CLI/API for scanning file and string content using `@metafactory/content-filter`.
- [ ] Scan decisions are represented in Soma-native types and include at least `ALLOWED`, `BLOCKED`, and `HUMAN_REVIEW`.
- [ ] Security/audit events are written through Soma writeback/memory contracts, not PAI-local paths.
- [ ] Claude Code projection installs Soma-owned content-filter hooks without overwriting user/PAI/non-Soma hooks.
- [ ] Codex projection reuses the existing policy hook machinery rather than introducing an unrelated second hook system.
- [ ] Tests cover clean content, blocked content, human-review content, audit/writeback behavior, and at least one substrate projection.
- [ ] Docs explain how this relates to PAI security hooks and why this is a Soma-native port rather than a direct PAI hook copy.

## Non-Goals

- Do not fork PAI security hooks into Soma unchanged.
- Do not make all substrates claim equal enforcement if the substrate only supports advisory projection.
- Do not require a daemon for the first integration unless a substrate genuinely needs one.
- Do not mirror raw external content into Soma memory by default.

## Notes

`content-filter` currently has Claude Code hook entrypoints and a library surface. The integration should prefer the library surface for Soma core behavior, then render substrate-specific hook wrappers from Soma adapters.

jcfischer:
Closes #250

## Summary
- Add Soma-owned inbound content security types, deterministic scanner boundary, hash-bound scan/promote API, and private SECURITY traces plus STATE events.
- Extend soma policy with scan/promote commands and document the Soma-native inbound security model.
- Project Codex inbound read enforcement through the existing PreToolUse hook for <soma-home>/memory/RAW/untrusted/ reads.

## Package gate
- Checked npm for @metafactory/content-filter; registry returned E404, so this PR does not vendor or GitHub-pin it. The production package adapter remains pending release behind the InboundContentScanner boundary.

## Verification
- bun test
- bun run typecheck
- bun run lint (fails on existing repo-wide lint baseline; this branch does not introduce new inbound-security lint findings, but the current lint config reports existing errors across the repo/worktrees)

### DIFF
--- a/docs/inbound-content-security.md
+++ b/docs/inbound-content-security.md
@@ -1,0 +1,103 @@
+# Inbound Content Security
+
+Soma treats externally sourced content as untrusted until Policy records a
+scanner-backed decision for the exact bytes being read into context.
+
+This is a Soma-native port of the useful PAI/content-filter security model, not
+a direct copy of PAI hooks. Soma owns the Policy vocabulary, audit/writeback
+contract, security trace format, and substrate projection. Scanner packages
+provide evidence behind the `InboundContentScanner` interface.
+
+## Model
+
+The default untrusted root is:
+
+```text
+<soma-home>/memory/RAW/untrusted/
+```
+
+Inbound security has two gates:
+
+- acquisition gate: routes external bytes into an untrusted root where a
+  substrate can enforce or encourage that routing
+- context-entry gate: scans exact bytes before they enter model context
+
+The first implementation slice ships the context-entry gate. Acquisition
+routing is adapter-specific follow-up work.
+
+## Decisions
+
+Soma normalizes scanner output into three decisions:
+
+| Decision | Meaning |
+| --- | --- |
+| `ALLOWED` | Content may enter context for this hash-bound scan result. |
+| `BLOCKED` | Content must not enter context. |
+| `HUMAN_REVIEW` | Content is ambiguous and blocks context entry until a future review/override flow exists. |
+
+Allowed content is represented by a content reference:
+
+```json
+{
+  "algorithm": "sha256",
+  "hash": "<content-hash>"
+}
+```
+
+The source path is never permanently trusted by itself. If the file changes, it
+gets a different hash and needs a new decision.
+
+## Audit
+
+`soma policy scan` writes bounded observability to:
+
+```text
+<soma-home>/memory/STATE/events.jsonl
+```
+
+It writes richer private traces to:
+
+```text
+<soma-home>/memory/SECURITY/inbound-content/
+```
+
+Traces contain the decision, reason, scanner id, source metadata, content hash,
+and findings. They do not mirror raw external content by default.
+
+## CLI
+
+Scan a file:
+
+```bash
+bun run soma policy scan --path <path> --json
+```
+
+Scan a string:
+
+```bash
+bun run soma policy scan --content "..." --json
+```
+
+Promote an allowed file reference:
+
+```bash
+bun run soma policy promote --path <path> --json
+```
+
+`promote` fails unless the scan decision is `ALLOWED`.
+
+## Scanner Dependency
+
+`@metafactory/content-filter` is the intended production scanner adapter, but
+it is not yet published on npm as of the #250 implementation. Soma therefore
+ships the interface and deterministic fallback scanner now, without vendoring
+or GitHub-pinning the package. When the package is released, it should implement
+`InboundContentScanner` behind the existing boundary.
+
+## Codex Projection
+
+Codex gets the first enforceable projection because Soma already has a
+`PreToolUse` hook path. The Codex home projection now includes inbound security
+config in `hooks/soma-lifecycle.config.json`; `Read` calls under
+`<soma-home>/memory/RAW/untrusted/` invoke `soma policy scan`. `BLOCKED` and
+`HUMAN_REVIEW` deny the read.
--- a/docs/private-source-guard-v0.md
+++ b/docs/private-source-guard-v0.md
@@ -98,7 +98,8 @@ calls into `~/.soma/isa/*.md` and `~/.soma/memory/*/` pass while
 ## Codex Projection
 
 The Codex home projection installs a `PreToolUse` hook for `Write`, `Edit`,
-`MultiEdit`, `apply_patch`, and shell tools. The hook calls:
+`MultiEdit`, `apply_patch`, shell tools, and inbound-content `Read` checks. The
+private-source write path calls:
 
 ```bash
 SOMA_POLICY_TARGETS='[{"filePath":"<path>","content":"<content>","action":"write"}]' \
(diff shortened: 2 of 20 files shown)


jcfischer (commented):
## Review: inbound content security (#261)

**Architecture: clean.** Hash-bound content references, deterministic scanner boundary, fail-closed hook enforcement, private SECURITY traces + normalized STATE events. Two-gate model (acquisition + context-entry) with context-entry shipped first is the right phasing.

**Findings:**

### 1. Path traversal in `isInsideInboundRoot` (medium)

`codex-hook-entry.mjs:185` — after the `JSON.parse` succeeds, if `scan.decision` is `"ALLOWED"`, execution falls through silently — good. But `scan` could be `undefined` if `JSON.parse` returns `null`/`undefined` for edge-case JSON like `"null"`. In practice `soma policy scan --json` always emits an object, but the hook is a security boundary — consider adding a guard:

```js
if (!scan || typeof scan.decision !== "string") {
  denyPreToolUse(`Soma inbound content scan returned unexpected structure: ${output}`);
}
```

### 2. Duplicate `isInsideInboundRoot` (low, DRY)

Same function implemented in three places:
- `codex-policy-targets.mjs` (plain JS hook asset)
- `src/adapters/codex/policy-targets.ts` (typed TS)
- `src/inbound-security.ts` → `isInboundUntrustedPath`

The `.mjs` duplication is necessary (hook runs standalone). But `policy-targets.ts` could import from `inbound-security.ts` instead of reimplementing. Not blocking.

### 3. `inboundContentHash` uses string encoding (note)

`createHash("sha256").update(content)` uses default UTF-8 string encoding. If the same file is later hashed as bytes (e.g., from a binary scanner adapter), the hashes won't match. Worth documenting that the hash contract is "UTF-8 string content", or switching to `readFile` → `Buffer` → hash to be encoding-neutral. The deterministic scanner only handles text, so this is fine for v0.

### 4. Race between scan and read (design note)

TOCTOU: file content could change between `soma policy scan --path <file>` and the actual `Read` tool execution. The hash-bound model documents this ("if the file changes, it gets a different hash and needs a new decision") but in the hook flow, the scan runs, returns ALLOWED, then Codex reads the file — if content changed in between, the hash in the audit trail doesn't match what entered context. Not exploitable in practice (single-user system, content sits in `untrusted/`), but worth noting for future multi-agent scenarios.

### 5. Test coverage: good

- Deterministic scanner blocked/allowed/human-review paths tested
- CLI scan + promote tested with real temp homes
- Hook integration test confirms deny for hostile content
- Promotion rejection for blocked content tested
- Home projection bundle assertions updated

### 6. Trace doesn't include raw content — correct

`writeInboundTrace` writes decision + hash + findings but not the raw content itself. Right call for a security trace.

**Verdict: approve with optional hardening.** Finding #1 (guard against malformed scan output structure) is worth fixing before merge. Rest are notes for follow-up.

---

jcfischer (commented):
## Review: inbound content security (#261) — Ivy

**Architecture: solid.** Hash-bound content references, deterministic scanner boundary, fail-closed hook enforcement, private SECURITY traces + STATE events. Two-gate model with context-entry shipped first is correct phasing.

**Prior review finding #1 (malformed scan output guard): resolved** in commit `aa2ec10`. `codex-hook-entry.mjs` now checks `!scan || typeof scan !== "object" || typeof scan.decision !== "string"` before trusting the decision field.

### Findings (all non-blocking)

| # | Severity | Finding |
|---|----------|---------|
| 1 | low/DRY | `isInsideInboundRoot` triplicated across `.mjs`, `policy-targets.ts`, `inbound-security.ts`. The `.mjs` copy is unavoidable (standalone hook); the two TS copies could share. |
| 2 | note | `inboundContentHash` uses default UTF-8 string encoding. Document this contract for when binary scanner adapters arrive. Fine for v0. |
| 3 | note | TOCTOU between scan and read — acknowledged in docs, not exploitable in single-user context. |
| 4 | nit | `normalizeDecision` silently maps unknown decisions → `HUMAN_REVIEW`. Defensively correct (fail-closed), but a stderr warning would help debugging typo'd scanner returns. |
| 5 | nit | `parsePolicyArgs` error on `command !== "policy"` shows check-specific usage instead of general usage string. |

**Test coverage:** deterministic scanner paths (blocked/allowed/human-review), CLI scan + promote, hook integration deny, promotion rejection, home projection bundle assertions, codex inbound target extraction. Good.

**Verdict:** blockers=0 majors=0 nits=2 — recommend: approve. Clean implementation, prior finding addressed, all remaining notes are follow-up material.

PR #275, opened 2026-05-30
Fix #243: harden direct path guard targets

### LINKED ISSUE #243: Path guard allows `rm -rf ~/.claude/memory` (delete honors modify-only allowed subpaths)
## Summary

`evaluatePathGuard` does **not** block a `delete` action on `~/.claude/memory`,
because `memory` is listed in `CLAUDE_HOME_ALLOWED_MODIFY_SUBPATHS`. The allowed
subpaths are meant to permit **modify** writes only — the `~/.soma` entry's own
comment states the intended invariant: *"Delete remains blocked everywhere under
~/.soma."* The Claude/PAI home does not uphold that invariant.

This is caught by the existing test, which **fails** on `main` @ v0.7.1 (`9fad69a`).

## Reproduction

```
bun test test/policy-path-guard.test.ts
```

```
test/policy-path-guard.test.ts:340
(fail) blocks rm -rf on Claude/PAI home
  evaluatePathGuard({ targetPaths: ["~/.claude/memory"], cwd: "/tmp", action: "delete" })
  expect(result.blocked).toBe(true)   // Expected: true, Received: false
```

Full suite: **1138 pass / 1 fail** (this test).

## Root cause

`src/policy-path-guard.ts`:
- L24 `CLAUDE_HOME_ALLOWED_MODIFY_SUBPATHS = ["memory", "memories", "PAI/MEMORY"]`
- L40 `{ path: "~/.claude", description: "Claude Code / PAI home", allowedSubpaths: CLAUDE_HOME_ALLOWED_MODIFY_SUBPATHS }`

`allowedSubpaths` is applied regardless of `action`. For `~/.soma` the delete
case happens to pass because the test targets a deep file; for `~/.claude` the
target is the allowed subpath root and the guard permits it.

## Impact

`rm -rf ~/.claude/memory` (and `~/.claude/memories`, `~/.claude/PAI/MEMORY`) is
**not** guarded. Security-adjacent — could allow destructive deletion of a
substrate home's memory tree.

## Suggested fix

Make `delete` ignore `allowedSubpaths` (those govern `modify` only), so delete
is blocked everywhere under a protected root — matching the documented `~/.soma`
behavior. Alternatively, if delete-on-memory is intentional, update the test and
the `~/.soma` comment to reflect it.

## Recommendation

**Block the 0.7.1 tag until resolved.**

> Found during an end-to-end 0.7.1 release test (install via `arc` + full source test battery).

jcfischer:
Closes #243\n\n## Summary\n- Normalize direct evaluatePathGuard targets through cwd and tilde expansion before protected-root matching.\n- Preserve normalized matched paths in guard findings.\n- Add regression coverage for literal tilde and relative direct delete targets.\n\n## Verification\n- bun test test/policy-path-guard.test.ts\n- bun test test/policy-targets.test.ts\n- bun run lint\n- bun run typecheck\n- bun test

### DIFF
--- a/src/policy-path-guard.ts
+++ b/src/policy-path-guard.ts
@@ -507,9 +507,10 @@ export function evaluatePathGuard(options: SomaPathGuardOptions): SomaPathGuardR
   const matchedDescriptions: string[] = [];
 
   for (const target of options.targetPaths) {
-    const match = findProtectedPath(target, protectedPaths, options.action, realScopeCache, protectedRootCache);
+    const resolvedTarget = resolvePath(target, options.cwd);
+    const match = findProtectedPath(resolvedTarget, protectedPaths, options.action, realScopeCache, protectedRootCache);
     if (match) {
-      matchedPaths.push(target);
+      matchedPaths.push(resolvedTarget);
       matchedDescriptions.push(match.description ? `${match.path} (${match.description})` : match.path);
     }
   }
--- a/test/policy-path-guard.test.ts
+++ b/test/policy-path-guard.test.ts
@@ -341,6 +341,33 @@ test("blocks rm -rf on Claude/PAI home", () => {
   expect(result.matchedDescriptions[0]).toContain("Claude Code / PAI home");
 });
 
+test("blocks direct tilde delete targets on Claude/PAI home", () => {
+  const claudeMemory = "~/." + "claude/memory";
+  const result = evaluatePathGuard({
+    targetPaths: [claudeMemory],
+    cwd: "/tmp",
+    action: "delete",
+  });
+
+  expect(result.blocked).toBe(true);
+  expect(result.matchedPaths[0]).toContain(".claude/memory");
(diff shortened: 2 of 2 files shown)


jcfischer (commented):
**Lenses:** CodeQuality, Security, Architecture

Clean security fix. `evaluatePathGuard` now normalizes `targetPaths` through `resolvePath(target, options.cwd)` before matching — closes the tilde-literal and relative-path bypass vectors that #243 identified.

**Findings:** 0 blockers, 0 majors, 0 nits.

- Fix is minimal and correct — `resolvePath` already handles tilde expansion + cwd-relative resolution; reusing it here is the right call
- `matchedPaths` now stores the resolved path, consistent with how `findProtectedPath` operates
- Both regression tests (literal tilde target, relative target under custom protected root) directly exercise the two bypass vectors
- String-split `"~/." + "claude/memory"` in the test avoids hook false-positives — understood and fine

Recommend approve. Can't self-approve since Jens-Christian authored it.

PR #254, opened 2026-05-29
feat(algorithm): sync ISA edits into resumable soma runs (hook bridge)

jcfischer:
## What

Automatically mirror PAI Algorithm ISA edits into resumable soma algorithm runs, so a run started on one substrate (Claude Code) is resumable on others (Codex, pi.dev) and its learning can be promoted home. This is the "hook bridge" enabling a cross-substrate Algorithm relay.

## Key design finding

`AlgorithmRun.isa` **is** an `IdealStateArtifact` (`src/types.ts`) — runs already embed an ISA; they are the same model, not parallel ones. So this does **not** create a second artifact. It create-or-resumes a run whose embedded ISA mirrors the PAI ISA file, keyed by the ISA `slug` via a new durable index `memory/STATE/isa-run-index.json` (substrate-agnostic — the cross-host join key).

## How

New verb:
```
soma algorithm sync-from-isa --isa <path> --substrate <id> [--soma-home <dir>] [--promote-on-complete]
```
- Parses the ISA (`parseIsa`), keys the run by `slug`, create-or-resume via the durable index.
- **Reconcile-then-advance**: `[x]` ISCs → `passed` first (so the LEARN gate passes), then advances **forward only** to the reachable phase, auto-satisfying gates (synthetic capability / plan step / build→execute `change` / step-done).
- Records a `learn` entry at learn/complete; `--promote-on-complete` reuses `promoteAlgorithmRunMemory`.
- **Idempotent + failure-isolated**: unchanged ISA → no-op; malformed/missing path → `{noop:true}`, never throws.

Hook wiring: the claude-code `PostToolUse` (`Write|Edit|MultiEdit|NotebookEdit`) writeback path now also detects ISA paths (`MEMORY/WORK/<slug>/ISA.md` or project-root `ISA.md`) and spawns `sync-from-isa` detached, after the telemetry queue append, wrapped in try/catch — cannot block or break existing writeback.

## Files

- `src/algorithm-isa-sync.ts` (new) — core sync + slug→runId index
- `src/cli/algorithm.ts` — `sync-from-isa` verb
- `src/index.ts` — re-export
- `src/adapters/claude-code/hook-runner.mjs` — ISA matcher + detached dispatch
- `test/algorithm-isa-sync.test.ts` (new, 12), `test/claude-code-install.test.ts` (+2)

## Gates

- `bun run typecheck` clean
- `bun test` **1158 pass / 0 fail** (+40)
- `bun run check-release-privacy` ok

## Notes / risks

[...]

PR #276, opened 2026-05-30
Fix #260: add sync-from-isa debug signal

### LINKED ISSUE #260: sync-from-isa: minor debug-logging nits (deferred from PR #254)
Non-blocking nits from sage's review of #254 (approved + merged):

1. `algorithm-isa-sync.ts:~113` — empty catch in `readIndex`. Intent is correct (missing/malformed index → start fresh, self-healing). A `process.stderr.write` on the corrupt-index path would aid debugging without breaking the no-throw contract.
2. `claude-code-install.test.ts:349-355` — `waitForRunFile` 80×50ms poll works; `Bun.sleep` would read clearer. Functionally identical.

Both low priority.

jcfischer:
Closes #260\n\n## Summary\n- Emit bounded sync-from-ISA debug output when isa-run-index.json is unreadable or malformed while preserving the no-throw contract.\n- Add corrupt-index coverage proving sync starts fresh and reports a useful debug line.\n- Replace the Claude Code hook-bridge polling helper's manual timeout promise with Bun.sleep.\n\n## Verification\n- bun test test/algorithm-isa-sync.test.ts\n- bun test test/claude-code-install.test.ts\n- bun run lint\n- bun run typecheck\n- bun test

### DIFF
--- a/src/algorithm-isa-sync.ts
+++ b/src/algorithm-isa-sync.ts
@@ -103,18 +103,54 @@ interface IsaRunIndex {
   bySlug: Record<string, string>;
 }
 
-async function readIndex(somaHome: string): Promise<IsaRunIndex> {
+function debugSyncFromIsa(message: string, error?: unknown): void {
   try {
-    const raw = await readFile(indexPath(somaHome), "utf8");
+    const detail = formatDebugError(error);
+    const line = `[soma sync-from-isa] ${message}${detail ? `: ${detail}` : ""}`;
+    process.stderr.write(`${line.slice(0, 300)}\n`);
+  } catch {
+    // Debug output must never compromise hook failure isolation.
+  }
+}
+
+function formatDebugError(error: unknown): string {
+  if (error === undefined) return "";
+  if (error instanceof Error) {
+    const errorWithCode: Error & { code?: unknown } = error;
+    const code = typeof errorWithCode.code === "string" ? ` ${errorWithCode.code}` : "";
+    return `${error.name}${code}: ${error.message}`.slice(0, 180);
+  }
+  if (typeof error === "string") return error.slice(0, 180);
+  if (typeof error === "number" || typeof error === "boolean" || typeof error === "bigint" || typeof error === "symbol") {
+    return String(error).slice(0, 180);
+  }
+  return "non-error thrown";
+}
+
+function isEnoent(error: unknown): boolean {
+  return typeof error === "object" && error !== null && "code" in error && (error as { code?: unknown }).code === "ENOENT";
+}
+
+async function readIndex(somaHome: string): Promise<IsaRunIndex> {
+  let raw: string;
+  try {
+    raw = await readFile(indexPath(somaHome), "utf8");
+  } catch (error) {
+    if (!isEnoent(error)) debugSyncFromIsa("could not read isa-run-index.json; starting fresh", error);
+    return { bySlug: {} };
+  }
+
+  try {
     const parsed: unknown = JSON.parse(raw);
     if (parsed && typeof parsed === "object" && !Array.isArray(parsed) && "bySlug" in parsed) {
-      const bySlug = (parsed).bySlug;
+      const bySlug = parsed.bySlug;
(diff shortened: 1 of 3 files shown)


jcfischer (commented):
## Review — soma#276: sync-from-isa debug signal

**Lenses applied:** CodeQuality, Architecture

### Findings

**0 blockers, 0 majors, 2 nits**

**N1 (nit):** `formatDebugError` — the `number | boolean | bigint | symbol` branch is defensive beyond what real callers produce (thrown values are almost always `Error` or `string`). Fine to keep as belt-and-suspenders, but could collapse to a single `String(error).slice(0, 180)` default branch for the same effect with less surface area. Author's call.

**N2 (nit):** Empty `catch` in `debugSyncFromIsa` — the comment explains the intent well, but per repo convention (CLAUDE.md: "NEVER use empty catch blocks") this should either name the variable `_err` or add a minimal `process.stderr.write` fallback. The comment alone doesn't satisfy the stated convention. Low severity since this IS the debug output path itself (recursion risk is real), so the empty catch is arguably the right call — but the convention says name it `_err`.

### Assessment

Clean, tightly scoped PR. Addresses both nits from sage's review of #254:

1. **Debug signal** — `debugSyncFromIsa` + `formatDebugError` provide bounded (`slice(0, 300)`) stderr output on corrupt/unreadable index without breaking the no-throw hook contract. Error detail extraction is thorough (Error with code, string, primitives). The split of `readIndex` into separate read-catch and parse-catch blocks makes the error paths clearer — ENOENT is silently expected (normal first-run), other read errors and parse errors get debug lines.

2. **Bun.sleep** — Drop-in replacement for `setTimeout` promise wrapper in test helper. Functionally identical, reads cleaner, idiomatic Bun.

3. **Test coverage** — New `"corrupt run index starts fresh and emits bounded debug output"` test writes invalid JSON to the index file, spies on stderr, verifies sync creates a fresh run, and asserts the debug output contains the expected prefix and is under 512 bytes. Good coverage of the new code path.

The `isEnoent` helper is clean and avoids the `NodeJS.ErrnoException` type dependency. `(parsed).bySlug` → `parsed.bySlug` cast cleanup is a nice bonus.

**Verdict:** blockers=0 majors=0 nits=2 — recommend approve. Can't self-approve since you authored it.

jcfischer (commented):
## Review — soma#276: sync-from-isa debug signal

**Lenses applied:** CodeQuality, Architecture

### Findings

**0 blockers, 0 majors, 1 nit**

**N1 (nit — code duplication):** `isEnoent` at `algorithm-isa-sync.ts:131-133` is a character-for-character duplicate of `fs-errors.ts:1-3` (also duplicated in `fs-utils.ts` and `pai-migration.ts`). The canonical export lives at `src/fs-errors.ts` and is already imported by `install.ts`, `hooks.ts`, `writeback.ts`, and `doctor.ts`. This file could import it rather than redeclare it. Not blocking — the function is 2 lines and the file may intentionally minimize imports for hook-isolation reasons.

### Positive observations

- **Bounded output (line 110, 122-127):** `.slice(0, 300)` on the stderr line and `.slice(0, 180)` on error details enforce bounded debug output — exactly what you want from code running inside CC hooks. No log-bomb risk.
- **No-throw contract preserved:** `debugSyncFromIsa` wraps its own body in try/catch (line 111-114), so a formatting bug in the debug helper can never surface as a hook failure. Correct defensive layering.
- **Split try/catch in `readIndex` (lines 136-156):** Separating the file-read error from the JSON-parse error is the right call — the ENOENT path is intentionally silent (file doesn't exist yet = normal), while corrupt-JSON gets a debug signal. Clean control flow.
- **Test is well-scoped (test lines 117-150):** Writes corrupt JSON, spies on stderr, asserts both content and bounded length (`< 512`). Restores the spy in `finally`. Good test hygiene.
- **Bun.sleep swap (claude-code-install.test.ts:355):** Trivial, correct, reads better. Addresses issue #260 item 2.

### Verdict

Clean, tightly-scoped PR that addresses both nits from issue #260. The no-throw contract is maintained, debug output is bounded, and the test proves the corrupt-index recovery path. Would approve — can't approve own PR via GH API.

PR #311, opened 2026-06-13
fix(install): first install lists the ISA skill in skills.md

klittle32:
Fixes #310

## Problem

A first `soma install <substrate>` rendered `skills.md` as "No Soma skills were declared." even though the ISA skill was installed alongside it; the skill only appeared on the second install. `bootstrapSomaHome` captures its context snapshot (which feeds `skills.md`) **before** `installSomaForSubstrate` writes the canonical ISA skill to `<somaHome>/skills/ISA`, so the first projection rendered from an empty skill list and only converged on re-run.

## Fix

Reload the Soma home context after the canonical ISA baseline is written, so the very first projection already lists the ISA skill: **install #1 == install #2**.

One subtlety: reloading puts the ISA skill into `profile.skills`, which would make the generic portable-skill file loop *also* emit the ISA skill's files — on top of its dedicated, managed projection (`installIsaSkillProjection`: baseline tracking, drift detection, `skillNameOverride`). That double-write is redundant (identical bytes) and would break the file-set contract. So the ISA skill is now delegated: a shared `projectableSkillFiles()` filter excludes it from the portable file loop in the adapters that project skill files (codex, pi-dev), while `renderSkills` still lists it in `skills.md`. The ISA skill is projected exactly once, by its dedicated installer.

## Result

- `skills.md` lists the ISA skill from the first install.
- **install #1 == install #2** — same reported file count and a byte-identical `skills.md` across the two installs (asserted by the new regression test below).
- The ISA skill is projected **exactly once**: the portable-skill loop no longer emits its files (asserted in `home-projection.test.ts`), leaving only the dedicated managed projection — no double-write.
- The existing dry-run/apply install tests (e.g. `cli dry-runs and applies pi.dev install`) are untouched and stay green; this PR does not change the dry-run==apply contract.

## Verification

- `bun run typecheck` clean.
- New regression test — *"first install already converges: skills.md lists ISA and the file set matches a re-install"* — fails before the fix, passes after.
- `test/home-projection.test.ts`: the pi-dev normalization test now uses a non-dedicated uppercase skill (FAQ) and asserts the ISA skill is excluded from the portable bundle.
- **A/B failure-set diff vs `upstream/main` on the same machine** (touched test files `install.test.ts` + `home-projection.test.ts`): identical pre-existing failure set, zero regressions, plus exactly one new passing test. Reproducible transcript (commands, environment, output) posted in a comment below.
[...]

PR #313, opened 2026-06-13
fix(export): soma export emits the ISA skill, completing the bundle

### LINKED ISSUE #312: soma export should project the ISA skill through its dedicated installer (complete, portable bundle)
Follow-up to #310. That issue is about *install* rendering skills.md without the ISA files on the first run; this one is the matching gap on the *export* path, which surfaces once the ISA skill is treated as a dedicated (non-portable-loop) projection.

## Desired behavior

`soma export <substrate>` should produce a bundle whose ISA skill file set **matches an installed home byte-for-byte**, for every substrate, with portable (forward-slash) bundle-relative paths.

## Current gap

The ISA skill has a dedicated managed projection at install time (`installIsaSkillProjection` — baseline tracking, drift detection, per-substrate `skillNameOverride` and destination dir). `soma export` runs only the generic projection builders, so the ISA files are not projected the way install projects them:

- Once the ISA skill is excluded from the generic portable-skill loop (the fix in #310/#311), an export bundle **lists the ISA skill in skills.md but omits its files** — an incomplete projection that does not match an installed home.
- The per-substrate destinations that install honors (cursor's `.cursor/rules/soma/skills/ISA`, pi-dev's custom dir + lowercase `isa` skill name) are not reflected, so even where files do land they don't match an installed home.

## Repro (on top of #311)

```sh
soma install codex --apply --home-dir "$TMP"
soma export codex --home-dir "$TMP"   # bundle has no skills/ISA/* entries, though skills.md lists ISA
```

## Proposal

Give export an ISA-aware step that projects the ISA skill through the same dedicated installer (as in-memory, forward-slash bundle entries — no disk writes), deriving the destination prefix per-substrate from the install spec. Verified end-to-end on Windows 11 and macOS. PR follows, stacked on #311.

klittle32:
Fixes #312. **Stacked on #311 — draft until that merges.**

> ⚠️ **Depends on #311.** This PR builds on the ISA-exclusion fix in #311, so its branch currently contains **two commits**: #311's (`first install lists the ISA skill in skills.md`) underneath, and this PR's export commit on top. Once #311 merges, GitHub will narrow this PR to the single export commit and I'll mark it ready for review. Reviewing #311 first is the intended order.

## Problem

`soma export <substrate>` ran only the generic projection builders. The ISA skill has a dedicated managed projection at install time (`installIsaSkillProjection` — baseline tracking, drift detection, per-substrate `skillNameOverride` and destination dir) and is excluded from the generic portable-skill loop (#311). So an export bundle listed the ISA skill in `skills.md` but omitted its files — an incomplete projection that does not match an installed home. Every substrate had the same omission.

## Fix

- **`isa-skill-installer.ts`**: new pure `projectIsaSkillBundleFiles` — returns the same content `installIsaSkillProjection` writes, as in-memory `{path, content}[]` with bundle-relative **forward-slash** paths, no disk writes / baseline / drift. Mirrors a fresh install.
- **`substrate-lifecycle.ts`**: `buildExportProjection` appends the ISA projection, deriving the destination prefix per-substrate from the install spec (codex/claude-code `skills/ISA`, cursor `.cursor/rules/soma/skills/ISA`, pi-dev honors its `skillNameOverride`).

## Single content producer (hardening)

Export and the install write loop independently re-listed, re-read, and applied the per-file content transform, so the byte-identity invariant (exported bundle == installed home) rested only on a comment. Extracted one `computeSourceFileEntries(...)` as the single source of truth for ISA file CONTENT: the on-disk install writes, the baseline hashes, and the in-memory export bundle all derive their bytes from it. Behavior-preserving.

## Tests

A per-substrate generated test over `INSTALL_SUBSTRATES` derives each ISA destination prefix from the install spec exactly as `buildExportIsaProjection` does, then asserts the exported bundle's ISA set == the installed home **byte-for-byte**. Covers the non-trivial prefixes (cursor's nested rules dir, pi-dev's custom dir + lowercase skill name) that were otherwise unverified.

## Verification

- `bun run typecheck` clean
- The 4 new per-substrate export tests pass; A/B against the #311 base shows an identical failure set (zero regressions) on the touched test files
- **Platform scope:** verified on Windows 11 (typecheck + the 4 export tests + the A/B above). Export is substrate-neutral shared-core with no OS-specific code paths. Full reproducible evidence (commands, environment, output) will be posted when this flips from draft to ready after #311 merges and the branch narrows to the single export commit.

### DIFF
--- a/src/adapters/codex/adapter.ts
+++ b/src/adapters/codex/adapter.ts
@@ -5,7 +5,7 @@ import { defaultSomaRepoPath } from "../../repo-path";
 import { resolveBunExecutable } from "../../bun-probe";
 import { readCodexHookAsset, renderCodexPolicyHook, renderCodexPolicyTargets } from "./hooks/assets";
 import { renderFeedbackHookModule } from "../shared/feedback-helper";
-import { renderAssistantCore, renderMemoryLayout, renderPolicyProjection, renderSkills } from "../shared";
+import { projectableSkillFiles, renderAssistantCore, renderMemoryLayout, renderPolicyProjection, renderSkills } from "../shared";
 import { activeIsaBundleFile } from "../../adapter-active-isa";
 import { somaPolicyPrivateMarkers } from "../../policy";
 import { somaMemoryPrivateRoots, somaProjectionPrivateRoots } from "../../projection-private-roots";
@@ -443,7 +443,7 @@ export function projectCodex(input: ProjectionInput): Projection {
 
 export function projectCodexHome(input: ProjectionInput, somaHome: string, homeDir?: string, somaRepoPath = defaultSomaRepoPath()): Projection {
   const instructions = renderHomeRules(input, somaHome);
-  const portableSkillFiles = input.profile.skills.flatMap((skill) =>
+  const portableSkillFiles = projectableSkillFiles(input.profile.skills).flatMap((skill) =>
     (skill.files ?? []).map((file) => ({
       path: `skills/${skill.name}/${file.path}`,
       content: rewriteSubstrateProjectionContent({
--- a/src/adapters/codex/adapter.ts
+++ b/src/adapters/codex/adapter.ts
@@ -5,7 +5,7 @@ import { defaultSomaRepoPath } from "../../repo-path";
 import { resolveBunExecutable } from "../../bun-probe";
 import { readCodexHookAsset, renderCodexPolicyHook, renderCodexPolicyTargets } from "./hooks/assets";
 import { renderFeedbackHookModule } from "../shared/feedback-helper";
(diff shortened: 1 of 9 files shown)


jcfischer (changes requested):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `PR description:43` — **Unsubstantiated cross-OS verification claim**
  The description claims "Verified end-to-end on Windows 11 and macOS", but the PR data includes no commands, logs, CI links, environment details, or reproducible evidence for either platform. Downstream reviewers are asked to rely on a broad compatibility claim that cannot be checked from the artifact.


---

Issue #316, opened 2026-06-13
soma install claude-code --apply creates hook commands with an improperly pathed bun


vpzed:
**Version**: soma 0.8.5

**What happened and how to reproduce**:   After running 'soma install claude-code --apply' the updated ~/.claude/settings.json file's hooks array contained command entries with a fully pathed bun value of '/tmp/bun-node-0d9b296af/bun' which is ephemeral. While it will work initially it could break after a reboot.

```
❯ soma install claude-code --apply
Soma install applied
substrate: claude-code
somaHome: /home/metauser/.soma
substrateHome: /home/metauser/.claude

Soma files:
- /home/metauser/.soma/profile/assistant.md
- /home/metauser/.soma/profile/principal.md
- /home/metauser/.soma/profile/telos.md
- /home/metauser/.soma/policy/README.md
- /home/metauser/.soma/skills/README.md
- /home/metauser/.soma/projections/README.md
- /home/metauser/.soma/isa/INDEX.md

Substrate files:
- /home/metauser/.claude/rules/soma/README.md
- /home/metauser/.claude/rules/soma/CONTEXT.md
- /home/metauser/.claude/rules/soma/PROFILE.md
- /home/metauser/.claude/rules/soma/TELOS.md
- /home/metauser/.claude/rules/soma/MEMORY_LAYOUT.md
- /home/metauser/.claude/rules/soma/SKILLS.md
- /home/metauser/.claude/rules/soma/POLICY.md
- /home/metauser/.claude/hooks/soma/soma-claude-code-hook.mjs
- /home/metauser/.claude/hooks/soma/soma-claude-code-hook.config.json
- /home/metauser/.claude/settings.json

❯ head -16  ~/.claude/settings.json | tail -13
  "hooks": {
    "SessionStart": [
      {
        "description": "Soma: Start Soma lifecycle session",
        "hooks": [
          {
            "type": "command",
            "command": "'/tmp/bun-node-0d9b296af/bun' '/home/metauser/.claude/hooks/soma/soma-claude-code-hook.mjs' session-start",
            "timeout": 30
          }
        ]
      }
    ],
```

**What was expected**: All soma created hook 'command' entries would use unpathed 'bun' so they use the user's PATH variable to locate bun.

jcfischer:
[...]

Issue #236, opened 2026-05-27
Claude Code install should wire Soma lifecycle and writeback hooks


jcfischer:
## Problem

`soma install claude-code --apply` currently projects Soma context into Claude Code, but it does not install or enable Soma lifecycle/writeback hooks.

That means Claude Code can be Soma-contextual without being Soma-observable:

- Claude/PAI Native Mode may write rich activity to its substrate-local memory.
- Claude Code transcripts may show active work and native hook execution.
- Soma shared memory may still receive no `claude-code` lifecycle or writeback events.

This breaks the expected cross-substrate contract: another Soma substrate cannot reliably discover that a Claude Code session happened, what session id it had, or which durable artifacts/events it produced.

## Needed Behavior

Claude Code install should wire lifecycle/writeback hooks, not only context projection.

At minimum, the Claude Code adapter should install hooks that:

1. On Claude `SessionStart`, call Soma lifecycle start:
   - `soma lifecycle session-start --substrate claude-code --session-id <claude-session-id>`
   - append `lifecycle.session_start` to the Soma event log
   - refresh projected startup context if that is part of the adapter contract

2. On Claude `SessionEnd`, call Soma lifecycle end:
   - `soma lifecycle session-end --substrate claude-code --session-id <claude-session-id>`
   - update shared work registry state where applicable
   - append `lifecycle.session_end` with metadata and artifact pointers

3. On relevant Claude writeback moments, append metadata-only Soma events through the writeback gate:
   - tool activity/result summaries where intentionally supported
   - created/changed artifact pointers
   - subagent start/stop summaries if exposed by Claude hooks
   - learning/result capture events only when they pass existing policy gates

4. Remain safe and idempotent:
   - do not mirror full raw transcripts or private prompts by default
   - do not duplicate events when hooks rerun
   - do not overwrite existing non-Soma Claude settings
   - fail soft: hook failure must not block normal Claude Code usage

## Implementation Notes

- Existing Claude projection code lives around `src/adapters/claude-code.ts`.
- `projectClaudeCodeHome` currently documents hook scripts and settings patching as out of scope/follow-up.
- The lifecycle implementation already exists in `src/lifecycle.ts`.
- Writeback/event append behavior already exists via `appendSomaMemoryEvent` and the writeback gate.
- The install path should patch Claude Code hook configuration in a Soma-owned/idempotent way and remove only Soma-owned hook entries on uninstall.

## Acceptance Criteria

[...]

PR #324, opened 2026-06-15
fix(bun-probe): validate the frozen bun path and resolve it win32-correctly (#316 sibling)

### LINKED ISSUE #323: bun path resolver freezes an unspawnable MSYS-dialect path on Windows (win32 sibling of #316)
## Problem

`resolveBunExecutable()` (`src/bun-probe.ts`) freezes an absolute bun path into the
substrate hook commands and runtime configs at install time. #316 (#320) hardened that
resolver against ephemeral `/tmp/bun-node-*` paths, but its PATH probe is still
`which bun` on **every** platform.

On native Windows `which` does not exist; under Git Bash / MSYS it resolves in a path
dialect — `/c/Users/.../bun` (forward slashes, drive as a root dir, no `.exe`) — that no
native Windows process can spawn. The resolver freezes that value without validating that
it actually spawns.

This is the **win32 sibling of #316**, and the two are independent: an ephemeral
`/tmp/bun-node-*` path *passes* a `--version` spawn probe, and an MSYS-dialect path is
*not* under the temp dir. Neither check catches the other's case.

## Reproduction (Windows 11, Git Bash, Bun 1.3.5)

On `main` @ `4ea8af8`:

```
$ bun -e 'import {resolveBunExecutable} from "./src/bun-probe.ts"; console.log(JSON.stringify(resolveBunExecutable()))'
"/c/Users/KyleLittle/scoop/persist/bun/bin/bun"
```

That `/c/...` value is the MSYS dialect; a native Windows process cannot resolve it
(`existsSync()` on it returns false, while the native form `C:/Users/.../bun.exe` exists).
It then gets frozen into the hook command / runtime config.

## Impact

- **claude-code**: the `settings.json` hook command carries the unspawnable `/c/...` path, which
  a native Windows process cannot launch. On a hook whose absence does not block the tool call
  (fail-open), that means the policy gate does not fire.
- More generally, on any hook platform that fails open, a frozen unspawnable command cannot
  launch the hook, so the policy gate does not fire.

## Proposed fix

Make the resolver win32-correct and validating, on top of #316 (keep `isEphemeralBunPath`):

- win32 resolution order `SOMA_BUN_PATH` → `process.execPath` under bun → `where bun`;
  never invoke `which` on win32. POSIX order unchanged.
- Normalize the MSYS dialect (`/c/Users/.../bun` → `C:\Users\...\bun.exe`) before validation.
- Validate every winning candidate (incl. the env override) with `existsSync` + a
  `<path> --version` spawn probe; abort loudly rather than freeze a present-but-broken value.

PR to follow.

klittle32:
Fixes #323.

### Problem

`resolveBunExecutable()` (`src/bun-probe.ts`) freezes an absolute bun path into substrate
hook commands and runtime configs. #316/#320 hardened it against ephemeral `/tmp/bun-node-*`
paths, but the PATH probe is still `which bun` on every platform. On native Windows `which`
does not exist, and under Git Bash/MSYS it answers in a dialect — `/c/Users/.../bun`
(forward slashes, no `.exe`) — that no native Windows spawn can resolve. The resolver freezes
that value unvalidated.

Reproduction on `main` @ `4ea8af8` (Windows 11, Git Bash, Bun 1.3.5):

```
$ bun -e 'import {resolveBunExecutable} from "./src/bun-probe.ts"; console.log(JSON.stringify(resolveBunExecutable()))'
"/c/Users/KyleLittle/scoop/persist/bun/bin/bun"
```

A native Windows process cannot launch the `/c/...` dialect form, so the value frozen into the
hook command is unspawnable. On any hook whose absence does not block the tool call (fail-open),
an unlaunchable hook means the policy gate does not fire. This is the win32/MSYS **sibling** of
#316 — an ephemeral path passes a spawn probe and an MSYS path is not under `/tmp`, so the
two checks are independent.

### Fix (`src/bun-probe.ts`)

- **win32 resolution order**: `SOMA_BUN_PATH` → `process.execPath` under bun (a native path
  by construction) → `where bun`. `which` is never invoked on win32. POSIX order unchanged
  (`SOMA_BUN_PATH` → `which bun` → execPath).
- **MSYS normalization**: `/c/Users/.../bun` → `C:\Users\...\bun.exe` before validation.
- **Validation**: every winning candidate — including `SOMA_BUN_PATH` — must pass `existsSync`
  + a `<path> --version` spawn probe, or the install aborts loudly naming the candidate and
  its source. No present-but-broken value falls through silently.
- **#316 preserved**: `isEphemeralBunPath` still screens every candidate; an auto-detected
  ephemeral path is skipped so a durable source can win; an explicit ephemeral
  `SOMA_BUN_PATH` now fails loudly (new behavior — the user named a doomed path).
- Resolution is injectable (`probe`/`platform`/`env` seam) so tests drive every branch
  hermetically. `resolveBunExecutable` is memoized per `SOMA_BUN_PATH` value.

The internal-only `chooseBunExecutable`/`locateBunOnPath` (no consumers outside this file +
its test) are replaced by `resolveValidatedBunPath`; the two production exports
(`resolveBunExecutable`, `requireBunInPath`) keep their signatures.

### Tests

- `test/bun-probe.test.ts`: #316's `isEphemeralBunPath` block is unchanged; its
  ephemeral-resolution cases are re-expressed against the new core (so the #316 guarantees
  stay covered); added a win32/MSYS/POSIX hermetic suite (injected probe fixture) covering
  the resolution order, MSYS normalization of both `SOMA_BUN_PATH` and `where` output,
  validated/broken overrides, spawn-probe failure, and the "`which` never invoked on win32"
  assertion.
- `test/claude-code-install.test.ts` (test-only): the existing test *"issue #236: uninstall
  removes settings entries using the installed Bun path"* simulated a prior install by
  `replaceAll`-ing the frozen path in **raw** `settings.json` text. With a native Windows
  path the on-disk backslashes are JSON-escaped, so the decoded-string replace silently
  no-ops and the simulated state never lands. Rewritten to edit the path on parsed JSON
  values (separator-agnostic). Production removal already matches on the parsed command and
  is unaffected.

### Verification

Windows 11, Bun 1.3.5, Git Bash. No `SOMA_BUN_PATH` pin (this branch resolves bun natively,
so the pin that bare `main` needs on this box is no longer required).

Deterministic:
- `bun run typecheck` — clean (exit 0).
- `bun test test/bun-probe.test.ts` — **26 pass / 0 fail**.
- `resolveBunExecutable()` on this branch returns `C:\Users\KyleLittle\scoop\persist\bun\bin\bun.exe`
  (native; `existsSync` true); on bare `main` it returns `/c/Users/KyleLittle/scoop/persist/bun/bin/bun`
  (`existsSync` false from a native process).
- The test fixed above: `issue #236: uninstall removes settings entries using the installed
  Bun path` — fails 3/3 on this branch before the test fix, passes 3/3 on bare `main`, passes
  3/3 after the fix.

A/B on the touched consumer suite (`test/claude-code-install.test.ts`), same box — **illustrative,
not load-bearing**, because the suite is non-deterministic. One representative run was branch
**19 pass / 3 fail** vs bare `main` **17 pass / 5 fail**, with all 3 branch failures present in
`main`'s set. Those residual failures are a Windows environmental baseline shared with `main`
(a Windows-incompatible `chmod`-based test, HOME-sensitive install tests, and Bun spawn jitter —
exact counts vary run to run), not introduced here. The load-bearing result is deterministic: the
`#236` uninstall test goes red → green (above), and the resolver change introduced no new failure
in the suites I ran (`bun-probe`, `codex-adapter`, `claude-code-install`) — the only test it broke,
`#236`, is now fixed.

**Platform scope: verified on Windows 11 only.** The change is shared-core with explicit
win32/POSIX branches; the POSIX path preserves the prior resolution order and is exercised by
the hermetic suite, but I have not run the suite on macOS or Linux. CI not yet run.

### DIFF
--- a/src/bun-probe.ts
+++ b/src/bun-probe.ts
@@ -1,63 +1,96 @@
 /**
- * Bun discovery + pre-flight (soma#73 sage r1).
+ * Bun discovery + pre-flight.
  *
  * Two callers in the install/adapter surfaces need the same logic:
  *   - `requireBunInPath` — installer pre-flight; throws + remediation
- *     when `env bun` cannot resolve from the substrate's later
- *     spawn environment
- *   - `resolveBunExecutable` — codex adapter; returns an explicit
- *     bun binary path to embed in the hook's runtime config
+ *     when no spawnable, durable bun can be resolved
+ *   - `resolveBunExecutable` — substrate adapters (codex, claude-code);
+ *     returns the validated bun binary path frozen into hook commands
+ *     and runtime configs
  *
- * Sage r1: the previous `process.versions.bun` shortcut bypassed the
- * `which bun` check when the installer happened to be running under
- * Bun. That proves nothing about whether `env bun` resolves at hook
- * fire time. Both probes now always invoke `which bun`.
+ * Two INDEPENDENT failure classes are guarded here, and both must be —
+ * neither check subsumes the other:
+ *
+ *   - soma#316 (ephemeral): under `bun run`, `process.execPath` (and
+ *     sometimes `which bun`) is `/tmp/bun-node-<hash>/bun`, a temp
+ *     self-extraction of Bun's runtime. It spawns fine NOW but is
+ *     deleted on cleanup, so a hook command frozen with it silently
+ *     breaks after a reboot. Passing a `--version` probe is therefore
+ *     not enough — a candidate must also be DURABLE. `isEphemeralBunPath`
+ *     screens it (kept from #320).
+ *   - win32/MSYS (unspawnable): `which` does not exist on native
+ *     Windows and, under Git Bash/MSYS, answers in a path dialect
+ *     (`/c/Users/...`, forward slashes, no `.exe`) that no native
+ *     Windows spawn can resolve. On a fail-open hook platform a frozen
+ *     unspawnable path is not an error: the hook never launches and the
+ *     policy gate silently disables. So on win32 the resolver probes
+ *     `where` (never `which`), normalizes the MSYS dialect, and every
+ *     candidate is checked on disk + spawn-probed (`<path> --version`)
+ *     before it is frozen.
+ *
+ * An ephemeral path passes the spawn probe; an MSYS-dialect path is not
+ * under the temp dir. Both screens run, on every candidate.
  */
 import { spawnSync } from "node:child_process";
+import { existsSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { sep } from "node:path";
 
-/**
- * Locate the bun binary on PATH. Returns the resolved path, or null
- * when bun is unavailable.
- *
- * Always runs `which bun` — running under bun is NOT evidence that
- * a shebang-resolved `env bun` will succeed in the hook's later
- * spawn environment (different PATH, different shell, restricted
- * env via the substrate, etc.).
- */
-export function locateBunOnPath(): string | null {
-  const which = spawnSync("which", ["bun"], { encoding: "utf8" });
-  if (which.status !== 0) return null;
-  const resolved = which.stdout.trim().split("\n")[0];
-  return resolved.length > 0 ? resolved : null;
-}
+/** Where a candidate bun path came from — named in validation errors. */
+export type BunPathSource = "SOMA_BUN_PATH" | "process.execPath" | "where bun" | "which bun";
 
 /**
- * Installer pre-flight. Throws with remediation when bun is not on
- * PATH; returns silently when it is.
- *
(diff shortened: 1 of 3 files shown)


jcfischer (approved):
Reviewed: validating win32 resolver. Two independent screens (ephemeral durability #316 + spawnability) both run per candidate; injectable probe seam gives hermetic branch coverage; per-source error attribution. Behavior change (SOMA_BUN_PATH now validated at preflight, not trusted verbatim) is the correct hardening. CI green. LGTM.

PR #326, opened 2026-06-15
feat: Grok (xAI grok-cli) substrate adapter

### LINKED ISSUE #325: Add a Grok (xAI grok-cli) substrate adapter
## Proposal

Soma installs into Codex, Pi.dev, Claude Code, and Cursor today. This proposes a
fifth substrate: **Grok**, xAI's coding-agent CLI (`grok`, config home `~/.grok/`).
The adapter gives Grok the same `soma install / doctor / reproject / upgrade /
uninstall` lifecycle the other substrates have — a first-class home projection,
not a doc dump.

The implementation is ready; the PR follows immediately and closes this issue.
Filing this as the tracking issue + design overview so scope and review-shape
discussion has a home.

## Why it needs care (and why a probe-first design)

Grok's hook platform, context auto-load surfaces, and tool vocabulary are
**under-documented and fail-open** — a tool runs even if its hook errors. Building
on assumptions would leave the security posture (hook deny) and the
context-delivery layer (skills auto-load) resting on guesses. Grok also shipped
`0.2.38 → 0.2.39` mid-implementation.

So the adapter is built on, and only on, **live-probed runtime facts** (`grok
inspect --json` plus direct hook-behavior probes), captured as a design decision
rather than left as tribal knowledge. The load-bearing probed facts:

- **Context auto-load.** `~/.grok/AGENTS.md` and `~/.grok/skills/<name>/SKILL.md`
  auto-load; home `~/.grok/rules/` and `~/AGENTS.md` do **not**. Context routes
  through the skills surface + an `AGENTS.md` pointer block.
- **Fail-closed deny on a fail-open platform.** A `PreToolUse` hook that writes a
  deny to stdout and exits 0 blocks the call; because the platform is otherwise
  fail-open, the chain denies on **every** error rather than trusting one gate.
- **Passive-hook stdout is ignored** — only `PreToolUse` reads stdout; lifecycle
  injection can't rely on `SessionStart`/`SessionEnd` stdout.
- **Windows hook execution.** Hooks are bare-exec'd against a resolved absolute
  Bun path with explicit `utf8` (NTFS ignores `chmod`, Windows ignores shebangs).
- **Version floor `0.2.38`**, read from `~/.grok/version.json` with no live `grok`
  exec; the empirical tool-name matchers were re-probed byte-identical on `0.2.39`.

These are written up for users in `docs/substrate-adapters.md` (Grok section) and
captured as a design decision (DD-14) in `design/design-decisions.md` — both ship
with the PR.

## Dependency on the bun-path work

The fail-open property is exactly why the Windows bun-path resolver matters here: a
frozen *unspawnable* interpreter path means the hook never launches and the policy
gate silently does not fire. This adapter freezes a **validated, native** bun path
into the hook commands, so it sits on top of the bun-path hardening (#316/#320 and
the win32 sibling in #324). Sequencing the adapter after that fix is intentional.

## What the PR delivers

- `soma install grok --apply` (home projection: auto-loaded `skills/soma` entry +
  Algorithm skill + `AGENTS.md` pointer); `--workspace --apply` overlays the
  project's `.grok/rules/`.
- A fail-closed `PreToolUse` policy hook + per-session lifecycle hooks.
- `soma doctor --substrate grok`, version-floor validation, and a marker-guarded
  `soma uninstall grok` that removes only Soma-owned artifacts.
- Tests that spawn hooks via `node` with a pinned `HOME` — **no test launches a
  live `grok`**, so CI stays deterministic and platform-independent.

## Questions for you

- **Scope/staging:** take it as one cohesive PR, or stage it (e.g. adapter +
  install first, doctor/hooks follow)? The internal seams allow a split.
- **Review shape:** the branch history is clean (failing-test-first, one concern
  per commit) — commit-by-commit, or would you rather squash?
- Anything about substrate-adapter conventions you'd want matched.

PR to follow.

klittle32:
Closes #325.

**Draft, stacked on #324 (win32 bun-path).** This adapter depends on #324 for the Windows fail-open guarantee; I'll rebase onto `main` and flip to ready once #324 lands. On Linux CI the branch is green as-is. The landed ISA work (#311/#313) is already underneath.

## What this adds

Grok (xAI's `grok` CLI, config home `~/.grok/`) as the **fifth Soma substrate** —
`INSTALL_SUBSTRATES = ["codex", "pi-dev", "claude-code", "cursor", "grok"]`
(`src/cli/substrate-lifecycle.ts`). Grok gets the same lifecycle the others have:
`install` (dry-run/apply), `doctor`, `reproject`, `upgrade`, `uninstall`.

The adapter core is **`src/adapters/grok/` — 16 files, ~3,367 lines** (adapter,
install, doctor, version validator, install-manifest, hook entry +
shell-policy-core + policy targets, lifecycle hook, hook-smoke). The rest of the
diff is substrate wiring (registry/types/CLI/install/home-projection), a shared
`version-floor` helper, the user doc + design decision, and the test suites.

## Design — probed, not documented (DD-14)

Grok's hook platform and context surfaces are under-documented and **fail-open**.
The adapter is built on live-probed runtime facts (`grok inspect --json` + direct
hook-behavior probes), captured as **DD-14** in `design/design-decisions.md` and
described for users in `docs/substrate-adapters.md` (Grok section). The
load-bearing facts:

- **Context auto-load.** `~/.grok/AGENTS.md` + `~/.grok/skills/<name>/SKILL.md`
  auto-load; home `~/.grok/rules/` and `~/AGENTS.md` do not. Context is projected
  through `skills/soma/SKILL.md` (entry) + the Algorithm skill + an `AGENTS.md`
  pointer block; project `.grok/rules/` is a workspace-only overlay.
- **Fail-closed deny on a fail-open platform.** A `PreToolUse` stdout-deny + exit 0
  blocks the call; because everything else fails open, the policy/lifecycle chain
  denies on **every** error instead of trusting one declarative gate.
- **Passive-hook stdout ignored.** Only `PreToolUse` reads stdout; lifecycle
  refresh lives in `SessionStart`/`SessionEnd` side effects.
- **Windows hook execution.** Hooks bare-exec a **resolved absolute Bun path** with
  explicit `utf8` (NTFS ignores `chmod`, Windows ignores shebangs). This is why the
  adapter depends on the bun-path resolver — a frozen unspawnable path would
  silently disable the gate on a fail-open platform.
- **Version floor `0.2.38`** read from `~/.grok/version.json`, no live `grok` exec;
  the tool-name matchers (`Shell/Read/Write/StrReplace/Grep/Glob`) were re-probed
  byte-identical on `0.2.39`.

## Tests

Five Grok suites — `test/grok-{adapter,install,doctor,hook,uninstall}.test.ts` —
plus grok cases added to `test/substrate-adapters.test.ts`. **No test launches a
live `grok`:** hook tests spawn via `node` with `HOME`/`USERPROFILE` pinned, so the
suite is deterministic and platform-independent (the cross-platform CI gate).

On-box run (Windows 11, Bun 1.3.5) on this branch, cut from `upstream/main`:

```
bun test test/grok-{adapter,doctor,hook,install,uninstall}.test.ts --timeout 20000
-> 112 pass / 2 fail  (114 tests)
```

The **2 failures are environmental, not introduced by this code**: both are the
"…repo is unusable" simulation (`grok pre-tool-use fails closed…`, `grok
session-start falls back…`), and both fail identically on the pre-merge baseline on
this Windows box (A/B failure-set diff) — sensitive to the Windows test
environment, not the adapter logic.

**Note on the win32 bun-path dependency.** This branch excludes the `bun-probe`
change (that's #324's slice). Without the win32 resolver, the post-install smoke
can't spawn the frozen MSYS-dialect bun path on **Windows**, so those install
tests fail there — the exact fail-open hazard #324 fixes, and the reason this
adapter stacks on it. On **Linux** that path resolves natively (`which bun`), so CI
is unaffected. The 112/2 above is with the win32 resolver present (locally
reproduced by pinning `SOMA_BUN_PATH` to the native `bun.exe`; once #324 lands and
this rebases onto it, no pin is needed).

## Verification

- **Integration:** branch cut from current `upstream/main` (@ `4ea8af8`), `tsc
  --noEmit` clean; the grok suites are 112/2 with the win32 resolver present (above),
  and an A/B failure-set diff against the pre-merge baseline shows **0 regressions**
  in the shared/upstream suites (shared failures are the known Windows-environmental
  baseline).
- **Windows end-to-end (manual):** `soma install grok` dry-run -> apply with
  post-install smoke passing; `soma doctor --substrate grok` -> ok; frozen hook is a
  native validated-Bun command; `soma uninstall grok` round-trips clean, touching
  only marker-guarded Soma artifacts. Also driven via `arc` distribution.
- **Cross-platform:** the suite above is platform-independent by construction — it
  spawns hooks via `node` with a pinned `HOME` and never execs a live `grok`
  (DD-14) — so it runs under Linux CI when this PR's checks fire. It has **not yet
  been run on Linux or macOS** for this adapter; the manual end-to-end evidence
  above is Windows-only.

## Excluded from the PR

Local authoring/tooling artifacts are dropped — they're not part of the adapter:
`.fallowrc.json` + the `.fallow/` `.gitignore` line, `CONCEPTS.md` and `docs/solutions/`
(compound-engineering pipeline outputs). The `bun-probe` change is **#324's** slice,
not duplicated here. Internal authoring tracker labels in comments/test names have been
stripped for upstream.

## Review

Submitted as a single squashed commit. The original development history is
failing-test-first, one concern per commit — happy to re-open it commit-by-commit
if you'd rather review it that way. The `shell-policy-core` extraction in
`src/adapters/grok/hooks/` is grok-internal here; promoting it to a shared module +
Codex adoption is a separate, later change (not in this PR).


jcfischer:
## Review — Grok substrate adapter

Reviewed via two parallel deep-reads: the fail-closed PreToolUse hook chain, and the adapter wiring/lifecycle against the codex/pi-dev baselines. **#324 (the win32 bun-path dependency) is now merged to `main`** — rebase onto it and the `SOMA_BUN_PATH` pin is no longer needed.

**Verdict: architecture is right. The fail-closed property is real and verified. Two MAJORs (one logic, one its test) before flipping to ready; everything else is polish.**

### Verified
- **Fail-closed holds.** Every error/exception path in the PreToolUse chain terminates in `denyPreToolUse` (stdout-deny + exit 2) — config-load failure, malformed input, non-zero CLI status, bad JSON, unexpected structure, uncaught throw all fail closed. Read-only allowed, writes deny. Checked against the actual decision vocabularies (`policy.ts` emits only deny/allow; inbound only ALLOWED/BLOCKED/HUMAN_REVIEW).
- **Wiring solid.** Registered in `INSTALL_SUBSTRATES`, all 5 lifecycle verbs accept grok, doctor registered, uninstall marker-guarded + hash-verified, round-trips clean (no risk to user files in `~/.grok/`; dirs pruned non-recursively → fails closed on ENOTEMPTY). Idempotent — no `Date`/`Math.random`/timestamp nondeterminism. Version floor robust. No TODO/FIXME/placeholder leftovers.

### Must-fix before ready
- **[MAJOR] `shell-policy-core.mjs:804-827` + `:539`** — the second-commit `@<priv>`-glue fix is only complete for **absolute** markers. An unknown verb gluing a *relative* private token behind a non-path prefix (`Frobnicate-Item @.soma/memory/x`) bypasses: `matchesRelativePathPrefix` requires a `.soma/` start (the `@` defeats it), `hasSomaPolicyMarker` is absolute-only, and the backstop text-match is `isAbsolute`-gated → **ALLOW**. Same fail-open class the commit claims to have closed, left open for relative prefixes. Fix: in the backstop, also scan segment text for the descriptor's relative `privatePathPrefixes` (substring, win32-case-folded) and emit the matched root as `sourcePath`. (Bite is narrowed to cwd==home, but the marker backstop exists precisely to not depend on path resolution.)
- **[MAJOR] coverage** — no test exercises that relative `@.soma/...` form. The existing fabricated-marker test uses an *absolute* forward-slash path the backstop already catches. Add the adversarial relative case.

### Should-fix
- **[MINOR] `grok-hook-entry.mjs:363-369`** — denies only on BLOCKED/HUMAN_REVIEW; an unrecognized non-empty `decision` falls through to ALLOW. Safe today only because the server normalizes unknowns to HUMAN_REVIEW — the hook itself isn't fail-closed on an unexpected value. Invert to an allowlist: `if (scan.decision !== "ALLOWED") deny`.
- **[MINOR] `soma-lifecycle.mjs:31-40`** — config fails closed on corrupt JSON, but an empty-but-valid config (`policyMarkers: []`) parses fine → zero targets → ALLOW for shell egress. Assert the required arrays are non-empty on the enforcing verb, else deny (+ test).
- **[MINOR] `StrReplace`** edit-target has no PreToolUse deny fixture carrying a marker in `new_string`; only `Write` is covered, despite both being in the matcher with different input keys.
- **[MINOR] `version-floor.ts:35`** — `compareVersions` coerces an unparseable version to `-1` (== "below floor"). Dead-defensive today (pre-rejected at `version.ts:2066`), but a footgun for future callers. Throw on unparseable instead of coercing.

### Nits
- `grok-hook-entry.mjs:158-161` — comment claims the stdout deny is honored "regardless of exit code (verified for 0 and 1)", but only exit 2 is asserted anywhere. Add a test that a stdout deny with exit 0/1 is still honored, or drop the claim.
- `install.ts:1412` — the `rules` → `rules-soma` marker guard is untested at home scope (a foreign `~/.grok/rules/<dir>` survives — harmless, but subtle). Add a comment or a home-scope test.
- PostToolUse matcher diverges from codex (no `apply_patch`) — legitimate (Grok has no patch-style edit tool); worth a one-line note so the divergence reads as intentional.

Recommend: rebase onto `main` (now has #324) → fix the relative-prefix backstop + add the adversarial test → flip to ready.

🤖 Review assisted by Claude Code (two parallel deep-read agents).

klittle32:
Thanks for the deep read — every finding was fair. Rebased onto `main` (so #324 is underneath and the `SOMA_BUN_PATH` pin is gone), addressed all of it, and squashed to a single commit. Mapping below.

### Must-fix
- **[MAJOR] relative `@.soma/...` fail-open** — fixed. The fail-closed backstop (`shell-policy-core.mjs`, `matchedPrivateMarkerSource`) now, after the absolute-marker text scan misses, scans the segment text for any descriptor relative `privatePathPrefix` at a path boundary (case-folded on win32, mirroring `matchesRelativePathPrefix`; the boundary keeps a glued `@`/quote/`(` matching while a benign `my.somatic` does not) and emits that prefix's absolute root anchored at the soma-home parent. The `.soma` root is `somaHome`, a private scope root `policy check` always honors regardless of `policyMarkers`, so the deny is enforceable. `Frobnicate-Item @.soma/memory/x` now denies.
- **[MAJOR] coverage** — added two tests: a deterministic extractor unit test (`relative private prefix glued behind a non-path prefix still fails closed (backstop)`) asserting the emitted source is the private root and that a benign `.somatic` token does **not** over-block, plus an end-to-end installed-hook fixture (`...fails closed on a RELATIVE private marker glued behind a non-path prefix`).

### Should-fix
- **inbound-scan allowlist** — `inboundScanDenyReason` now denies unless `decision === "ALLOWED"`, so an unrecognized non-empty decision no longer falls through to allow.
- **empty-but-valid config** — the `soma-lifecycle.mjs` bootstrap now fails closed on the enforcing verb when the config parses but carries no enforcement inputs (empty `policyMarkers`/`privateRoots` or no `somaHome`), with a deny test.
- **StrReplace fixture** — added a PreToolUse deny fixture carrying a marker in `new_string` (Write's sibling input key).
- **version-floor** — `compareVersions` now throws on an unparseable version instead of coercing to `-1` (both callers already pre-reject unparseable input, so this only hardens future callers).

### Nits
- Reworded the `denyPreToolUse` comment: dropped "verified for 0 and 1" (no test launches a live grok) and attributed the any-exit-code behavior to grok's hook docs.
- Added a comment at the `rules`→`rules-soma` guard noting the home-scope intent (a foreign `~/.grok/rules/<dir>` is intentionally left untouched; the marker, not the path, decides removal).
- Noted that omitting an `apply_patch` analogue from the PostToolUse matcher is intentional (Grok exposes no patch-style edit tool).

### Verification
- `tsc --noEmit` clean.
- Full grok suite **116 pass / 2 fail** on Windows 11 (Bun 1.3.14), unpinned. The 2 failures are the pre-existing Windows-only environmental "repo unusable" sims (they no-op `chmod` on Windows and fail on bare `main` too); the four new tests above are green. CI on this push will run the Linux suite.

🤖 Fixes assisted by Claude Code.

klittle32:
Small correction to my verification note above, in the interest of precision: the 2 failing tests are grok-specific (`grok session-start … repo is unusable`, `grok pre-tool-use fails closed … repo is unusable`), so they don't exist on bare `main` — "fail on bare main too" was imprecise. The accurate claim: they are Windows-only environmental "repo unusable" sims (they no-op `chmod` on Windows) and fail **identically before this PR's review changes**, so they're not regressions; the grok suite runs 114/0 on Linux. And to be exact about CI: the reporting check here is "Deterministic portability" (green) — I shouldn't have implied a full Linux test run.

jcfischer (approved):
Re-reviewed after the update. Verified in code (not the summary): all 9 prior findings genuinely fixed — inbound allowlist, empty-config fail-closed, StrReplace fixture, compareVersions throws, the 3 nits, and the reported `@.soma/` backstop bypass + 2 new tests.

Fail-closed property re-confirmed across all error paths.

One residual, not blocking: the relative-marker backstop closes `@.soma/` but `@./.soma/` and `@~/.soma/` (a path-continue char between the non-path prefix and the marker) still fall through to ALLOW — same class, one glue-char wider. It's a 4th-line defense-in-depth backstop with thin exploitability; tracking as a follow-up issue rather than blocking. LGTM.

PR #328, opened 2026-06-22
fix(grok): close the glued-prefix backstop fail-open for the whole home-anchor class (#327)

### LINKED ISSUE #327: Grok hook backstop: relative-marker fail-open not fully closed (@./.soma/, @~/.soma/)
## Summary

The Grok PreToolUse fail-closed backstop (`src/adapters/grok/hooks/shell-policy-core.mjs`, `matchedPrivateMarkerSource`) closes the relative-marker-glued-behind-a-non-path-prefix fail-open for the form reported and fixed in #326 (`Frobnicate-Item @.soma/memory/x` → DENY), but **the class is not fully closed**. A path-continue char between the non-path prefix and the marker still falls through to ALLOW:

- `Frobnicate-Item @./.soma/memory/x` → **ALLOW** (should DENY)
- `Frobnicate-Item @~/.soma/memory/x` → **ALLOW** (should DENY)

## Root cause

The backstop's relative-prefix scan uses:

```js
const boundary = "[^A-Za-z0-9._/~-]";
const prefixed = new RegExp(`(?:^|${boundary})${escaped}/`); // escaped = "\.soma"
```

`/`, `.`, `~`, `-` are treated as path-continue chars (intentionally, so a nested `someproject/.soma` is not over-blocked). But that also means when the marker is glued as `@./.soma/` or `@~/.soma/`, the char immediately before `.soma` is `/` (a continue char, not a boundary), so the regex misses. The other two branches also miss: these `@`-glued tokens resolve via `resolveToolPath` to a literal `@.`/`@~` segment not under any private root (so `resolvableToken` misses), and they carry no absolute marker (so the absolute scan misses).

## Reproduction (boundary regex in isolation)

```js
const boundary = "[^A-Za-z0-9._/~-]";
const prefixed = new RegExp(`(?:^|${boundary})\\.soma/`);
prefixed.test("Frobnicate-Item @.soma/memory/x");   // true  (DENY — fixed in #326)
prefixed.test("Frobnicate-Item @./.soma/memory/x"); // false (ALLOW — residual)
prefixed.test("Frobnicate-Item @~/.soma/memory/x"); // false (ALLOW — residual)
prefixed.test("Copy-Item my.somatic/x public");     // false (correct — no over-block)
```

## Severity

Low. This is the 4th-line defense-in-depth backstop — structured passes catch real verbs first, and the precise per-form handling plus `resolveShellPath` already cover `~/.soma`, `$HOME/.soma`, `%USERPROFILE%/.soma`, and bare `.soma/` that resolve under a root. The residual is limited to unknown verbs gluing the marker behind a non-path prefix with an interposed continue char, a form that does not correspond to how a real tool reads the file. Surfaced during the #326 review; merged with this as an agreed follow-up rather than a blocker.

## Suggested fix

Make the relative-prefix scan recognize `.soma` as a **path-component start** even when a `./` or `~/` is glued behind a non-path prefix — e.g. after the existing match misses, also test for `.soma/` (or the bare form) immediately preceded by `./` or `~/` that is itself glued behind a boundary (`(?:^|boundary)(?:\./|~/)\.soma(?:/|$)`), emitting the same soma-home-anchored root. Keep the benign `my.somatic` and nested `someproject/.soma` cases allowed (regression-guard both).

Add adversarial tests for `@./.soma/…` and `@~/.soma/…` alongside the existing `@.soma/…` cases in `test/grok-hook.test.ts`.

jcfischer:
Fixes #327.

### Problem

The #326 fail-closed backstop (`src/adapters/grok/hooks/shell-policy-core.mjs`, `matchedPrivateMarkerSource`) denied the reported `Frobnicate-Item @.soma/memory/x`, but the **class** was not fully closed. An interposed `./` or `~/` between the non-path prefix and the marker still fell through to ALLOW:

- `Frobnicate-Item @./.soma/memory/x` → **ALLOW** (should DENY)
- `Frobnicate-Item @~/.soma/memory/x` → **ALLOW** (should DENY)

The `/` from `./`|`~/` is a path-continue char, so it sits immediately before `.soma` and the plain `(?:^|boundary)\.soma/` scan misses. These `@`-glued tokens also resolve via `resolveToolPath` to a literal `@.`/`@~` segment under no root (branch-1 miss) and carry no absolute marker (absolute-scan miss), so they reached `policy check` as a toothless ALLOW.

### Fix

Add an optional, **boundary-gated** relative anchor `(?:\./|~/)?` between the boundary and the marker. It stays boundary-gated, so the over-block guards hold:
- benign `my.somatic` → ALLOW (no `.soma/`, no boundary-anchored bare `.soma`)
- nested `proj/.soma/x` → ALLOW (project-local soma, not the home root — `/` before `.soma` with no boundary-gated `./`|`~/` anchor)

### Tests

- Unit (`createShellPolicyExtractor`): both `@./.soma/` and `@~/.soma/` emit the private **root** as `sourcePath` (rides the relative leg — empty `policyMarkers`), plus a nested-`proj/.soma` over-block guard.
- Installed-hook e2e: both glued forms deny end-to-end (`decision: deny`, exit 2).

### Verification

- `bun test test/grok-hook.test.ts` — 49 pass / 0 fail (47 + 2 new).
- `bun test` grok-adapter/doctor/install/uninstall + substrate-adapters — 77 pass / 0 fail.
- Regex matrix checked directly: the two `@./`/`@~/` variants now DENY; `my.somatic` and `proj/.soma` (and deeper `a/b/.soma`) stay ALLOW.

macOS run only (the suite is platform-independent by construction — hooks spawn via node with pinned HOME, no live grok); CI runs the Linux gate.

### DIFF
--- a/src/adapters/grok/hooks/shell-policy-core.mjs
+++ b/src/adapters/grok/hooks/shell-policy-core.mjs
@@ -867,16 +867,38 @@ export function createShellPolicyExtractor(descriptor) {
     );
     if (marker) return resolve(marker);
 
-    // 3. A candidate matches a descriptor RELATIVE private prefix (`.soma/…`,
-    //    `./.soma/…`, bare `.soma`, and their deglued `@`-prefixed forms). No
-    //    resolution under a root is needed — the prefix's somaHome-anchored
-    //    root is the enforceable source. A bare-alnum `my.soma` never deglues,
-    //    and matchesRelativePathPrefix is boundary-exact, so no over-block.
+    // 3. A descriptor RELATIVE private prefix in the segment TEXT, at a
+    //    token/path boundary. The full-text scan (not just per-token) is what
+    //    catches a marker glued INSIDE an opaque alnum-led token that does not
+    //    deglue — e.g. `iex"@.soma/x"`, where the `@` is a boundary in the
+    //    joined text but the token starts `iex` so step 1/1b's per-token deglue
+    //    leaves it intact. The optional `(?:\./|~/)?` anchor also covers the
+    //    `@./.soma/`, `@~/.soma/` standalone forms. Boundary-gated and
+    //    case-folded on win32 (mirroring matchesRelativePathPrefix), so a
+    //    benign `my.somatic` and a nested `proj/.soma` (project-local, not the
+    //    home root) do NOT match. Emits the prefix's somaHome-anchored root —
+    //    a private scope root `policy check` honors regardless of policyMarkers.
+    //
+    // Note: an env-var home spelling embedded inside an opaque alnum-led token
+    // (`iex"@$HOME/.soma/x"`) is reachable by neither the per-token deglue nor
+    // this text scan (env-var spellings are built from boundary chars); that
+    // residual is out of scope here and is not a valid file-reading form.
     const home = somaHomeParent(config);
     if (home) {
-      for (const token of candidates) {
-        for (const entry of privatePathPrefixes) {
-          if (matchesRelativePathPrefix(token, entry)) return resolve(home, entry.path);
+      const fold = process.platform === "win32";
+      const folded = fold ? text.toLowerCase() : text;
+      // A boundary char is anything that cannot continue a path token.
+      const boundary = "[^A-Za-z0-9._/~-]";
+      // Optional home-relative anchor glued between the boundary and the
+      // marker: `./` (cwd-relative) or `~/` (home-relative).
+      const relAnchor = "(?:\\./|~/)?";
+      for (const entry of privatePathPrefixes) {
+        const prefix = fold ? entry.path.toLowerCase() : entry.path;
+        const escaped = prefix.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+        const prefixed = new RegExp(`(?:^|${boundary})${relAnchor}${escaped}/`);
+        const bare = entry.bare ? new RegExp(`(?:^|${boundary})${relAnchor}${escaped}(?:$|${boundary})`) : undefined;
+        if (prefixed.test(folded) || (bare && bare.test(folded))) {
+          return resolve(home, entry.path);
         }
       }
     }
--- a/src/adapters/grok/hooks/shell-policy-core.mjs
+++ b/src/adapters/grok/hooks/shell-policy-core.mjs
@@ -249,6 +249,24 @@ function matchesRelativePathPrefix(token, entry) {
   return candidate.startsWith(`${prefix}/`) || candidate.startsWith(`./${prefix}/`);
 }
 
+// soma#327: a private reference glued behind a non-path prefix
+// (`@.soma/…`, `@./.soma/…`, `@~/.soma/…`, `@$HOME/.soma/…`,
+// `@%USERPROFILE%/.soma/…`, `@${env:USERPROFILE}/.soma/…`) defeats both the
+// per-token relative match (the token starts with the glue char, not the
+// anchor) and `resolveShellPath`'s `^`-anchored home normalizers, so it
+// resolved under no root and ALLOWED. Strip a single LEADING run of non-path
+// "glue" chars (`@`, `"`, `'`, `(`, `=`, …) and re-run the SAME resolvers on
+// the remainder — closing every home-anchor spelling through
+// normalizeShellPathToken instead of re-encoding each one here (which is how
+// #326→#327 leaked a spelling at a time). The kept set includes every char
+// that can BEGIN a path/home anchor (`$ % ~ . / \ -` and alnum), so a bare
+// `my.soma` is never turned into `.soma` — no over-block.
+function degluePrivateToken(token) {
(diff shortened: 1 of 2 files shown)


jcfischer:
## Self-review — #328

### The change is correct and a strict improvement
- Closes the reported `@./.soma/…` and `@~/.soma/…` via a boundary-gated optional `(?:\./|~/)?` anchor. Verified end-to-end (decision: deny, exit 2) and at the extractor unit level.
- No over-block: `my.somatic`, nested `proj/.soma/x`, deeper `a/b/.soma/x` all still ALLOW — confirmed by direct matrix run. The fix stays boundary-gated, so it doesn't widen into project-local `.soma` dirs.
- typecheck clean; `grok-hook` 49/0 (+2 new), grok-adapter/doctor/install/uninstall + substrate-adapters 77/0; CI "Deterministic portability" green.

### But the class is still not fully closed (found reviewing this PR)
Probing the real extractor for sibling forms, the **env-var home anchors glued behind a non-path prefix still fail open**:

```
ALLOW  @$HOME/.soma/memory/x
ALLOW  @${HOME}/.soma/memory/x
ALLOW  @%USERPROFILE%/.soma/memory/x
ALLOW  @${env:USERPROFILE}/.soma/memory/x
```

Same root cause as #326→#327→this: a glued prefix defeats `resolveShellPath`/`matchesRelativePathPrefix` (their home-anchor normalizers are `^`-anchored, so `@$HOME/…` isn't recognized), and the backstop's text scan only knows `./`/`~/`, not the env-var spellings.

### Recurrence diagnosis — stop instance-patching the regex
This is the **third** patch to one fail-open (`@.soma/` in #326-review → `@./`,`@~/` here → env-var now). Each fix teaches the regex one more anchor spelling; the env-var forms (`$HOME`, `${HOME}`, `$env:USERPROFILE`, `%USERPROFILE%`) are built from regex-boundary chars (`$ { } : %`), so a boundary-based text scan structurally cannot deglue them without duplicating `normalizeShellPathToken`'s spelling list — the duplication that breeds the next miss. A naive "strip leading junk and re-resolve" over-blocks `my.soma/x` (loses the boundary discriminator). A real structural fix means teaching the shared resolver to tolerate one leading glue run *before a recognized home anchor* — a riskier change to shared resolution semantics that wants its own PR + over-block regression suite.

### Severity of the residual: LOW
The backstop is the 4th-line defense-in-depth for unknown verbs only. The glued env-var forms (`@$HOME/.soma`) are not valid pwsh that would actually read the file (`@` is splat; `$HOME` is a var) — closer to non-exploitable than the `./`/`~/` forms this PR closes.

### Recommendation
Merge #328 as the realistic-forms fix (it closes the two demonstrable variants and regresses nothing), and track the env-var-glued residual as a documented LOW limitation — either reopen #327's scope or a fresh issue, with the structural-deglue option noted. Do **not** add a 3rd anchor-spelling regex.

jcfischer:
## Update — structural fix, whole class closed

Reworked from the per-spelling regex into a **structural deglue** so this stops being whack-a-mole. Per the review on this PR, the regex approach closed `@.soma/`, `@./.soma/`, `@~/.soma/` but left the env-var home-anchor spellings open (`@$HOME/.soma`, `@${HOME}/`, `@%USERPROFILE%/`, `@${env:USERPROFILE}/`) — same class, because those spellings are built from regex-boundary chars and can't be deglued by a text scan without re-encoding `normalizeShellPathToken`'s spelling list.

**New approach:** strip a single leading run of non-path glue chars (`@`, quote, `(`, `=`, …) from each token and re-run the **same resolvers** (`resolveShellPath` / `matchesRelativePathPrefix`) on the remainder. Every home-anchor spelling the resolver already understands collapses to the somaHome private root through one resolver-backed check — no spelling duplication, so there's no "next spelling" to leak.

The deglue keeps every char that can BEGIN a path/home anchor (`$ % ~ . / \ -` and alnum), so a bare-alnum `my.soma` is never turned into `.soma`.

### Verified (real extractor, empty policyMarkers)

```
DENY   @.soma/…   @./.soma/…   @~/.soma/…
DENY   @$HOME/.soma/…   @${HOME}/.soma/…   @%USERPROFILE%/.soma/…   @${env:USERPROFILE}/.soma/…
ALLOW  proj/.soma/x   my.soma/x   a/b/.soma/x   my.somatic-notes.txt
```

### Tests / suites
- `bun run typecheck` clean.
- `test/grok-hook.test.ts` 49/0 — unit + installed-hook e2e now assert deny across the bare/`./`/`~/`/`$HOME`/`${HOME}`/`%USERPROFILE%`/`${env:USERPROFILE}` and paren/`=`-glued forms, plus the over-block guards.
- grok-adapter/doctor/install/uninstall + substrate-adapters + policy: 96/0.

Preserved: the precise-source emit under an absolute marker root (so a transfer still reports the exact file, not the root); the absolute-marker substring scan; read-only allow; `Copy-Item`/structured passes untouched (backstop skips read-only and write-sink verbs).

Issue #331, opened 2026-06-22
Algorithm: OBSERVE current-state-verification floor (attacks 63% observe-stall)


jcfischer:
**63% of runs (118/188) stall at OBSERVE.** The most-repeated reflection across runs: *"a smarter algorithm would have verified all current-state assumptions before the interview / before planning."* Runs either die at OBSERVE or advance on unverified assumptions and rework later.

## Prompt change (P2)
- OBSERVE prompt forces probing current-state assumptions before advancing: read the repo/system, confirm each named field/flag/route/file actually exists, record the probe as evidence.
- Advancing OBSERVE→THINK should require at least one current-state probe on record.

## Acceptance
- [ ] OBSERVE prompt includes an explicit current-state-verification floor.
- [ ] OBSERVE→next gate checks for ≥1 recorded current-state probe.
- [ ] A run that advances with zero current-state probes is flagged/blocked.
- [ ] Tests cover the gate.

Evidence: `Plans/2026-06-22-algorithm-runner-prompt-findings.md` (P2). Related: tautological-verification issue.

jcfischer:
Context: P2 is the first harvested output of PAI's meta-reflection layer (one recurring `reflection_q2`: 'a smarter algorithm would have verified current-state assumptions'). The layer that generates P2-class fixes is now #333 — P2 is upstream-linked to it. Do P2 now (cheap, already harvested); #333 is the engine that finds the next ones.

jcfischer:
Claiming (autonomous SOP run). Stacked on #330/PR #335 since both touch the OBSERVE/LEARN gates in algorithm.ts. Plan: TDD the OBSERVE current-state-verification floor (gate requires >=1 recorded current-state probe before OBSERVE->THINK) + the prompt half.

jcfischer:
Claiming — building the OBSERVE current-state floor: typed `AlgorithmObservation` array + observe→think gate requiring ≥1 probed/tested current-state observation (reuses #330's EvidenceKind). Worktree from main.

jcfischer:
Engine-side gate **done** in PR #336 (Sage-clean: commented, 0 blockers / 0 majors / 1 HonestOracle nit; 1428 tests pass).

Delivered:
- Typed `AlgorithmObservation {claim, evidence, evidenceKind}` + `observe → think` gate requiring ≥1 `probed`/`tested` observation (`hasCurrentStateProbe`, shared by the gate and sync so they can't drift).
[...]

Issue #330, opened 2026-06-22
Algorithm: kill tautological verification — type evidence (specified/probed/tested) + gate


jcfischer:
**Top failure mode found in 188 real runs.** The verify gate accepts "the design says X ∴ passed" as evidence. PROMOTED review names it: *"All 37 ISCs are marked 'passed-design' … tautological verification … should claim 37/37 specified, not verified."* Cross-referenced with the data: only **62/188 runs (33%) are ever verified**, and where verification exists it is often assertion, not a probe.

## Runner change (R1)
- Add an evidence **kind** to criterion verification: `specified | probed | tested`.
- Split criterion status: `passed-design` vs `passed-verified`.
- Completion gate: behavioral criteria require `probed`/`tested`. When no executable evidence exists, force a `deferred-probe` state instead of green.

## Prompt change (P1)
- Verifier prompt: a specification/design statement does **not** verify a *behavioral* criterion. Require an executable probe (test output, curl, grep of running state) or mark `deferred-probe`.

## Acceptance
- [ ] `soma algorithm verify` requires/records an evidence kind.
- [ ] `complete` gate blocks on a behavioral criterion that is only `specified`.
- [ ] Verifier prompt updated; a run that only inspects its own output text cannot reach `passed-verified`.
- [ ] Tests cover specified-only → deferred-probe.

Feeds the `checkpoint` direction (a checkpoint verdict carries an evidence kind). Evidence: `Plans/2026-06-22-algorithm-runner-prompt-findings.md` (R1+P1).

jcfischer:
Claiming for implementation (autonomous SOP run). Plan: fresh worktree from main, TDD on the verification evidence-kind model + completion gate, then prompt (P1). PR → review loop → merge.

PR #338, opened 2026-06-23
feat(algorithm): port the meta-reflection layer — the generator behind P2 (#333)

### LINKED ISSUE #333: Algorithm: port the meta-reflection layer (the generator behind P2/#331)
Promoted from the #332 backlog (R7b) to its own priority issue. **Both this and #331 (OBSERVE current-state floor) came from the same source: PAI's per-run meta-reflection.** P2 was one harvested `reflection_q2` ("a smarter algorithm would have verified current-state assumptions before proceeding"). Porting the layer means P2-class improvements surface themselves instead of waiting for a manual 188-run dig.

## Context (verified)
PAI captured a per-run meta-reflection (`REFLECTIONS/algorithm-reflections.jsonl`, fields `reflection_q1/q2/q3`, `doctrine_fired`, `satisfaction_prediction`, `within_budget`). It ended 2026-05-07 at the PAI→Soma boundary and was **never ported** — Soma's `src` has no reflection writer. Soma kept *criterion-level* learning ("a lesson from this run") but dropped the *meta* layer ("how the algorithm itself should have behaved"). That meta layer is the highest-value improvement signal in the corpus.

## Port it, substrate-neutral (not PAI's jsonl)
On a run reaching a late/terminal phase (or a periodic reflection pass), capture a Soma-native meta-reflection:
- **What a smarter run would have done** — missed-early-step / missed-verification-or-parallelization / highest-value-move (the q1/q2/q3 signal).
- **Which gates fired vs not** — Soma's gate equivalents of `doctrine_fired` (current-state floor, completeness, probe).
- **Effort-vs-criteria sanity** + within-budget.

This is the Algorithm's LEARN phase done right, and the first concrete instance of the dream-cycle direction (`Plans/2026-06-22-soma-direction-verification-primitive.md`) applied reflexively to the algorithm itself: the agent *proposes* a meta-reflection, a checkpoint records it.

## Make it generative (the point)
- Meta-reflections accumulate into a queryable improvement backlog.
- Recurring ones get promoted to runner/prompt changes (P2 is the proof: one recurring reflection → one prompt fix).
- A digest surfaces "the N most-repeated 'a smarter algorithm would have…' signals."

## Acceptance
- [ ] A run reaching verify/learn/complete produces a meta-reflection record (Soma-native, substrate-neutral).
- [ ] Reflections are queryable/diggable (`soma algorithm reflections` or a digest).
- [ ] Recurring-signal surfacing demonstrated on the existing corpus (would have surfaced P2).
- [ ] Tests cover capture + the recurring-signal digest.

Related: #331 (P2, the first harvested output), #330 (verification integrity — meta-reflection is itself evidence the algorithm should change). Sequencing: behind Cortex, but ahead of the #332 backlog.

jcfischer:
Closes #333.

Ports PAI's per-run meta-reflection ("how the Algorithm *itself* should have behaved") — the layer that died at the PAI→Soma boundary and that produced P2/#331 in the first place. Design steered with JC.

## The record (typed, on the run)

`AlgorithmMetaReflection`, two halves by design — the deterministic-spine / model-proposal split the direction doc settled on:
- **`gatesFired`** — for a LIVE reflection, computed by `computeGatesFired` from real run state via the *same predicates the live gates enforce* (`currentStateFloor` #331, `learnGateClean` #330, `completeness`). For an IMPORTED historical record it's a documented, lossy best-effort mapping from PAI's `doctrine_fired` (PAI has no Soma run to probe) — caller-asserted history, not re-derived facts.
- **`smarterRun`** — the MODEL's proposal (q1/q2/q3: missed-early-step / missed-verify-or-parallel / highest-value-move). Free text, caller-asserted.

## Surfaces

- `soma algorithm reflect --id <run> [--missed-early-step …] [--missed-verify-or-parallel …] [--highest-value-move …] [--satisfaction 0-10] [--within-budget|--over-budget]` — gate-flags computed, q-signals supplied. At least one q-signal required.
- `soma algorithm reflections --id <run>` — list a run's reflections (`--id` + `--digest` together is an error).
- `soma algorithm reflections --digest [--pai-source <jsonl>]` — rank the cross-run improvement backlog. **Gate-miss counts RANK, q-signal keyword buckets ENRICH** (keyword buckets only steer enrichment, never the ranking, so they can't corrupt the backlog).
- Session-end mirrors reflections into the run's `LEARNING/ALGORITHM/<run>.md`.

## Generative proof (acceptance #3)

The digest over a committed sample of the historical PAI corpus (`test/fixtures/pai-algorithm-reflections-sample.jsonl`, vendored from the real jsonl at `~/.soma` which isn't version-controlled) ranks **current-state verification first** — 3 of 4 runs missed `live_probe`. That is P2/#331, the top finding of the manual 188-run dig, re-surfaced automatically. The digest *ranks* the backlog; a human or agent still writes the fix.

## Notes

- Back-compat: runs without `metaReflection` default to `[]` on load (same pattern as #331's `observations`). New provenance op `reflection.record`.
- Pure digest + import modules (`algorithm-reflection-digest.ts`, `algorithm-reflection-import.ts`) — no IO, fully unit-tested. `compactSmarterRun` is the shared smarterRun builder (recorder + CLI + importer).

## Acceptance

- [x] A run reaching verify/learn/complete produces a meta-reflection record (Soma-native, substrate-neutral)
- [x] Reflections are queryable/diggable (`soma algorithm reflections [--digest]`)
- [x] Recurring-signal surfacing demonstrated on a sample of the existing corpus (re-surfaces P2)
- [x] Tests cover capture + the recurring-signal digest

## Tests

`bun run typecheck` ✓, `bunx eslint` ✓, **1438 pass / 0 fail**.


jcfischer (commented):

7 finding(s): 2 important, 3 suggestion, 2 nit.


### CodeQuality

- **[suggestion]** `src/cli/algorithm.ts:210` — **parseSatisfaction tolerates trailing garbage**
  `Number.parseInt(value, 10)` parses `--satisfaction 7abc` as `7` silently rather than rejecting it. The integer check passes since parseInt stops at the first non-digit.
  Fix: Use `Number(value)` then `Number.isInteger`, so non-numeric trailing chars are rejected.

- **[nit]** `src/algorithm.ts:332` — **satisfaction range allows non-integers via API**
  `recordAlgorithmMetaReflection` only checks `satisfaction < 0 || satisfaction > 10`, accepting fractional values like 5.5, while the CLI's `parseSatisfaction` requires an integer. Minor contract mismatch between the two entry points.
  Fix: Add `Number.isInteger` to the core validation to match the CLI.

### Architecture

- **[nit]** `src/types.ts:359` — **Collection field named in singular**
  `metaReflection: AlgorithmMetaReflection[];` is an array but named singular, while the sibling collection `observations` on the same interface is plural. Inconsistent public-surface naming on a permanent exported field.
  Fix: Name it `metaReflections` for consistency with `observations`.

### Maintainability

- **[suggestion]** `src/algorithm-reflection-digest.ts:88` — **q-signal triple destructured/filtered in 4 places**
  The `[missedEarlyStep, missedVerifyOrParallel, highestValueMove]` filter-truthy shape recurs: `signalsOf` here, the `some(...)` validation in algorithm.ts `recordAlgorithmMetaReflection`, and the `.filter(...)` listing in cli/algorithm.ts reflections — each re-enumerates the same three fields. Adding a fourth q-signal means editing every site.
  Fix: Export `signalsOf(smarterRun)` from the types/digest module and reuse it in the validation and CLI listing instead of re-destructuring.

### HonestOracle

- **[important]** `test/algorithm-meta-reflection.test.ts:110` — **"Demonstrated on the real PAI corpus" is tested against a synthetic fixture**
  The PR sells acceptance #3 as proof over "the real 4-record PAI corpus," but the only test (`digest over the imported PAI corpus...`) builds an inline `const jsonl = [...]` with 3 of 4 records hardcoded `live_probe: false`. Current-state ranks first by construction, not by surfacing anything from the actual `algorithm-reflections.jsonl`, which no test reads.

- **[important]** `CHANGELOG.md:21` — **Overclaim: ranking a backlog ≠ "generate runner/prompt fixes"**
  CHANGELOG asserts the digest is "the proof that recurring meta-reflections generate runner/prompt fixes." The code sorts categories by `gateMissCount` and prints text (`renderReflectionDigest`); it produces a ranked list, not a fix. The load-bearing word "generate" is not supported by the diff.

- **[suggestion]** `src/algorithm.ts:296` — **Acceptance "a run produces a meta-reflection" is manual and optional**
  Acceptance #1 ("A run reaching verify/learn/complete produces a meta-reflection record") is checked, but `recordAlgorithmMetaReflection` only fires when an operator runs `soma algorithm reflect` with caller-supplied q-signals; nothing auto-produces one. Session-end only mirrors reflections that already exist.


---

jcfischer (commented):

6 finding(s): 1 important, 3 suggestion, 2 nit.


### CodeQuality

- **[nit]** `src/algorithm-reflection-digest.ts:181` — **Digest header labels output by signal frequency, but ranking is gate-miss**
  Header reads `most-repeated 'a smarter run would have…' signals:` yet entries sort by `gateMissCount` first — a top entry can have few signals. The header advertises the enrichment axis, not the actual ranking axis.
  Fix: Reword header to reference gate-miss ranking, e.g. "ranked by recurring unmet gates".

### Architecture

- **[suggestion]** `src/algorithm.ts:22` — **Core recorder depends on the digest/analysis module**
  `import { compactSmarterRun } from "./algorithm-reflection-digest"` makes the core run-mutation path depend on the reporting layer; the digest module's own header calls it "the generative half… turn accumulated reflections into a ranked backlog." Shared shape-builder living in the consumer module inverts the dependency direction.
  Fix: Move `compactSmarterRun` to a neutral module (e.g. alongside the AlgorithmMetaReflection type) so recorder, CLI, importer, and digest all depend down onto it.

### Maintainability

- **[nit]** `src/lifecycle.ts:271` — **Gate-boolean rendering duplicated across CLI and lifecycle**
  lifecycle renders `gates: currentStateFloor=${currentStateFloor}, learnGateClean=...` while cli/algorithm.ts renders the same three flags as `floor=${r.gatesFired.currentStateFloor} learn=... complete=...`. Two hand-maintained enumerations of the same AlgorithmGatesFired keys drift if a fourth gate is added.
  Fix: Extract a `formatGatesFired(gates)` helper next to compactSmarterRun; both sites call it (accept a format flag if the two layouts must differ).

### HonestOracle

- **[important]** `src/algorithm-reflection-digest.ts:162` — **Keyword buckets do affect ranking, contradicting the stated guarantee**
  CHANGELOG claims "gate-miss counts rank, q-signal keyword buckets enrich" (and the PR says buckets "can't corrupt the backlog"), but the sort is `b.gateMissCount - a.gateMissCount || b.signalCount - a.signalCount || a.label...` — signalCount comes from keyword bucketing and orders all gateless categories and breaks gate-miss ties. A reader told bucket edits are ranking-safe would be wrong.

- **[suggestion]** `src/algorithm-reflection-import.ts:56` — **"Re-surfaced automatically" is circular for the imported corpus**
  For imports, `currentStateFloor: bool(doctrine.live_probe)` means the digest's gate-miss ranking just counts the PAI record's own `live_probe:false` flag — "caller-asserted history, not re-derived facts" per the same file. The CHANGELOG's "re-surfacing P2 (#331)... evidence" presents counting a pre-labeled field as independent discovery.

- **[suggestion]** `src/algorithm.ts:290` — **"Can never disagree with the gates" overclaims a point-in-time snapshot**
  computeGatesFired's doc asserts it "can never disagree with what the gates would say," but gatesFired is captured at reflect-time in whatever phase getRunPhase returns; a later criterion/observation change makes the stored snapshot diverge from the gate's verdict at enforcement time. The "never" is stronger than a snapshot supports.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/cli/algorithm.ts:858` — **--pai-source silently ignored when --id is given**
  The list branch (`if (options.id !== undefined)`) returns before `options.paiSource` is read, so `reflections --id X --pai-source Y` discards `--pai-source` with no error, unlike the `--id`+`--digest` combination which throws. Silent flag-drop is a clarity trap.
  Fix: Reject `--pai-source` together with `--id`, mirroring the `--id`+`--digest` guard.

- **[suggestion]** `src/algorithm-reflection-import.ts:47` — **Imported learnGateClean derived from PAI completeness flag**
  `learnGateClean: bool(doctrine.completeness_gate_met)` maps the LEARN evidence gate (#330) from PAI's `completeness_gate_met`, conflating it with the distinct `completeness` flag computed below from criteria counts. The two Soma gates end up sourced from semantically different PAI signals.
  Fix: Comment why completeness_gate_met stands in for the evidence gate, or leave learnGateClean unknown when PAI has no analogous signal.

### Architecture

- **[suggestion]** `src/types.ts:240` — **gatesFired carries two provenance semantics with no discriminator**
  `AlgorithmGatesFired` holds both `computeGatesFired` re-derived facts and the importer's "documented, lossy best-effort mapping" from PAI `doctrine_fired`, and the digest "reads both uniformly" — so caller-asserted history ranks with the same weight as the deterministic spine, with nothing at the type level to separate them. Future need to weight or filter by provenance would require a type change across recorder, importer, and digest.
  Fix: Add a provenance marker (e.g. `gatesSource: "computed" | "imported"`) on AlgorithmMetaReflection so downstream consumers can distinguish re-derived from asserted flags.

### HonestOracle

- **[important]** `src/algorithm-reflection-import.ts:54` — **Imported #330 LEARN-gate flag is filled from PAI's completeness gate**
  `gatesFromPai` sets `learnGateClean: bool(doctrine.completeness_gate_met)`, but types.ts documents that flag as "learnGateClean: no unresolved or hollow-`specified` criteria blocked LEARN (#330)" — a different gate. Since gate-miss counts RANK the backlog, imported #330 misses are attributed from data that actually measures PAI completeness, silently mislabeling the ranking spine the PR sells as gate-fidelity.

- **[suggestion]** `src/algorithm-reflection-import.ts:55` — **`completeness` defined differently for imported vs live, merged as one column**
  Imported completeness is `passed === count && (failed ?? 0) === 0`; live `computeGatesFired` uses `criteria.every(isClosedCriterion)` (dropped/deferred count as closed). The import doc claims "The digest reads both uniformly," yet a dropped criterion is a completeness pass live but a miss when imported — divergent semantics under one ranked label.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/cli/algorithm.ts:866` — **Bare `reflections` silently runs the cross-run digest**
  After the `--id` branch, control falls through to `// Cross-run digest (default, or explicit --digest)`, so `soma algorithm reflections` with no flags produces a digest even though the help text frames `--digest` as the explicit trigger (`--id lists one run's reflections; --digest ranks`). The no-arg default isn't documented in the usage string.
  Fix: Either require `--id` or `--digest` explicitly, or note the default-to-digest behavior in the `reflections` usage help.

### HonestOracle

- **[important]** `src/algorithm.ts:292` — **Digest conflates 'gate not yet reached' with 'gate missed'**
  `recordAlgorithmMetaReflection` calls `computeGatesFired(run)` with no phase guard, and the test records a reflection while `completeness: false // C1 still open`. The digest then counts every `false` as a gate-miss, but types.ts asserts "a gate recurrently unmet across runs is the empirical case for a runner/prompt fix" — an in-progress run's unmet gate is not a should-have, so premature reflections inflate the backlog the feature exists to rank.

- **[suggestion]** `src/algorithm-reflection-import.ts:8` — **4-record fixture equated with the 188-run finding**
  The description claims the sample "re-surfaces P2 (#331), the top finding of the manual 188-run dig" — but an author-built 4-line fixture where 3/4 records were set to `live_probe:false` can only echo that finding, not independently re-derive it; the import comment likewise asserts the corpus "surfaces P2" while the only richly-populated gate is currentStateFloor, making the result structurally preordained.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/algorithm-reflection-digest.ts:25` — **Abandoned phase ordered after all gates inflates gate-misses**
  `PHASE_ORDER = [..."learn","complete","abandoned"]` places `abandoned` last, so `gateReached` treats a run abandoned at OBSERVE as having reached every gate — its all-`false` flags then count as three misses. Abandoned is terminal/orthogonal, not past LEARN.
  Fix: Exclude `abandoned` reflections from gate-miss counting, or map abandoned to the phase actually reached before abandonment.

### Architecture

- **[suggestion]** `src/algorithm-reflection-digest.ts:28` — **GATE_ENFORCED_AT duplicates harness gate→phase bindings**
  `GATE_ENFORCED_AT = { currentStateFloor: "think", learnGateClean: "learn", completeness: "learn" }` hardcodes which phase enforces each gate, independently of the harness that actually enforces them (algorithm.ts predicates + advance logic). If a gate's enforcement phase moves, `gateReached` drifts silently and mis-ranks the backlog.
  Fix: Derive the gate→phase map from the harness's gate definitions (single source) instead of re-declaring it in the digest module.

### HonestOracle

- **[important]** `src/algorithm-reflection-digest.ts:24` — **Abandoned-phase placement defeats the tested 'reached the gate' invariant**
  `PHASE_ORDER` ends with `"complete", "abandoned"` and `gateReached` uses `indexOf(phase) >= indexOf(enforcedAt)`, so a reflection with `phase: "abandoned"` scores index 8 ≥ every gate and counts as missing all three — even if the run was abandoned at observe. This silently re-introduces the exact 'false ≠ not-reached' conflation the guard and its test claim to prevent, polluting the ranked backlog with phantom misses.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/algorithm-reflection-import.ts:96` — **Imported satisfaction bypasses the 0–10 validation**
  `const satisfaction = num(record.satisfaction_prediction)` stores any finite number, while the live `recordAlgorithmMetaReflection` rejects satisfaction outside 0–10. A PAI record with `satisfaction_prediction: 50` lands in `metaReflection` violating the type's documented "0–10" contract.
  Fix: Clamp or drop out-of-range values: `satisfaction !== undefined && satisfaction >= 0 && satisfaction <= 10`.

### HonestOracle

- **[important]** `src/algorithm-reflection-import.ts:95` — **Imported records hard-code phase:"learn", faking gate-reach for every PAI run**
  `phase: "learn"` is stamped on every imported record, so the digest's `gateReached` treats all of them as having reached every gate — counting completeness/current-state misses even for PAI runs that abandoned before LEARN. This inflates the deterministic gate-miss ranking that the PR cites as 'automatically re-surfacing P2'; the import docstring documents the flag mapping but never this phase assumption.


---

jcfischer (commented):

### CodeQuality

- **[nit]** `src/cli/algorithm.ts:524` — **--within-budget and --over-budget accept each other silently**
  `case "--within-budget": options.withinBudget = true;` and `case "--over-budget": options.withinBudget = false;` are independent flags; passing both resolves to whichever parses last with no conflict error, so a contradictory invocation is accepted silently.
  Fix: Throw if withinBudget is already set when the opposite flag is seen.

- **[suggestion]** `src/algorithm-reflection-import.ts:99` — **Imported reflection timestamp defaults to empty string**
  `timestamp: str(record.timestamp) ?? ""` stores an empty string as a timestamp when a PAI record lacks one. An empty timestamp is a malformed value that renders blank if such a record ever reaches the single-run listing or learning-file mirror.
  Fix: Skip the record or fall back to a sentinel that reads as unknown rather than an empty timestamp.

### HonestOracle

- **[important]** `src/algorithm-reflection-digest.ts:7` — **Claims gate-miss digest is 'how P2/#331 was found' — it wasn't**
  Docstring states "a gate recurrently unmet across runs is the empirical case for a runner/prompt fix (this is exactly how P2 / #331 was found)" — repeated in types.ts:231 — but the PR description says P2 was "the top finding of the manual 188-run dig," and #330/#331's gates did not exist during the PAI runs (imported records hard-code learnGateClean=true). The digest re-surfaces P2 post-hoc on a 4-record hand-picked sample; presenting that as the discovery method misleads readers about how the finding was actually validated.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/algorithm-reflection-import.ts:87` — **JSON `null` line crashes parser despite skip contract**
  Only `JSON.parse` is wrapped in try/catch; a line containing `null` parses to `null`, then `str(record.prd_id)` dereferences null and throws uncaught — violating the documented "Malformed lines are skipped, not fatal" guarantee for a historical artifact.
  Fix: After parse, guard: `if (record === null || typeof record !== "object") continue;`

### HonestOracle

- **[important]** `src/algorithm-reflection-import.ts:83` — **Gate-miss 'deterministic spine' is gated by prose presence**
  PR/header sells a guarantee: 'gate-miss counts RANK, q-signal keyword buckets ENRICH ... they can't corrupt the backlog' and 'the deterministic gatesFired flags drive the ranking.' But `if (Object.keys(smarterRun).length === 0) continue;` drops any PAI record lacking prose BEFORE its gate-flags are counted, and live `recordAlgorithmMetaReflection` likewise throws unless ≥1 smarterRun signal is present. So a run that missed `live_probe` but recorded no q-signal never contributes a gate-miss — prose presence gates the very 'deterministic spine' claimed to be independent of it.

- **[suggestion]** `src/algorithm-reflection-import.ts:73` — **Silently-dropped records leave digest ratios uncountable**
  `catch { continue; // skip malformed line }` plus the no-signal skip discard corpus records silently, yet `renderReflectionDigest` presents the ranking as authoritative ('3 of 4 runs missed live_probe') with no count of dropped lines. The denominator behind the headline gate-miss ratio is unknown to the reader.


---

jcfischer (commented):

2 finding(s): 2 important; 1 lens(es) failed to run: CodeQuality.


### CodeQuality — DID NOT RUN

  claude extraction recovered JSON but it is not lens-shaped (no `summary` or `findings` field) — extractor=raw

### HonestOracle

- **[important]** `src/algorithm-reflection-import.ts:52` — **Absent live_probe silently scored as a current-state miss, manufacturing the headline P2 result**
  `currentStateFloor: bool(doctrine.live_probe)` with `bool = value === true` scores any PAI record lacking a true `live_probe` as a miss; combined with the hard-coded `phase:"learn"`, every such record counts toward the current-state bucket the PR holds up as 'P2 re-surfaced automatically.' The ranking thus reflects how PAI happened to populate one optional field, not an independent rediscovery — the evidence is biased toward the conclusion it claims to find.


---

jcfischer (commented):

No findings. Sage approves.



---

PR #343, opened 2026-06-24
feat(vsa): de-isa the on-disk storage/wire surfaces (#329 slice 3, Piece A)

jcfischer:
Closes the storage/wire half of #329 slice 3. Everything user-facing already says VSA (slices 1+2); this renames the four deliberately-deferred **on-disk / wire** spellings, each with real back-compat (dual-read / migration), not sed.

## Surfaces (one commit each)

1. **`~/.soma/isa/` compartment dir → `vsa/`** — `vsaDir()` dual-reads (`vsa/` else legacy `isa/` else `vsa/`). New `migrateVsaStorageDir` snapshot-first renames `isa/`→`vsa/`, best-effort + idempotent, hooked at the top of `bootstrapSomaHome` so it runs on install/reproject/upgrade/init **before** the compartment mkdir recreates the dir. policy-path-guard allows modify on both `vsa/` and legacy `isa/`.
2. **`AlgorithmRun.isa` field → `vsa`** (live TS field + on-disk JSON key, ~12 consumers). `loadAlgorithmRun` dual-reads the embedded key (`vsa ?? isa`): v2 (isa key) upgrades to v3 (vsa key, stale `isa` stripped); v1 legacy migrate preserved; v3 reads native. `schemaVersion` bumped 2→3 (on-disk shape changed); reader still accepts 1/2.
3. **`lifecycle.isa_updated` events → `vsa_updated`** + sibling event-kind wire names (`writeback.isa_log`, `algorithm.isa_route`, `algorithm.isa_hint` → `vsa_*`). Emit-only — no src consumer matches these strings; historical `events.jsonl` stays parseable. `SomaLifecycleEventName` keeps `isa_updated` as a legacy member.
4. **work.json `isa` session field → `vsa`** — dual-read `entry.vsa ?? entry.isa`; `buildRegistryEntry` hoists `artifacts.vsa ?? artifacts.isa` → top-level `vsa`; session-harvester dual-reads the pointer. Artifacts-map keys stay caller-controlled.

## Back-compat
Existing homes (legacy `isa/` dir, `isa`-field runs, `isa_updated` events, `isa` work.json pointers) load unchanged via dual-read; `soma upgrade` migrates the dir (snapshot-first, reversible). New tests cover: dir migration + dual-read, v1→v3 / v2-isa-key-upgrade / native-v3 run loads, work.json legacy-isa dual-read + artifacts hoist.

## Verification
- `bun run typecheck` clean
[...]

Issue #329, opened 2026-06-22
Rename vocabulary in code/docs/paths: Telos→Purpose, ISA→Verification/VSA


jcfischer:
Mechanical rename to land the canonical vocabulary already decided in `CONTEXT.md` and `docs/adr/0001-de-miessler-vocabulary.md`. The glossary is the source of truth today; code/docs/paths still use the old PAI-inherited names.

## Decided (already in CONTEXT.md + ADR 0001)

- `Telos` compartment → **`Purpose`**
- `ISA` compartment → **`Verification`**
- `ISA` (Ideal State Artifact) → **`VSA`** (Verification State Artifact)
- Primitive term **`checkpoint`**, direction anchor **`intent`** (these are new glossary terms; `IdealStateCriterion` is the existing code shape that maps to `checkpoint`)
- Algorithm / effort-tier / euphoric-surprise apparatus retired as canonical → optional pack (names can stay in code for now)

Rejected names recorded in the ADR (heading, bearing, Method, Verifact, VA) — do not relitigate here.

## Footprint (measured 2026-06-22)

- **Telos:** 7 typed src files (`interface Telos` in `src/types.ts`), ~16 src files referencing the `telos/` path, the `telos/` home dir, `~/.claude/rules/soma/TELOS.md` projection.
- **ISA:** 44 src files (`Isa*` identifiers, `soma isa` CLI in `src/cli.ts` + `src/cli-isa.ts`, `isa-accessors`/`isa-parse`/`isa-reconcile`/`isa-schema`/`isa-skill-installer`/`algorithm-isa-*`, adapter `isa-checklist`, `adapter-active-isa.ts`), `src/skills/ISA/` (22 files incl. e1–e5 examples), 22 docs.
- **Tests:** ~30 test files touch isa/telos.

## Back-compat (the real work — not pure find-replace)

Existing Soma homes have `~/.soma/telos/` and `~/.soma/isa/` dirs, projected `TELOS.md`, and muscle-memory `soma isa`. The rename must not break installed users:

- [ ] CLI: `soma vsa` is canonical; `soma isa` kept as a deprecated alias (warn, don't remove yet).
- [ ] Home paths: migrate `telos/`→`purpose/` and `isa/`→`verification/` (or VSA dir) on `soma upgrade`, with snapshot-first (existing snapshot/rollback machinery).
- [ ] Projection: emit `PURPOSE.md` / new name; keep reading old projected files during transition.
[...]

PR #344, opened 2026-06-24
feat(vsa): rename IdealStateCriterion type → Checkpoint with deprecated alias (#329 slice 3, Piece B)

jcfischer:
The ISC→`checkpoint` primitive rename for #329 slice 3. **TS-type-only scope** (per JC steer): rename the type, keep a deprecated alias, no markdown/CLI/on-disk change.

## What changed
- `interface IdealStateCriterion` → `interface Checkpoint` — the extracted evidence-gated verification primitive (`{ criterion, required evidence, typed verdict, completion gate }`) per CONTEXT.md + ADR 0001. The existing code shape already *was* a checkpoint.
- `export type IdealStateCriterion = Checkpoint` retained as an `@deprecated` back-compat alias (mirrors slice-1 `Telos`→`Purpose`). Both re-exported from `index.ts` (the deprecated re-export carries a scoped eslint-disable).
- All in-repo consumers migrated to `Checkpoint` (src + test).

## Deliberately unchanged (per chosen scope)
- VSA markdown stays `## Criteria`; criterion ids stay `ISC-N`.
- No CLI vocabulary change.
- **No on-disk migration** — pure type rename, zero runtime behavior change.
- ADR 0001 keeps `ISC` as an optional Algorithm-pack surface; this PR doesn't touch that.

## Verification
- `bun run typecheck` clean
- `bun test` = **1444 pass / 0 fail**
- lint clean on changed files (deprecated-alias re-export suppressed intentionally)

## Relationship to #343
Independent of #343 (Piece A, storage/wire isa→vsa). Both branch off `main`; the only shared file is `src/types.ts` (different regions). Either can merge first.

### DIFF
--- a/src/adapters/pi-dev/extensions/vsa-checklist.ts
+++ b/src/adapters/pi-dev/extensions/vsa-checklist.ts
@@ -12,7 +12,7 @@
  *   dropped  → [-]   (deliberately removed)
  *   pending  → [ ]   (everything else: unverified, in-progress, etc.)
  *
- * Status strings mirror `IdealStateCriterion.status` in `src/types.ts`.
+ * Status strings mirror `Checkpoint.status` in `src/types.ts`.
  * This module deliberately accepts any string status and falls through
  * to "pending" for unrecognized values rather than crashing the widget.
  *
--- a/src/algorithm-execution-modes.ts
+++ b/src/algorithm-execution-modes.ts
@@ -5,7 +5,7 @@ import type {
   AlgorithmNotificationEvent,
   AlgorithmPhase,
   AlgorithmRun,
-  IdealStateCriterion,
+  Checkpoint,
   IdeateParameters,
   IdeatePresetName,
   OptimizeParameters,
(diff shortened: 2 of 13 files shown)


jcfischer (commented):

1 finding(s): 1 suggestion.


### HonestOracle

- **[suggestion]** `src/index.ts:40` — **"All consumers migrated" not provable by the cited verification**
  PR claims "All in-repo consumers migrated to `Checkpoint`" with `bun run typecheck` clean as proof, but `export type IdealStateCriterion = Checkpoint` plus its re-exported alias (`IdealStateCriterion,` here) make any straggler still typecheck silently — so typecheck-clean cannot demonstrate completeness. A reader trusting the claim could later delete the alias unsafely.


---

jcfischer:
Re Sage HonestOracle nit (src/index.ts:40) — fair point that typecheck-clean can't prove completeness while the alias exists. Backing the claim with a grep instead:

\`\`\`
$ grep -rn 'IdealStateCriterion' src test
src/types.ts:59:export type IdealStateCriterion = Checkpoint;   # the deprecated alias definition
src/index.ts:41:  IdealStateCriterion,                          # intentional back-compat re-export
\`\`\`

Zero call-site stragglers — the only two occurrences are the alias declaration and its re-export. So deleting the alias later would surface any future regression at those two sites only. Nit acknowledged; no code change needed.

PR #345, opened 2026-06-24
fix(grok): fail closed when trustedSomaRepo is unusable (close fail-open hole)

jcfischer:
## The hole
The grok pre-tool-use policy chain ran `bun run soma <cmd>` with `cwd = config.trustedSomaRepo`. `bun run soma` resolves the `soma` script by walking up from cwd and, failing that, a **global** `soma` on `PATH` — so a missing/wrong `trustedSomaRepo` silently ran some other soma and could return `allow`. Fail-open.

Reproduced: `bun run soma --version` from an empty dir resolves `~/bin/soma` (`soma 0.8.8`, exit 0).

## The fix
`runSomaCommand` invokes the trusted repo's own declared entrypoint instead of `bun run soma`:
- `resolveSomaEntrypoint(repo)` reads `<repo>/package.json` `scripts.soma` (this repo: `"soma": "bun src/cli.ts"`), extracts the script file, returns `<repo>/<file>`; returns `undefined` if absent.
- `undefined` → `runSomaCommand` returns a synthetic `{ status: 1, stdout: "", stderr }`. The deny-legs read `status` + `stdout||stderr`, so `status !== 0` → deny. Otherwise it runs `bun <entrypoint> <args>` — **no PATH fallback**.

So an unresolvable/non-soma `trustedSomaRepo` denies; a real one runs the configured checkout's own soma.

## Regression test
`grok pre-tool-use denies when trustedSomaRepo is a non-soma dir (no global-soma fallback)` plants a global `soma` on `PATH` that returns `allow`, points `trustedSomaRepo` at a `package.json`-but-no-soma dir, and asserts the hook **denies** — i.e. the resolved-entrypoint path never consults `PATH`, so the planted global is ignored.

## Also fixes a pre-existing flaky test
`grok pre-tool-use fails closed when the soma repo is unusable` failed in **isolation** (`allow`), only passing in orderings that perturbed `PATH`. Verified deterministic now (isolated single-test run passes).

## Verification
- `bun run typecheck` clean
- isolated previously-failing test passes; new bypass-regression test passes
- `bun test` full suite **1454 pass / 0 fail**

`grok-hook-entry.mjs` is hand-authored source, shipped verbatim by `install.ts`.

### DIFF
--- a/src/adapters/grok/hooks/grok-hook-entry.mjs
+++ b/src/adapters/grok/hooks/grok-hook-entry.mjs
@@ -52,8 +52,34 @@ function hookSessionId(input) {
   return typeof candidate === "string" && candidate.length > 0 ? candidate : undefined;
 }
 
+// Resolve soma's entrypoint file from the trusted repo's declared `scripts.soma`
+// (e.g. "bun src/cli.ts" → "<repo>/src/cli.ts"), anchored to trustedSomaRepo.
+// Returns undefined if the repo has no such script — i.e. it isn't a soma
+// checkout. Tracks the declared invocation contract rather than hardcoding a path.
+function resolveSomaEntrypoint(repo) {
+  try {
+    const script = JSON.parse(readFileSync(join(repo, "package.json"), "utf8")).scripts?.soma;
+    if (typeof script !== "string") return undefined;
+    const file = script.split(/\s+/).find((token) => /\.[cm]?[jt]s$/.test(token));
+    return file ? join(repo, file) : undefined;
+  } catch {
+    return undefined;
+  }
+}
+
+// Invoke soma by the trusted repo's OWN entrypoint, NOT `bun run soma`. `bun run
+// soma` resolves the script by walking up from cwd and, failing that, a GLOBAL
+// `soma` on PATH (the repo declares scripts.soma but no bin) — so a missing or
+// wrong trustedSomaRepo would silently run some other soma and could return
+// `allow`. Running the resolved file directly has no PATH fallback: an
+// unresolvable repo yields a synthetic non-zero result, and the caller's
+// `status !== 0` legs fail closed. Ties every policy decision to the trusted repo.
 function runSomaCommand(config, args, env = {}) {
-  return spawnSync(config.bunPath, args, {
+  const entrypoint = resolveSomaEntrypoint(config.trustedSomaRepo);
+  if (entrypoint === undefined) {
+    return { status: 1, stdout: "", stderr: `trusted soma repo unusable: no soma entrypoint under ${config.trustedSomaRepo}` };
+  }
+  return spawnSync(config.bunPath, [entrypoint, ...args], {
     cwd: config.trustedSomaRepo,
     encoding: "utf8",
     timeout: 25000,
@@ -62,7 +88,7 @@ function runSomaCommand(config, args, env = {}) {
 }
 
 function runSomaLifecycle(config, event, sessionId) {
-  const args = ["run", "soma", "lifecycle", event, "--soma-home", config.somaHome, "--substrate", "grok"];
+  const args = ["lifecycle", event, "--soma-home", config.somaHome, "--substrate", "grok"];
   if (sessionId) {
     args.push("--session-id", sessionId);
   }
@@ -71,13 +97,11 @@ function runSomaLifecycle(config, event, sessionId) {
 }
 
 function runSomaClassification(config, prompt) {
-  return runSomaCommand(config, ["run", "soma", "algorithm", "classify", "--prompt", prompt || "", "--json"]);
+  return runSomaCommand(config, ["algorithm", "classify", "--prompt", prompt || "", "--json"]);
 }
 
 function runSomaPolicyCheck(config, targets, action = "write") {
   const args = [
-    "run",
-    "soma",
     "policy",
     "check",
     "--soma-home",
@@ -101,8 +125,6 @@ function runSomaPolicyCheck(config, targets, action = "write") {
 
 function runSomaInboundContentScan(config, target) {
   return runSomaCommand(config, [
-    "run",
-    "soma",
     "policy",
     "scan",
     "--soma-home",
@@ -119,8 +141,6 @@ function runSomaInboundContentScan(config, target) {
 
 function runSomaRuntimePolicyInspect(config, surface, payload) {
   const args = [
-    "run",
-    "soma",
     "policy",
     "inspect",
     "--soma-home",
--- a/test/grok-hook.test.ts
+++ b/test/grok-hook.test.ts
@@ -1239,6 +1239,53 @@ test("grok pre-tool-use fails closed when the soma repo is unusable", async () =
   });
 });
 
+test("grok pre-tool-use denies when trustedSomaRepo is a non-soma dir (no global-soma fallback)", async () => {
+  // The exploit the fix closes: before, `bun run soma` from a dir without a
+  // local soma resolved a GLOBAL soma on PATH and returned allow. A dir that
+  // looks like a project (has package.json) but has no soma entrypoint must
+  // still deny — the hook runs <repo>/src/cli.ts explicitly, with no PATH fallback.
+  await withTempHome(async (homeDir) => {
+    const { chmod } = await import("node:fs/promises");
+    await installSomaForGrok({ homeDir });
+    const hook = join(homeDir, ".grok/hooks/soma-lifecycle.mjs");
+    const configPath = join(homeDir, ".grok/hooks/soma-lifecycle.config.json");
+    const config = JSON.parse(await readFile(configPath, "utf8"));
+    const nonSomaRepo = join(homeDir, "not-soma");
+    await mkdir(nonSomaRepo, { recursive: true });
+    await writeFile(join(nonSomaRepo, "package.json"), JSON.stringify({ name: "not-soma" }), "utf8");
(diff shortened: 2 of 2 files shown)


jcfischer (commented):

### HonestOracle

- **[important]** `src/adapters/grok/hooks/grok-hook-entry.mjs:413` — **package.json check is a surrogate for "is a soma checkout"**
  Code gate is `!existsSync(join(repo, "package.json"))`, yet the comment and description claim it catches a repo that "isn't a soma checkout" and that the change "closes the hole." Any directory with a package.json (not necessarily soma) passes the gate, so `bun run soma` can still resolve global soma — the fail-open path the PR claims to have closed survives for that case.

- **[suggestion]** `src/adapters/grok/hooks/grok-hook-entry.mjs:417` — **"no happy-path change" rests on unverified assumption**
  Description asserts "Fail-closed direction only — no happy-path change (a real soma repo has package.json)"; the diff cannot verify that every valid trustedSomaRepo contains a top-level package.json, so a legitimate-but-unconventional checkout could now be denied.

- **[suggestion]** `src/adapters/grok/hooks/grok-hook-entry.mjs:1` — **"rendered .mjs, shipped verbatim" implies untracked source**
  Description says "rendered `.mjs` hook, shipped verbatim," implying it is generated; the diff edits the generated output with no shown change to a source template, so the claim gives no evidence the fix survives the next render.


---

jcfischer (commented):

4 finding(s): 1 important, 2 suggestion, 1 nit.


### CodeQuality

- **[nit]** `src/adapters/grok/hooks/grok-hook-entry.mjs:430` — **Missing-name case yields 'package "undefined"' message**
  When package.json parses but has no `name`, `name !== "soma"` is true and the message reads `is not a soma checkout (package "undefined")`. Correct fail-closed behavior, but the literal `undefined` reads as a bug.
  Fix: Special-case a missing name: `package "${name ?? "<none>"}"`.

### HonestOracle

- **[important]** `src/adapters/grok/hooks/grok-hook-entry.mjs:412` — **"provably comes from it" overclaims what the name check proves**
  Comment asserts the name check ties the decision so it "provably comes from" the repo, and the description says "a global soma on PATH can't satisfy the gate." Reading `package.json.name === "soma"` proves only what the file declares, not that `bun run soma` resolves the local soma rather than a global one — a repo named "soma" with no local soma script still falls back to PATH, so the hole the comment claims to close can persist.

- **[suggestion]** `src/adapters/grok/hooks/grok-hook-entry.mjs:408` — **Green suite offered as proof the fail-open hole is closed**
  Verification cites "bun test full suite green; grok-hook 49 pass / 0 fail" and one "repo unusable" test, but the named tests cover an unusable/unset repo — not the actual exploit (a non-soma directory that previously resolved a global soma to `allow` now being denied). No test reproduces the reported bypass, so "close fail-open hole" rests on a manual one-off repro, not a regression guard.

- **[suggestion]** `src/adapters/grok/hooks/grok-hook-entry.mjs:449` — **Flaky-test root cause stated as fact without evidence**
  Description claims the test "only passed in full-suite orderings where another test perturbed PATH/env" as the diagnosed cause, but the diff shows no such test or env interaction; the assertion is an unverifiable causal story attached to the now-deterministic pass.


---

jcfischer (commented):

### CodeQuality

- **[nit]** `src/adapters/grok/hooks/grok-hook-entry.mjs:422` — **Error message renders "undefined" when name field absent**
  When package.json parses but has no `name`, `name !== "soma"` is true and the message interpolates `package "${name}"` as `package "undefined"`, which reads as a bug rather than a missing field.
  Fix: Distinguish missing name: `package "${name ?? "(no name)"}"`.

### HonestOracle

- **[important]** `src/adapters/grok/hooks/grok-hook-entry.mjs:449` — **Determinism claim verified by the ordering that always passed**
  Description: flake 'failed in isolation (got allow) and only passed in full-suite orderings'; Verification cites only 'bun test full suite green' — the exact ordering that masked the flake. Full-suite-green is offered as proof of 'Now deterministic' but cannot demonstrate the isolation case now passes; the isolated/EX-00003 run is the one that needed showing.

- **[suggestion]** `src/adapters/grok/hooks/grok-hook-entry.mjs:411` — **'provably comes from configured checkout' exceeds the name check**
  Comment claims requiring name 'soma' makes the decision 'provably comes from it'; but the check validates package.json name only — it does not establish that `bun run soma` resolves the repo-local binary rather than a global PATH soma. 'Provably' overstates what the name field shows.


---

jcfischer (commented):

### EcosystemCompliance

- **[suggestion]** `src/adapters/grok/hooks/grok-hook-entry.mjs:61` — **Hook hardcodes src/cli.ts, bypassing declared scripts.soma entrypoint**
  `const entrypoint = join(config.trustedSomaRepo, "src/cli.ts")` couples the hook to soma's internal layout; the PR itself notes package.json defines `scripts.soma` as the canonical entrypoint. If soma relocates its entrypoint, this hook silently diverges from the declared invocation contract.
  Fix: Resolve the entrypoint from the trusted repo's package.json `scripts.soma` (still anchored to trustedSomaRepo, no PATH fallback) so the hook tracks the declared contract.

### HonestOracle

- **[important]** `test/grok-hook.test.ts:1242` — **Regression test never plants the global soma it claims to guard against**
  The test comment asserts "before, `bun run soma`... resolved a GLOBAL soma on PATH and returned allow" and that this dir is "the exact prior bypass, now denies" — but the test only creates `package.json` (name `not-soma`) and no global soma anywhere on PATH. Against the old code with no soma on the test PATH, `bun run soma` also fails and denies, so the test would pass against the vulnerable code and proves nothing about closing the fallback.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/adapters/grok/hooks/grok-hook-entry.mjs:60` — **Entrypoint regex picks first file-token, not the script's**
  `const file = script.split(/\s+/).find((token) => /\.[cm]?[jt]s$/.test(token));` returns the FIRST token ending in a JS/TS extension. A chained or multi-arg `scripts.soma` (e.g. `bun build.ts && bun src/cli.ts`) resolves to `build.ts` and runs the wrong entrypoint.
  Fix: Take the file token immediately following the `bun` interpreter token, or assert exactly one file token, instead of `.find` first-match.

### Security

- **[suggestion]** `src/adapters/grok/hooks/grok-hook-entry.mjs:60` — **Entrypoint regex picks first file-like token, not the real one**
  `script.split(/\s+/).find((token) => /\.[cm]?[jt]s$/.test(token))` returns the first extension-matching token; a scripts.soma like `bun --preload ./setup.ts src/cli.ts` would resolve `setup.ts` and run the wrong file as the policy engine. The repo is trusted, so this is hardening, not an open hole.
  Fix: Take the last file-like token, or require it to be the token after the runner (e.g. the arg following `bun`).

### EcosystemCompliance

- **[suggestion]** `src/adapters/grok/hooks/grok-hook-entry.mjs:60` — **Entrypoint resolution couples hook to unwritten scripts.soma shape**
  `script.split(/\s+/).find((token) => /\.[cm]?[jt]s$/.test(token))` assumes `scripts.soma` is `bun <file>` with exactly one source-file token; if soma's package.json later expresses the soma script differently (bin alias, chained command, build step), the hook resolves undefined and denies everything. This treats a parse heuristic as the ecosystem contract.
  Fix: Document the `scripts.soma` = `bun <entrypoint>` shape as a contract soma's package.json must honor, or assert the resolved file exists before relying on it.

### HonestOracle

- **[important]** `src/adapters/grok/hooks/grok-hook-entry.mjs:56` — **Description says hardcoded src/cli.ts; code derives it from package.json**
  The description claims the fix "invokes the explicit file `<trustedSomaRepo>/src/cli.ts`" and "removes the resolution path entirely," but `resolveSomaEntrypoint` still parses `JSON.parse(readFileSync(join(repo, "package.json"))).scripts?.soma` and resolves the path from it — the trust anchor is package.json's `scripts.soma`, not a fixed file. A reviewer trusting the description would misjudge what controls which binary runs.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/adapters/grok/hooks/grok-hook-entry.mjs:56` — **Synthetic result diverges from spawnSync shape**
  The short-circuit returns `{ status: 1, stdout: "", stderr: ... }`, omitting `signal`, `error`, `output`, and `pid` that real spawnSync results carry. Callers reading any field beyond `status`/`stdout` would see undefined for the unresolvable-repo path.
  Fix: Include the missing fields (e.g. `signal: null, error: undefined`) or document that callers must branch on `status` only.

### HonestOracle

- **[important]** `src/adapters/grok/hooks/grok-hook-entry.mjs:78` — **"Callers' status!==0 legs fail closed" generalized across 5 callers, none shown**
  Comment asserts "the caller's `status !== 0` legs fail closed" but `runSomaCommand` feeds five callers — lifecycle, classify, policy check, scan, inspect — and the diff shows none of their status-handling. Lifecycle/classify are not policy gates; if they degrade silently on non-zero rather than deny, the "fail closed" headline holds only for the policy path, not the whole hook as claimed.

- **[suggestion]** `test/grok-hook.test.ts:1245` — **"Fails against the old code" asserted but red-against-old never shown**
  Description and test comment claim the regression "fails against the old `bun run soma` code" yet the Verification section lists only "new bypass-regression test passes" against the fix — no run against reverted code is shown. A regression test's worth is the red-before-green; that red is asserted, not demonstrated.


---

PR #351, opened 2026-06-24
feat(install): self-reconciling generated projections + pi-dev skill name (case-clean — part 1)

jcfischer:
**Part 1** of making projection case-clean. Lands the reconcile mechanism for Soma's *generated* subtrees and fixes the pi-dev skill name. Does **NOT** yet remove orphaned *skill* dirs (the visible `ISA/` leftovers) — that's part 2.

## What this PR fixes (and what it doesn't)
✅ **Generated** Soma-owned subtrees (`rules/soma`, `memories/soma`, `hooks/soma`, `agent/soma`) self-reconcile: renamed/recased/removed projected files pruned + case-normalized, identically on both FS.
✅ pi-dev VSA skill projects as canonical `vsa` (was stale `isa`).
❌ Orphaned **skill** dirs (`~/.soma/skills/ISA`, projected `ISA/` in codex/claude) — still present after this PR; fixed in part 2.

## Mechanism
- `reconcileOwnedDir(root, desired, {excludeRelPrefixes})` — keep exact; case-normalize (crash-safe temp-hop rename on same-inode/case-insensitive, else remove distinct); remove other; prune empty dirs via `rmdir` (TOCTOU-safe). **It DELETES**: sound only for fully-generated, exclusively-owned dirs; a projection bug dropping a file would silently delete it (recover via `soma snapshot`). Guards: resolve vs substrateRoot, skip on empty desired set, `sameFile`/catches ENOENT-only.
- `removeObsoleteHomeFiles` made recursive. claude/cursor `obsoleteHomeFiles` for TELOS.md/ACTIVE_ISA.md dropped (subsumed by reconcile).
- Cursor's VSA skill nests under its owned subtree; excluded from reconcile via the auto-derived `vsaSkillProjection.destinationDir` (and `removeEmptyDirs` honors that exclusion).
- Owned subtrees may be Soma private/protected roots BY DESIGN — reconcile is Soma's own authorized projection (same trust as the writes here + the pre-existing obsoleteHomeFiles deletes). `policy-path-guard` governs agent Bash commands, not install fs ops.

## pi-dev skill name
`PI_DEV_VSA_SKILL_ID` `isa`→`vsa`; `removeLegacyPiDevVsaSkillProjection` prunes prior names `LEGACY_PI_DEV_VSA_SKILL_DIRS = ["VSA", "isa"]` (pi-dev lowercases ids, so it never produced a capital `ISA`). New dir name → fresh clean content.

## Scope of "no per-rename bookkeeping"
True for the generated-subtree reconcile. NOT for skills: the pi-dev legacy list is an explicit **interim** per-rename list, removed when part 2's projected-skills manifest generalizes skill pruning.

## Test coverage
- standalone codex test: stale pruned + projected file intact + shared dir untouched.
- parametrized loop over cursor / pi-dev / claude-code `rules/soma` / claude-code `hooks/soma`: each asserts stale pruned AND a real projected file survives intact (an over-pruning regression can't pass).
- unit: stale-removal, empty-dir prune, idempotent, case-variant per-branch outcomes.
- Cross-FS: temp-hop rename branch runs on case-insensitive hosts (dev/macOS), remove-distinct on case-sensitive (Linux CI); post invariant asserted on both. Full suite **1469 pass / 0 fail**, typecheck + lint clean.

## Part 2 (follow-up, decisions.md)
Per-substrate projected-skills manifest → prune Soma-recorded skill dirs no longer projected (`ISA/`), + `~/.soma/skills/ISA` source cleanup. Then a reproject cleans every home and the pi-dev legacy list is removed.

### DIFF
--- a/src/adapters/claude-code/install.ts
+++ b/src/adapters/claude-code/install.ts
@@ -23,6 +23,9 @@ export const claudeCodeInstallSpec: SubstrateInstallSpec<"claude-code"> = {
   // ACTIVE_VSA.md. Claude Code auto-loads rules/soma/, so an upgrade must delete
   // the stale copies or it would keep loading frozen old content every session.
   obsoleteHomeFiles: ["rules/soma/TELOS.md", "rules/soma/ACTIVE_ISA.md"],
+  // Soma-exclusive subtrees — reconciled to the projected set each install so any
+  // renamed/recased/removed projection self-cleans (no per-rename bookkeeping).
+  ownedSubtrees: ["rules/soma", "hooks/soma"],
   optionalHomeFiles: (options) => isClaudeCodeInstallOptions(options) && options.modeClassifier === true
     ? [SOMA_CLAUDE_MODE_CLASSIFIER_RELATIVE_PATH, SOMA_CLAUDE_MODE_CLASSIFIER_CONFIG_RELATIVE_PATH]
     : [],
--- a/src/adapters/claude-code/install.ts
+++ b/src/adapters/claude-code/install.ts
@@ -19,11 +19,8 @@ export const claudeCodeInstallSpec: SubstrateInstallSpec<"claude-code"> = {
     SOMA_CLAUDE_HOOK_CONFIG_RELATIVE_PATH,
     "settings.json",
   ],
-  // Soma-exclusive subtrees — reconciled to the projected set each install so any
-  // renamed/recased/removed projection self-cleans (no per-rename bookkeeping).
-  // This subsumes the former obsoleteHomeFiles for TELOS.md/ACTIVE_ISA.md (both
-  // live under rules/soma); obsoleteHomeFiles is reserved for stale files in
-  // SHARED, non-owned dirs only.
+  // Owned (Soma-exclusive) dirs — see ownedSubtrees JSDoc. Subsumes the former
+  // obsoleteHomeFiles for TELOS.md/ACTIVE_ISA.md, which live under rules/soma.
   ownedSubtrees: ["rules/soma", "hooks/soma"],
   optionalHomeFiles: (options) => isClaudeCodeInstallOptions(options) && options.modeClassifier === true
     ? [SOMA_CLAUDE_MODE_CLASSIFIER_RELATIVE_PATH, SOMA_CLAUDE_MODE_CLASSIFIER_CONFIG_RELATIVE_PATH]
--- a/src/adapters/claude-code/install.ts
+++ b/src/adapters/claude-code/install.ts
@@ -19,12 +19,11 @@ export const claudeCodeInstallSpec: SubstrateInstallSpec<"claude-code"> = {
     SOMA_CLAUDE_HOOK_CONFIG_RELATIVE_PATH,
     "settings.json",
   ],
-  // soma#329: projections renamed TELOS.md → PURPOSE.md and ACTIVE_ISA.md →
-  // ACTIVE_VSA.md. Claude Code auto-loads rules/soma/, so an upgrade must delete
-  // the stale copies or it would keep loading frozen old content every session.
-  obsoleteHomeFiles: ["rules/soma/TELOS.md", "rules/soma/ACTIVE_ISA.md"],
   // Soma-exclusive subtrees — reconciled to the projected set each install so any
   // renamed/recased/removed projection self-cleans (no per-rename bookkeeping).
+  // This subsumes the former obsoleteHomeFiles for TELOS.md/ACTIVE_ISA.md (both
+  // live under rules/soma); obsoleteHomeFiles is reserved for stale files in
+  // SHARED, non-owned dirs only.
   ownedSubtrees: ["rules/soma", "hooks/soma"],
   optionalHomeFiles: (options) => isClaudeCodeInstallOptions(options) && options.modeClassifier === true
     ? [SOMA_CLAUDE_MODE_CLASSIFIER_RELATIVE_PATH, SOMA_CLAUDE_MODE_CLASSIFIER_CONFIG_RELATIVE_PATH]
--- a/src/adapters/codex/install.ts
+++ b/src/adapters/codex/install.ts
@@ -64,6 +64,9 @@ export const codexInstallSpec: SubstrateInstallSpec<"codex"> = {
   substrate: "codex",
   defaultHome: CODEX_DEFAULT_HOME,
   homeFiles: CODEX_HOME_FILES,
+  // Soma-exclusive subtree — reconciled to the projected set each install. Only
+  // memories/soma is fully owned; codex hooks/ and skills/ are shared.
+  ownedSubtrees: ["memories/soma"],
   vsaSkillProjection: {
     destinationDir: vsaSkillUnder(),
   },
--- a/src/adapters/codex/install.ts
+++ b/src/adapters/codex/install.ts
@@ -64,8 +64,7 @@ export const codexInstallSpec: SubstrateInstallSpec<"codex"> = {
   substrate: "codex",
   defaultHome: CODEX_DEFAULT_HOME,
   homeFiles: CODEX_HOME_FILES,
-  // Soma-exclusive subtree — reconciled to the projected set each install. Only
-  // memories/soma is fully owned; codex hooks/ and skills/ are shared.
+  // Owned (Soma-exclusive) dir — see ownedSubtrees JSDoc. (hooks/ + skills/ are shared.)
   ownedSubtrees: ["memories/soma"],
   vsaSkillProjection: {
     destinationDir: vsaSkillUnder(),
--- a/src/adapters/cursor/install.ts
+++ b/src/adapters/cursor/install.ts
@@ -49,10 +49,9 @@ export const cursorInstallSpec: SubstrateInstallSpec<"cursor"> = {
   substrate: "cursor",
   defaultHome: ".",
   homeFiles: CURSOR_HOME_FILE_PATHS,
-  // soma#329: projections renamed TELOS.md → PURPOSE.md and ACTIVE_ISA.md →
-  // ACTIVE_VSA.md; drop the stale copies on reproject/upgrade.
-  obsoleteHomeFiles: [".cursor/rules/soma/TELOS.md", ".cursor/rules/soma/ACTIVE_ISA.md"],
-  // Soma-exclusive subtree — reconciled to the projected set each install.
+  // Soma-exclusive subtree — reconciled to the projected set each install, which
+  // subsumes the former obsoleteHomeFiles for TELOS.md/ACTIVE_ISA.md (both under
+  // .cursor/rules/soma). obsoleteHomeFiles is reserved for shared, non-owned dirs.
   ownedSubtrees: [".cursor/rules/soma"],
   vsaSkillProjection: {
     destinationDir: vsaSkillUnder(".cursor/rules/soma"),
--- a/src/adapters/cursor/install.ts
+++ b/src/adapters/cursor/install.ts
@@ -49,9 +49,8 @@ export const cursorInstallSpec: SubstrateInstallSpec<"cursor"> = {
   substrate: "cursor",
   defaultHome: ".",
   homeFiles: CURSOR_HOME_FILE_PATHS,
-  // Soma-exclusive subtree — reconciled to the projected set each install, which
-  // subsumes the former obsoleteHomeFiles for TELOS.md/ACTIVE_ISA.md (both under
-  // .cursor/rules/soma). obsoleteHomeFiles is reserved for shared, non-owned dirs.
+  // Owned (Soma-exclusive) dir — see ownedSubtrees JSDoc. Subsumes the former
+  // obsoleteHomeFiles for TELOS.md/ACTIVE_ISA.md under .cursor/rules/soma.
   ownedSubtrees: [".cursor/rules/soma"],
   vsaSkillProjection: {
     destinationDir: vsaSkillUnder(".cursor/rules/soma"),
--- a/src/adapters/cursor/install.ts
+++ b/src/adapters/cursor/install.ts
@@ -52,6 +52,8 @@ export const cursorInstallSpec: SubstrateInstallSpec<"cursor"> = {
   // soma#329: projections renamed TELOS.md → PURPOSE.md and ACTIVE_ISA.md →
   // ACTIVE_VSA.md; drop the stale copies on reproject/upgrade.
   obsoleteHomeFiles: [".cursor/rules/soma/TELOS.md", ".cursor/rules/soma/ACTIVE_ISA.md"],
+  // Soma-exclusive subtree — reconciled to the projected set each install.
+  ownedSubtrees: [".cursor/rules/soma"],
   vsaSkillProjection: {
     destinationDir: vsaSkillUnder(".cursor/rules/soma"),
   },
--- a/src/adapters/pi-dev/install.ts
+++ b/src/adapters/pi-dev/install.ts
@@ -40,6 +40,9 @@ export const piDevInstallSpec: SubstrateInstallSpec<"pi-dev"> = {
   substrate: "pi-dev",
   defaultHome: PI_DEV_DEFAULT_HOME,
   homeFiles: PI_DEV_HOME_FILES,
+  // Soma-exclusive subtree — reconciled to the projected set each install.
+  // agent/extensions and agent/skills are shared, so only agent/soma is owned.
+  ownedSubtrees: ["agent/soma"],
   vsaSkillProjection: {
     destinationDir: piDevVsaSkillDestinationDir,
     skillNameOverride: PI_DEV_VSA_SKILL_ID,
--- a/src/adapters/pi-dev/install.ts
+++ b/src/adapters/pi-dev/install.ts
@@ -40,8 +40,7 @@ export const piDevInstallSpec: SubstrateInstallSpec<"pi-dev"> = {
   substrate: "pi-dev",
   defaultHome: PI_DEV_DEFAULT_HOME,
   homeFiles: PI_DEV_HOME_FILES,
-  // Soma-exclusive subtree — reconciled to the projected set each install.
-  // agent/extensions and agent/skills are shared, so only agent/soma is owned.
+  // Owned (Soma-exclusive) dir — see ownedSubtrees JSDoc. (agent/extensions + agent/skills shared.)
   ownedSubtrees: ["agent/soma"],
   vsaSkillProjection: {
     destinationDir: piDevVsaSkillDestinationDir,
--- a/src/adapters/pi-dev/skill-projection.ts
+++ b/src/adapters/pi-dev/skill-projection.ts
@@ -4,7 +4,12 @@ import { rewriteSkillNameFrontmatter } from "../../skill-frontmatter";
 import { rewriteSubstrateProjectionContent } from "../../substrate-projection-rewrites";
 import type { Projection, SomaSkill } from "../../types";
 
-export const PI_DEV_VSA_SKILL_ID = "isa";
+export const PI_DEV_VSA_SKILL_ID = "vsa";
+
+// Prior names the Soma VSA skill was projected under in pi-dev before settling on
+// the canonical lowercase "vsa": legacy capital "VSA", and the pre-rename "ISA"/
+// "isa". All were Soma-projected, so they are pruned before reprojecting "vsa".
+const LEGACY_PI_DEV_VSA_SKILL_DIRS = ["VSA", "ISA", "isa"] as const;
 
 export function piDevSkillId(name: string): string {
   const id = name
@@ -50,10 +55,6 @@ export function piDevVsaSkillDestinationDir(substrateHome: string): string {
   return resolve(substrateHome, "agent/skills", PI_DEV_VSA_SKILL_ID);
 }
 
-function legacyPiDevVsaSkillDestinationDir(substrateHome: string): string {
-  return resolve(substrateHome, "agent/skills/VSA");
-}
-
 export async function removeLegacyPiDevVsaSkillProjection(substrateHome: string): Promise<void> {
   const skillsDir = resolve(substrateHome, "agent/skills");
   let entries;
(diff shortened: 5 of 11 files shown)


jcfischer (commented):

6 finding(s): 2 important, 2 suggestion, 2 nit.


### CodeQuality

- **[suggestion]** `test/install.test.ts:316` — **Vacuous assertion `toContain("")` verifies nothing**
  `expect(await readFile(...)).toContain("")` passes for any string — every string contains the empty string. The reconcile test means to confirm the projected file survives with content, but this asserts nothing beyond what readFile's implicit throw-on-missing already covers.
  Fix: Assert real content, e.g. `.not.toBe("")` or a known substring of startup-context.md.

- **[nit]** `src/adapters/pi-dev/skill-projection.ts:68` — **Redundant `as string` casts in literal guard**
  `if ((legacy as string) === (PI_DEV_VSA_SKILL_ID as string)) continue;` casts both operands only to suppress TS's disjoint-literal comparison warning; the comparison can never be true given the current constants, making the guard dead defensive code.
  Fix: Drop the casts and the guard, or document why it exists for future name overlaps.

### Architecture

- **[suggestion]** `src/adapters/claude-code/install.ts:25` — **obsoleteHomeFiles redundant with ownedSubtrees for in-subtree paths**
  `obsoleteHomeFiles: ["rules/soma/TELOS.md", "rules/soma/ACTIVE_ISA.md"]` sit inside `ownedSubtrees: ["rules/soma", "hooks/soma"]`, which reconcile already prunes since they are no longer projected. Two mechanisms covering the same paths invites drift over which to maintain.
  Fix: Drop obsoleteHomeFiles entries that fall under an ownedSubtree; reserve obsoleteHomeFiles for stale files in shared (non-owned) dirs only.

### Maintainability

- **[nit]** `src/install.ts:226` — **Repeated path-prefix containment idiom**
  The "is X under root" check `abs === root || abs.startsWith(`${root}/`)` appears twice in `reconcileOwnedSubtrees` and again as `rel === prefix || rel.startsWith(`${prefix}/`)` in projection-reconcile.ts. Three near-identical hand-rolled containment checks.
  Fix: Extract a shared `isUnderOrEqual(path, base): boolean` helper and call it in all three spots.

### HonestOracle

- **[important]** `test/projection-reconcile.test.ts:80` — **Cross-FS guarantee half-tested; rename branch skipped on case-sensitive CI**
  Description claims behavior "identically on case-sensitive and case-insensitive filesystems" proven by "full suite 1465 pass / 0 fail". But the sameFile-true rename-via-temp-hop branch in reconcileOwnedDir only runs when `caseInsensitive = await exists(join(root, "purpose.md"))` is true; on a case-sensitive CI host that branch — half the claimed guarantee — never executes, so a green suite there does not cover it.

- **[important]** `test/install.test.ts:314` — **Vacuous assertion in reconcile integration test proves nothing**
  The test named "reproject reconciles an owned subtree" backs the "desired file survives" half with `expect(await readFile(...startup-context.md, "utf8")).toContain("")` — `.toContain("")` is true for any string including empty, so it passes even if reconcile wrongly emptied or corrupted the file. It counts toward "1465 pass" while verifying nothing.


---

jcfischer (commented):

6 finding(s): 2 important, 4 suggestion.


### CodeQuality

- **[suggestion]** `src/install.ts:201` — **Reconcile silently assumes projectedFiles are absolute**
  `new Set(projectedFiles.map((file) => resolve(file)))` resolves any relative entry against process cwd, not substrateRoot; such a path fails `isUnderOrEqual(abs, root)`, producing an empty desired set that would prune the whole owned subtree. Correctness rides on an unstated invariant.
  Fix: Assert/resolve against substrateRoot: `resolve(substrateRoot, file)`, or document that callers must pass absolute paths.

- **[suggestion]** `src/projection-reconcile.ts:33` — **sameFile catch-all routes transient stat errors to deletion**
  `catch { return false }` masks any error (not just ENOENT); on a case-insensitive FS a transient stat failure makes `sameFile` return false, sending the only copy of the file down the `rm(abs)` branch instead of the rename branch.
  Fix: Distinguish ENOENT (legitimately distinct) from other errno and rethrow unexpected errors.

### Architecture

- **[suggestion]** `src/install.ts:230` — **Reconcile exclusion hardcoded to single nested installer**
  `excludeRelPrefixes` is derived solely from `spec.vsaSkillProjection.destinationDir`, so reconcile assumes the VSA skill is the only edit-preserving subtree that can nest under an owned subtree. A future second edit-preserving installer nested under an owned subtree would be silently pruned.
  Fix: Source exclusions from a spec field (e.g. `ownedSubtreeExclusions`) rather than only `vsaSkillProjection.destinationDir`.

### HonestOracle

- **[important]** `src/projection-reconcile.ts:116` — **Reconcile fails open to deletion; PR sells only the upside**
  Desired set = files projected *this run* (`desiredRel` from `allProjectedFiles`), and anything else under the owned subtree hits `await rm(abs, { force: true })`. The PR frames this solely as "self-cleaning" — it never states the flip side: any future regression that drops a file from the projection set will silently delete the user's existing copy with no backup.

- **[important]** `src/adapters/claude-code/install.ts:1` — **Title overclaims the orphan "PAI tragedy" as solved**
  PR title/description claim "no rename/recase orphans (the PAI tragedy)" and "self-reconciling projection," but the same description admits "the general orphan-*skill* prune is a tracked follow-up." A reader scanning the headline concludes the orphan problem is fixed; orphaned skill dirs still survive.

- **[suggestion]** `src/adapters/pi-dev/skill-projection.ts:11` — **"No per-rename bookkeeping" claim sits beside hardcoded rename list**
  Owned-subtree comments tout "no per-rename bookkeeping," yet this PR ships exactly that for the skill: `LEGACY_PI_DEV_VSA_SKILL_DIRS = ["VSA", "ISA", "isa"]`, a manual list that must grow on every future rename. The generalized claim is narrower than it reads.


---

jcfischer (commented):

5 finding(s): 1 important, 2 suggestion, 2 nit.


### CodeQuality

- **[suggestion]** `src/projection-reconcile.ts:122` — **Temp-hop rename not crash-safe; canonical file can be stranded**
  The sequence `rename(abs, tmp)` → `mkdir` → `rename(tmp, canonicalAbs)` leaves the file at the hidden `.soma-case.*.tmp` name with no canonical file if the second rename throws. The desired file then appears missing until the next reconcile.
  Fix: Wrap in try/catch that renames `tmp` back to `abs` on failure, or document the self-heal-on-next-run recovery.

- **[suggestion]** `src/projection-reconcile.ts:110` — **Case-normalization fixes file names but not directory casing**
  `listFilesRecursive` yields files and only files are renamed; on a case-insensitive FS a recased parent dir (e.g. `Sub/` vs desired `sub/`) keeps its wrong-case name after the file hops, leaving an orphan dir that contradicts the "case-normalized identically on both FS" claim.
  Fix: Normalize directory-segment casing too, or scope the doc claim to file basenames.

- **[nit]** `src/projection-reconcile.ts:105` — **desiredByLower silently collapses case-only duplicate desired paths**
  `desiredByLower.set(rel.toLowerCase(), rel)` lets two desired paths differing only in case overwrite each other with last-write-wins, so the canonical target for a case-variant could be the unintended one.
  Fix: Assert no case-insensitive collisions in `desiredRelPaths`, or document the precondition.

### Maintainability

- **[nit]** `src/adapters/codex/install.ts:67` — **Near-identical ownedSubtrees comment cloned across four adapters**
  The same prose ("Soma-exclusive subtree — reconciled to the projected set each install...") is copy-pasted into claude-code, codex, cursor, and pi-dev specs with small wording drift. Four copies of one explanation drift apart when the mechanism changes.
  Fix: State the mechanism once in the ownedSubtrees JSDoc in install-spec.ts; per-adapter comments only note which dirs are owned vs shared.

### HonestOracle

- **[important]** `test/install.test.ts:302` — **Green suite cited for four subtrees; only codex wired-tested**
  Description claims all four subtrees "(rules/soma, memories/soma, hooks/soma, agent/soma) self-reconcile ... identically on both FS" backed by "Full suite 1465 pass / 0 fail", but the only integration test (`installSomaForCodex`) exercises codex `memories/soma`; claude `hooks/soma`, cursor `.cursor/rules/soma`, and pi-dev `agent/soma` reconcile wiring is unexercised. A wrong subtree path in those three specs would pass the suite yet leave orphans, so the pass count overstates coverage of the per-substrate claim.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/projection-reconcile.ts:145` — **removeEmptyDirs ignores excludeRelPrefixes**
  The file loop honors `excluded.some((prefix) => isUnderOrEqual(rel, prefix))`, but the final `await removeEmptyDirs(root)` walks the whole tree unconditionally; a protected (edit-preserving) installer's dir that is transiently empty would be pruned despite the exclusion. Safe only because such dirs are normally populated.
  Fix: Pass excludeRelPrefixes into removeEmptyDirs and skip excluded subtrees.

- **[suggestion]** `src/projection-reconcile.ts:60` — **isUnderOrEqual hardcodes `/` separator**
  `return path === base || path.startsWith(`${base}/`);` assumes POSIX separators while the inputs come from `resolve`/`relative`, which emit platform separators. Containment checks silently fail on a `\`-separator platform.
  Fix: Use `path.sep` (or compare via `relative()` not starting with `..`).

### Maintainability

- **[suggestion]** `src/projection-reconcile.ts:124` — **Case-rename block deepens reconcileOwnedDir nesting**
  Inside `reconcileOwnedDir` the for→if(canonical)→if(sameFile)→try/catch temp-hop (`const tmp = join(...); await rename(abs, tmp); try { ... } catch (error) { await rename(tmp, abs)... }`) is four levels deep in the loop body. The crash-safe rename is a self-contained unit.
  Fix: Extract `async function caseNormalizeRename(abs, canonicalAbs)` holding the temp-hop + restore-on-failure; the loop body then reads keep / normalize / remove at one level.

### HonestOracle

- **[important]** `src/adapters/claude-code/install.ts:25` — **hooks/soma claimed self-reconciling but no test exercises it**
  claude-code declares `ownedSubtrees: ["rules/soma", "hooks/soma"]` and the description's ✅ lists `hooks/soma` as self-reconciling, backed by "1465 pass / 0 fail". But the four-adapter loop test only wires claude-code as `owned: ".claude/rules/soma"` — `hooks/soma` is exercised by zero tests, so the green suite is offered as proof for a delete-on-reconcile subtree it never touches.

- **[suggestion]** `src/adapters/pi-dev/skill-projection.ts:9` — **Legacy-prune comment asserts prior names without evidence**
  Comment states `["VSA", "ISA", "isa"]` "All were Soma-projected" so safe to prune, but only `isa` (the prior `PI_DEV_VSA_SKILL_ID`) and `VSA` (old legacy code) have evidence of ever being projected; capital `ISA` as a pi-dev projected name is asserted, not shown.


---

jcfischer (commented):

7 finding(s): 2 important, 4 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `test/install.test.ts:338` — **Adapter reconcile loop asserts deletion, not survival**
  The parametrized test only checks `await expect(stat(stale)).rejects.toThrow()` — it never asserts the genuine projected files survive. An over-pruning regression that wiped the whole owned subtree would still pass for every adapter except codex.
  Fix: Also read a known projected file in each owned subtree after reproject and assert it still exists with intact content, mirroring the codex test.

- **[suggestion]** `src/install.ts:230` — **Empty exclusion prefix would make reconcile a silent no-op**
  `.map((dir) => relative(root, dir))` yields `""` when a protected dir equals the owned root; in reconcileOwnedDir `excluded.some((prefix) => isUnderOrEqual(rel, prefix))` treats every file as under `""`, excluding all of them and pruning nothing. No current spec hits this, but the failure is silent rather than guarded.
  Fix: Drop empty-string prefixes (or assert protected dirs are strictly nested below root) before passing excludeRelPrefixes.

- **[nit]** `src/projection-reconcile.ts:70` — **removeEmptyDirs swallows all errors, not just ENOENT**
  Both `catch { return; }` and the trailing `catch {}` discard every error, unlike `statOrEnoent` which deliberately rethrows non-ENOENT. A permission or I/O error during cleanup vanishes with no signal.
  Fix: Narrow the catches to ENOENT and rethrow otherwise, consistent with statOrEnoent.

### Architecture

- **[important]** `src/adapters/codex/install.ts:68` — **Reconcile deletes inside protected/private roots, bypassing path guard**
  `ownedSubtrees: ["memories/soma"]` (and pi-dev's `["agent/soma"]`) name roots CONTEXT.md lists as private roots (`~/.codex/memories/soma/`, `~/.pi/agent/soma/`), and `private`+`protected` means "Cannot be destructively mutated (delete, move, overwrite) without explicit override" enforced by `policy-path-guard.ts`. reconcileOwnedDir `rm`s orphans there with no override and no guard, recoverable only via snapshot — breaching the documented protected-root boundary.
  Fix: Route reconcile deletions through the policy path guard, or confine ownedSubtrees to public/generated-only roots and exclude any path the private/protected-root list covers.

### Maintainability

- **[suggestion]** `test/install.test.ts:318` — **Parametrized reconcile loop duplicates standalone codex test**
  The new `for (const c of [...])` loop includes `{ name: "codex", ... owned: ".codex/memories/soma" }` and prunes a stale file there, repeating the just-added `reproject reconciles an owned subtree` test which does the same codex prune. The standalone keeps extra assertions (shared-dir untouched, content intact), so it's near-duplicate setup.
  Fix: Drop codex from the loop array (it's covered by the richer standalone test), or fold the shared-dir/content assertions into the loop and delete the standalone.

### HonestOracle

- **[important]** `src/install.ts:228` — **Marketed empty-set safety guard has no test**
  The PR sells safety via `if (desiredRel.length === 0) continue; // skip rather than prune the whole subtree` against a projection bug, and cites "1465 pass / 0 fail" as assurance — but no test feeds an empty/incomplete desired set to verify this guard fires. The one mechanism standing between a projection bug and silent mass-deletion is unverified; the green suite is a surrogate for safety it never measured.

- **[suggestion]** `test/projection-reconcile.test.ts:88` — **"Only canonical survives, asserted on both" untrue for case-sensitive branch**
  Description claims the invariant "only canonical survives, content intact" is "asserted on both" filesystems, but the case-sensitive branch only asserts `names not contain "Purpose.md"` and `removed contains "Purpose.md"` — the test never writes `purpose.md`, so canonical-survives is not asserted there. The honesty section overstates what the case-sensitive path verifies.


---

jcfischer (commented):

7 finding(s): 1 important, 3 suggestion, 3 nit.


### CodeQuality

- **[suggestion]** `src/projection-reconcile.ts:75` — **Empty-dir removal uses recursive rm, risking a TOCTOU delete**
  `if ((await readdir(root)).length === 0) await rm(root, { force: true, recursive: true })` confirms emptiness then deletes recursively; if any file appears between the readdir and the rm, `recursive: true` would delete it instead of failing.
  Fix: Use `rmdir(root)` for the confirmed-empty dir so a non-empty state errors instead of silently cascading.

- **[nit]** `src/projection-reconcile.ts:154` — **ReconcileResult asymmetry: renamed=new name, removed=old name**
  `result.renamed.push(canonical)` records the target name while `result.removed.push(rel)` records the source path, so the two arrays mean different things and complicate debugging/log correlation.
  Fix: Record both as the pre-reconcile source rel (or document the asymmetry on ReconcileResult).

- **[nit]** `src/projection-reconcile.ts:80` — **Case-normalize temp path can clobber a colliding file**
  `const tmp = join(dirname(canonicalAbs), ".soma-case.${basename(canonicalAbs)}.tmp")` then `await rename(abs, tmp)` overwrites any pre-existing file at that temp path without checking, e.g. a stale temp from a crashed prior run.
  Fix: Guard against an existing temp target or include a uniquifier in the temp name.

### Architecture

- **[suggestion]** `src/install-spec.ts:73` — **Three overlapping prune mechanisms accreting on install spec**
  `ownedSubtrees` is documented as subsuming `obsoleteHomeFiles` ("Subsumes the former obsoleteHomeFiles for TELOS.md/ACTIVE_ISA.md"), yet both fields persist, and the PR plans a third "projected-skills manifest" for skill pruning. Cleanup responsibility is fragmenting across the spec surface.
  Fix: Track collapsing obsoleteHomeFiles into the owned-subtree model once part 2 lands, so the spec exposes one pruning concept rather than three.

### Maintainability

- **[nit]** `src/adapters/pi-dev/skill-projection.ts:68` — **Redundant existence guard before idempotent force-rm in loop**
  `if (entries.some((entry) => entry.isDirectory() && entry.name === legacy))` gates an `rm(resolve(skillsDir, legacy), { recursive: true, force: true })`, but `force: true` already no-ops on a missing path. The per-iteration `.some()` scan adds branch and cost without changing behavior.
  Fix: Drop the `.some()` guard and call `rm(..., {recursive:true, force:true})` directly for each legacy name.

### HonestOracle

- **[important]** `src/adapters/pi-dev/skill-projection.ts:10` — **PR description misquotes the legacy-prune array**
  Description states `LEGACY_PI_DEV_VSA_SKILL_DIRS = ["VSA","ISA","isa"]` (three entries) but the code is `["VSA", "isa"]` (two) — and the code comment explicitly says pi-dev "never produced a capital ISA". A reviewer trusting the description believes a non-existent entry is pruned; the description contradicts its own diff.

- **[suggestion]** `src/install-spec.ts:82` — **ownedSubtreeExclusions added but unused; cursor credit is misleading**
  Description says "`ownedSubtreeExclusions` spec field excludes ... cursor's VSA skill from reconcile," but no adapter in the diff sets the field, and cursor's VSA skill is excluded by the automatic `vsaSkillProjection.destinationDir` path, not by this field. The field is dead config this PR, sold as solving a case it doesn't touch.


---

jcfischer (commented):

5 finding(s): 1 important, 3 suggestion, 1 nit.


### CodeQuality

- **[important]** `src/adapters/pi-dev/skill-projection.ts:60` — **Legacy-prune of "VSA" can delete canonical "vsa" on case-insensitive FS**
  The comment claims `LEGACY_PI_DEV_VSA_SKILL_DIRS` "never includes the canonical PI_DEV_VSA_SKILL_ID, so each prior name is safe to prune," but on a case-insensitive filesystem (macOS dev) `resolve(skillsDir, "VSA")` and the canonical `vsa` are the same path — so `rm("VSA", {recursive,force})` would delete the freshly projected `vsa` skill if pruning runs after projection. The stated invariant does not hold for case-insensitive hosts.
  Fix: Guard the loop: skip any legacy name whose case-insensitive form equals PI_DEV_VSA_SKILL_ID, and document the required prune-before-project ordering.

### Architecture

- **[suggestion]** `src/projection-reconcile.ts:116` — **Delete-safety invariants live in caller, not the primitive**
  reconcileOwnedDir DELETES but pushes its preconditions onto callers — its docstring says "Callers must pass the COMPLETE projected set... and guard against an empty desired set," while the empty-set guard `if (desiredRel.length === 0) continue;` lives in install.ts. A second future caller that forgets the guard silently wipes a subtree; the safety responsibility is split across two modules.
  Fix: Move the empty-desired-set refusal inside reconcileOwnedDir so the destructive primitive self-enforces its own non-deletion invariant.

### Maintainability

- **[nit]** `src/projection-reconcile.ts:158` — **Stale-removal block duplicated across two branches**
  `await rm(abs, { force: true }); result.removed.push(rel);` appears verbatim in the case-sensitive-distinct branch and the final otherwise branch. Two copies of the same remove-and-record action drift independently.
  Fix: Hoist to a local `removeStale(abs, rel)` helper, or fall through to a single trailing remove when `canonical === undefined || !sameFile`.

### HonestOracle

- **[suggestion]** `src/adapters/claude-code/install.ts:23` — **'Subsumes obsoleteHomeFiles' hides a weakened guarantee**
  New comment claims reconcile 'Subsumes the former obsoleteHomeFiles for TELOS.md/ACTIVE_ISA.md'. The deleted code's own justification was unconditional: 'an upgrade must delete the stale copies or it would keep loading frozen old content every session.' Reconcile is gated (`if (desiredRel.length === 0) continue;`), so on an empty desired set the frozen TELOS.md survives and is auto-loaded — the exact harm the old code prevented. 'Subsumes' overstates: it is not a strict superset.

- **[suggestion]** `test/install.test.ts:311` — **'Shared dir untouched' test asserts a path reconcile never targets**
  The codex test writes `.codex/hooks/user-custom.mjs` and asserts it survives as proof 'shared dir untouched', but `hooks/` is not in codex's `ownedSubtrees` (only `memories/soma`), so reconcile never targets it — the assertion passes trivially. It markets coverage of shared-dir safety while testing a path reconcile cannot reach; nothing guards or tests against a shared dir being mistakenly listed in `ownedSubtrees`, which the PR's only safeguard ('Do NOT list shared dirs') leaves to human discipline.


---

jcfischer (commented):

### CodeQuality

- **[nit]** `src/projection-reconcile.ts:122` — **Dead mkdir inside crash-safe rename block**
  `caseNormalizeRename` only runs on a case-insensitive FS where `abs` and `canonicalAbs` differ by basename case only (per the JSDoc, dir segments aren't normalized), so `dirname(canonicalAbs)` always already exists; `await mkdir(dirname(canonicalAbs), { recursive: true })` is a guaranteed no-op that adds a throw site to the critical section.
  Fix: Drop the mkdir call — the destination directory always exists when this branch is reached.

### Architecture

- **[suggestion]** `src/install.ts:230` — **`protectedDirs` reuses locked glossary adjective off-meaning**
  `const protectedDirs = [...]` names the reconcile-excluded VSA skill dir, but CONTEXT.md locks `protected` to "Cannot be destructively mutated... without explicit override" (private/protected/generated data adjectives). Here it means "owned by another installer, exclude from reconcile" — drift that erodes the locked term.
  Fix: Rename to `excludedDirs` / `externallyOwnedDirs` to match the actual concept and free `protected` for its glossary sense.

### HonestOracle

- **[important]** `test/install.test.ts:360` — **Case-insensitive-FS wipe test is vacuous on CI**
  Test `"reinstall does not wipe a user file in pi-dev's canonical vsa skill (case-insensitive FS)"` guards the same-inode collision where prune of "VSA" hits "vsa". On case-sensitive Linux CI readdir yields "vsa" which never matches the legacy list, so the dangerous path cannot occur — the green CI run proves nothing about the guarantee the test name asserts; only dev macOS exercises it.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### Architecture

- **[suggestion]** `src/projection-reconcile.ts:33` — **Primitive still uses locked term `protected` off-meaning**
  `removeEmptyDirs(root, protectedAbs)` names exclude-from-prune dirs `protectedAbs`, but CONTEXT.md locks `protected` to mean "Cannot be destructively mutated (delete, move, overwrite) without explicit override." install.ts was deliberately corrected to `excludedDirs` ("Named 'excluded', not 'protected', to avoid the locked protected-root term") yet the canonical primitive keeps the off-meaning term.
  Fix: Rename `protectedAbs` → `excludedAbs` for consistency with the corrected caller and the glossary.

### Maintainability

- **[suggestion]** `src/projection-reconcile.ts:5` — **Two recursive dir walkers duplicate readdir/ENOENT skeleton**
  `listFilesRecursive`'s `walk` and `removeEmptyDirs` both implement the identical `try { readdir(dir,{withFileTypes:true}) } catch { if(isEnoent) return; throw }` + `for entry of entries / if isDirectory recurse` skeleton. Future change to walk semantics must touch two places.
  Fix: Extract a `walkDirs(root, onFile?, onDirExit?)` helper both callers share, or have removeEmptyDirs reuse the same traversal primitive.


---

PR #352, opened 2026-06-25
feat(install): prune renamed-away 'ISA' skill orphan (case-clean — part 2)

jcfischer:
**Part 2** of the case-clean work (part 1 = #351, merged). Removes the orphaned `ISA` skill dirs that soma#329's ISA→VSA rename left behind — the leftover #351 explicitly deferred.

## Orphans this PR removes (verified here)
- **Source root** `~/.soma/skills/ISA` — `loadSomaSkills` enumerates every dir under the soma home skills dir and projects it, so this stale ISA re-propagated to every substrate each install. The propagation root. (pruned before `loadSomaHome`.)
- **Projected copies** in the shared `skills/` roots of **codex** and **claude-code** (via `vsaSkillProjection.prepare`).

Handled by prior work, NOT re-verified in this PR: cursor's `ISA` (under the owned `.cursor/rules/soma` → removed by part-1 reconcile, #351) and pi-dev (`removeLegacyPiDevVsaSkillProjection`, #351). After this lands a full reproject is expected to clean every home, but only codex/claude/source are wired+tested here.

## Safety — the hard part
`~/.soma/skills` and the substrate `skills/` roots also hold the principal's own skills (dozens). `pruneLegacyVsaSkill(skillsDir)` removes the ISA dir ONLY behind a two-signal provenance gate:
1. exact on-disk dir name `ISA` (load-bearing on case-insensitive FS — a user dir cased `isa` is never matched; now has a dedicated test), AND
2. `SKILL.md` frontmatter `name: ISA` **AND** description contains the VSA-skill identity sentence `"Owns the Ideal State Artifact"`.

A user skill named `ISA` lacking the marker (or with a different `name`) is preserved. `isEnoent`-safe; any other fs error re-throws.

## Tests (10)
helper unit: removes soma ISA; preserves a user ISA lacking the marker; preserves an ISA with a different name; **preserves a lowercase `isa`** (case-insensitive guard); no-ops (absent ISA / absent skills dir / no SKILL.md). Integration: codex & claude install prune sibling ISA while projected VSA **and** a planted user skill survive a reproject; source-home prune (ISA pruned, user skill + canonical VSA survive). typecheck + eslint clean; full suite **1481 pass / 0 fail**.

### DIFF
--- a/src/adapters/claude-code/install.ts
+++ b/src/adapters/claude-code/install.ts
@@ -29,7 +29,8 @@ export const claudeCodeInstallSpec: SubstrateInstallSpec<"claude-code"> = {
   vsaSkillProjection: {
     destinationDir: vsaSkillUnder(),
     // soma#329: before reprojecting VSA, prune a sibling renamed-away "ISA" skill
-    // from <home>/skills (provenance-gated — never a user skill).
+    // from <home>/skills (provenance-gated to Soma's published ISA identity — a
+    // user skill lacking that identity is preserved; see pruneLegacyVsaSkill doc).
     prepare: vsaSiblingPrunePrepare(),
   },
   postProjection: [
--- a/src/adapters/claude-code/install.ts
+++ b/src/adapters/claude-code/install.ts
@@ -1,4 +1,6 @@
+import { resolve } from "node:path";
 import { vsaSkillUnder, type SubstrateInstallSpec } from "../../install-spec";
+import { pruneLegacyVsaSkill } from "../../legacy-skill-prune";
 import { CLAUDE_CODE_RULES_FILES } from "../claude-code";
 import {
   SOMA_CLAUDE_HOOK_CONFIG_RELATIVE_PATH,
@@ -27,6 +29,10 @@ export const claudeCodeInstallSpec: SubstrateInstallSpec<"claude-code"> = {
     : [],
   vsaSkillProjection: {
     destinationDir: vsaSkillUnder(),
+    // soma#329: before reprojecting VSA, prune a sibling "ISA" skill left over
+    // from the rename. `vsaSkillUnder()` puts VSA at <home>/skills/VSA, so the
+    // shared skills dir is <home>/skills. Provenance-gated — never a user skill.
+    prepare: (substrateHome) => pruneLegacyVsaSkill(resolve(substrateHome, "skills")).then(() => undefined),
   },
   postProjection: [
     {
--- a/src/adapters/claude-code/install.ts
+++ b/src/adapters/claude-code/install.ts
@@ -1,6 +1,5 @@
-import { resolve } from "node:path";
 import { vsaSkillUnder, type SubstrateInstallSpec } from "../../install-spec";
-import { pruneLegacyVsaSkill } from "../../legacy-skill-prune";
+import { vsaSiblingPrunePrepare } from "../../legacy-skill-prune";
 import { CLAUDE_CODE_RULES_FILES } from "../claude-code";
 import {
   SOMA_CLAUDE_HOOK_CONFIG_RELATIVE_PATH,
@@ -29,10 +28,9 @@ export const claudeCodeInstallSpec: SubstrateInstallSpec<"claude-code"> = {
     : [],
   vsaSkillProjection: {
     destinationDir: vsaSkillUnder(),
-    // soma#329: before reprojecting VSA, prune a sibling "ISA" skill left over
-    // from the rename. `vsaSkillUnder()` puts VSA at <home>/skills/VSA, so the
-    // shared skills dir is <home>/skills. Provenance-gated — never a user skill.
-    prepare: (substrateHome) => pruneLegacyVsaSkill(resolve(substrateHome, "skills")).then(() => undefined),
+    // soma#329: before reprojecting VSA, prune a sibling renamed-away "ISA" skill
+    // from <home>/skills (provenance-gated — never a user skill).
+    prepare: vsaSiblingPrunePrepare(),
   },
   postProjection: [
     {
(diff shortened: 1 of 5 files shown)


jcfischer (commented):

5 finding(s): 1 important, 3 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/legacy-skill-prune.ts:24` — **Unescaped key interpolated into dynamic RegExp**
  `new RegExp(\`^${key}:\\s*(.+?)\\s*$\`, "m")` injects `key` unescaped. Current callers pass literal "name"/"description" so it's safe today, but a key with regex metacharacters would silently misbehave.
  Fix: Since keys are a fixed closed set, leave as-is or escape: `key.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")`.

- **[nit]** `src/legacy-skill-prune.ts:19` — **Single-line-only description match misses folded YAML**
  `frontmatterField` captures `(.+?)` to end of one line, so a multi-line/folded YAML `description` would not surface the identity marker. Harmless (fails safe to preserve), and real ISA descriptions are single-line.
  Fix: No change needed unless descriptions become multi-line; note the assumption in the comment.

### Maintainability

- **[suggestion]** `src/adapters/codex/install.ts:72` — **Identical prepare wiring duplicated across two adapters**
  Both codex and claude-code specs copy the same 3-line comment and `prepare: (substrateHome) => pruneLegacyVsaSkill(resolve(substrateHome, "skills")).then(() => undefined)` verbatim. A future change to the prune wiring must be made in two places.
  Fix: Extract a shared `vsaSiblingPrunePrepare` helper (or factory) and reference it from both specs.

### HonestOracle

- **[important]** `src/legacy-skill-prune.ts:7` — **Load-bearing case-insensitive safety claim is untested**
  The comment asserts "the exact-name match is load-bearing on case-insensitive filesystems, where a blind rm of 'ISA' could resolve to the SAME inode as a user dir cased 'isa'" and the description repeats "a user dir cased `isa` is never matched" — but no test in the 9-test suite plants a lowercase `isa` dir. The headline safety property is offered as proven yet has zero coverage; on the author's macOS (case-insensitive) this is exactly the unverified path.

- **[suggestion]** `src/legacy-skill-prune.ts:50` — **"Every home clean" conclusion exceeds what this diff verifies**
  Description claims "a reproject of all substrates leaves every home clean," but only codex, claude-code, and the source home are wired/tested here; cursor and pi-dev are asserted handled "by part-1 reconcile" and a separate function with no verification in this PR. The all-substrates conclusion rests on an unre-checked cross-PR claim.


---

jcfischer (commented):

5 finding(s): 1 important, 2 suggestion, 2 nit.


### CodeQuality

- **[nit]** `src/legacy-skill-prune.ts:25` — **Asymmetric quote-strip accepts mismatched YAML quotes**
  `.replace(/^["']|["']$/g, "")` strips a leading and trailing quote independently, so a malformed value like `'ISA"` is normalized to `ISA` and matches the name gate. Harmless in practice but technically loosens the provenance check.
  Fix: Match a balanced pair only, e.g. `/^"(.*)"$|^'(.*)'$/`, or rely on a YAML parser.

### Architecture

- **[suggestion]** `src/install.ts:135` — **One-time migration prune permanently on install path**
  `await pruneLegacyVsaSkill(createPaths(somaHome.somaHome).skills())` runs on every install forever with no version guard or removal path, so a #329-specific cleanup becomes standing core-install behavior. Future renames would accrete sibling prune calls here.
  Fix: Gate behind a one-shot migration marker or note an expiry/removal condition so the shim doesn't live indefinitely on the hot path.

### Maintainability

- **[nit]** `test/legacy-skill-prune.test.ts:60` — **Repeated inline SKILL.md frontmatter builder across tests**
  The `["---", "name: X", 'description: "..."', "---", ...].join("\n")` idiom is hand-rolled ~6 times (user ISA, differing name, user skill, user-source-skill, etc.). A `makeSkillMd({name, description})` helper alongside the existing `plantSkill` would cut the repetition and make the provenance-signal under test obvious per case.
  Fix: Add `makeSkillMd({name, description})` and build the per-test markdown from it.

### HonestOracle

- **[important]** `src/adapters/claude-code/install.ts:31` — **'Never a user skill' overclaims what the content gate proves**
  Comment asserts "provenance-gated — never a user skill," and the PR repeats the gate keys on `name: ISA` + description containing "Owns the Ideal State Artifact." Both signals are soma's OWN published strings, so a user who copied/forked the ISA skill (retaining frontmatter) is indistinguishable from the stale orphan and gets silently `rm`'d — the gate verifies 'lacks the marker is safe', not 'never a user skill'.

- **[suggestion]** `test/legacy-skill-prune.test.ts:73` — **Case-insensitive guard test only bites on a case-insensitive FS**
  Test plants `isa` and asserts preservation as "the load-bearing case-insensitive guard," but on case-sensitive CI (Linux ext4) `isa` and `ISA` are simply distinct dirs, so the assertion passes without exercising the collision the comment describes. The claim that this validates the guard holds only on the author's APFS host, not necessarily where CI runs.


---

jcfischer (commented):

2 finding(s): 1 suggestion, 1 nit.


### CodeQuality

- **[nit]** `src/legacy-skill-prune.ts:16` — **Self-contradictory comment: 'begins (anywhere)'**
  Comment says description "begins (anywhere) with this identity sentence" — 'begins' implies start, '(anywhere)' implies not. The implementation uses `description?.includes(...)`, i.e. anywhere, so the word 'begins' misleads a reader about the actual match semantics.
  Fix: Reword to "description contains this identity sentence (anywhere)".

### HonestOracle

- **[suggestion]** `src/legacy-skill-prune.ts:89` — **"Safety" framing omits irreversible no-backup delete**
  The PR's Safety section frames the hard part purely as provenance ("is this dir soma's or the user's"), but the action is `rm(..., { recursive: true, force: true })` with no backup — if the gate ever mis-fires (e.g. a user fork, which the code doc concedes is indistinguishable), user data inside the dir is destroyed irrecoverably, a consequence the description never names.


---

Issue #354, opened 2026-06-25
Official skill collection + shared projection primitive


jcfischer:
Per **ADR 0002** (`docs/adr/0002-official-skill-collection-and-projection-primitive.md`).

## Problem

Primitive-operating skills (Purpose, VSA, the-algorithm) are bespoke code-gen importers emitting markdown as TS string literals — `renderSkill()` in `src/algorithm-importer.ts:37` / `:246`, and `src/vsa-skill-installer.ts` (20.5 KB). Every new official skill = another 15–20 KB renderer. Two compounding gaps:

1. **No invocable-skill projection.** `soma install` projects only the catalog (`rules/soma/SKILLS.md` via `renderSkills`, scanned by `loadSomaSkills` `src/soma-home.ts:154`). It never materialises invocable skill dirs into the substrate loader (`~/.claude/skills/`). The new `Purpose` skill had to be hand-symlinked to become usable.
2. **Tier confusion.** ~104 entries in `~/.soma/skills/` mix a few official skills with ~100 migrated third-party skills (art, research…) core has no reason to own.

## Decision (3 moves)

1. **Extract a projection primitive** — `soma project-skill <skill-dir> --substrate <id[,id…]>` (+ `unproject-skill`, + a library export). Pulls the skill-dir → loader step (symlink into loader + refresh catalog) out of `install`. Idempotent; must reconcile a prior hand-materialised copy (e.g. existing `~/.claude/skills/{VSA,Interview}`) without clobbering unrelated user skills.
2. **Official skills → plain `.md` in repo `skills/`** — retire `vsa-skill-installer.ts` and `algorithm-importer.ts`'s `renderSkill()` path; ship `skills/{purpose,vsa,the-algorithm}/` as files, versioned with core (schema-coupled). The `Purpose` skill is the prototype. Add `soma install <substrate> --skills purpose,memory` to project only chosen skills. Default install stays minimal.
3. **Arc delegates** — arc calls this primitive on install/uninstall (cross-repo, see arc issue below). Soma stays the single projection truth; arc never reimplements adapter logic.

## Open decision (settle at impl time)

Default fan-out of `project-skill`: one substrate vs. all installed. Contract is substrate-list-parameterised either way → `ready-for-human` on this point.

## Out of scope (separate cleanup)

Evicting the ~100 migrated third-party skills squatting in `~/.soma/skills/`.

## Acceptance

[...]

PR #355, opened 2026-06-25
feat(skills): soma project-skill / unproject-skill projection primitive (#354 slice 1)

jcfischer:
Slice 1 of #354 (per ADR 0002). Adds a standalone, idempotent, multi-substrate projection primitive (CLI + lib): `soma project-skill` / `unproject-skill`.

**Scope of THIS PR:** the primitive itself + its CLI wiring (`src/skill-projection.ts`, `src/cli/skill-projection-cli.ts`, `cli.ts`) and the `loadSomaSkills` symlink-scanner fix. It does **not** yet wire `soma install --skills` or arc — those are later #354 / arc#251 slices that will *call* this primitive. The ADR's "single projection truth" is the design intent the primitive enables, not a claim that existing install paths route through it today.

## What lands
- `projectSkill` / `unprojectSkill` / `planProjectSkill` / `planUnprojectSkill`. Links (not copies) source → `~/.soma/skills/<Name>` (registry the catalog scans) + each substrate loader, then refreshes only the `SKILLS.md` catalog file (content-match, no per-substrate path hardcoding). Identity = frontmatter `name`.
- Both verbs default to **dry-run**; `--apply` to execute. `--force` to replace a real (non-symlink) dir occupying a slot (otherwise refused — guards user skills from clobber).
- `src/soma-home.ts`: `loadSomaSkills` admits a symlink-to-dir (latent bug — symlinked skills were silently skipped).

## Review history
Self-review + sage review (changes-requested) addressed:
- **blocker** unproject ignored `--dry-run` → now gated on `--apply` (+ `planUnprojectSkill`).
- **major** symlink-dir arg rejected → `stat` (follows links) for name resolution.
- **major** provenance: refuse to clobber a non-symlink dir without `--force`.
- **major** PR-description delegation claim → corrected (this paragraph).
- **suggestions** dedup catalog-refresh loop; split parser apply state.
- **deferred** → #356: move loader projection behind an adapter/install-spec method (boundary cleanup).

## Verification
typecheck PASS · lint clean · 11 unit tests · full suite **1492/0** · e2e: dry-run/apply/force/unproject all confirmed.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -25,6 +25,16 @@ import {
   type ParsedUpgradeArgs,
 } from "./cli/substrate-lifecycle";
 import {
+  PROJECT_SKILL_COMMAND_HELP,
+  UNPROJECT_SKILL_COMMAND_HELP,
+  parseProjectSkillArgs,
+  parseUnprojectSkillArgs,
+  runProjectSkillCli,
+  runUnprojectSkillCli,
+  type ParsedProjectSkillArgs,
+  type ParsedUnprojectSkillArgs,
+} from "./cli/skill-projection-cli";
+import {
   MIGRATE_COMMAND_HELP,
   parseMigrateArgs,
   runMigrateCli,
@@ -131,6 +141,8 @@ type ParsedArgs =
   | ParsedUpgradeArgs
   | ParsedExportArgs
   | ParsedDaemonArgs
+  | ParsedProjectSkillArgs
+  | ParsedUnprojectSkillArgs
   | ParsedOnboardingArgs
   | ParsedImportArgs
   | ParsedMigrateArgs
@@ -167,8 +179,10 @@ const TOP_LEVEL_COMMANDS = [
   "migrate",
   "opinion",
   "policy",
+  "project-skill",
   "relationship",
   "reproject",
+  "unproject-skill",
   "result",
   "rollback",
   "session",
@@ -195,6 +209,8 @@ const COMMAND_HELP: Record<string, { usage: string; subcommands?: Record<string,
   install: SUBSTRATE_LIFECYCLE_COMMAND_HELP.install,
   uninstall: SUBSTRATE_LIFECYCLE_COMMAND_HELP.uninstall,
   reproject: SUBSTRATE_LIFECYCLE_COMMAND_HELP.reproject,
+  "project-skill": PROJECT_SKILL_COMMAND_HELP,
+  "unproject-skill": UNPROJECT_SKILL_COMMAND_HELP,
   upgrade: SUBSTRATE_LIFECYCLE_COMMAND_HELP.upgrade,
   export: SUBSTRATE_LIFECYCLE_COMMAND_HELP.export,
   daemon: SUBSTRATE_LIFECYCLE_COMMAND_HELP.daemon,
@@ -291,6 +307,14 @@ function parseArgs(args: string[]): ParsedArgs {
 
   if (args[0] === "reproject") {
     return parseReprojectArgs(args);
+  }
+
+  if (args[0] === "project-skill") {
+    return parseProjectSkillArgs(args);
+  }
+
+  if (args[0] === "unproject-skill") {
+    return parseUnprojectSkillArgs(args);
   }
 
   if (args[0] === "upgrade") {
@@ -514,6 +538,14 @@ export async function runSomaCli(args: string[]): Promise<string> {
 
   if (parsed.command === "snapshot" || parsed.command === "history" || parsed.command === "rollback") {
     return runSnapshotCli(parsed);
+  }
+
+  if (parsed.command === "project-skill") {
+    return runProjectSkillCli(parsed);
+  }
+
+  if (parsed.command === "unproject-skill") {
+    return runUnprojectSkillCli(parsed);
   }
 
   if (isSubstrateLifecycleArgs(parsed)) {
--- a/src/cli/skill-projection-cli.ts
+++ b/src/cli/skill-projection-cli.ts
@@ -14,18 +14,18 @@ export interface ParsedProjectSkillArgs {
   skillDir: string;
   substrates: InstallSubstrate[];
   apply: boolean;
-  options: { homeDir?: string; somaHome?: string; substrateHome?: string };
+  options: { homeDir?: string; somaHome?: string; substrateHome?: string; force?: boolean };
 }
 
 export interface ParsedUnprojectSkillArgs {
   command: "unproject-skill";
   skill: string;
   substrates: InstallSubstrate[];
-  options: { homeDir?: string; somaHome?: string; substrateHome?: string };
+  options: { homeDir?: string; somaHome?: string; substrateHome?: string; force?: boolean };
 }
 
 const SUBSTRATE_LIST = INSTALL_SUBSTRATES.join("|");
-const SKILL_OPTIONS = "[--substrate <id[,id…]>] [--dry-run] [--apply] [--home-dir <dir>] [--soma-home <dir>] [--substrate-home <dir>]";
+const SKILL_OPTIONS = "[--substrate <id[,id…]>] [--dry-run] [--apply] [--force] [--home-dir <dir>] [--soma-home <dir>] [--substrate-home <dir>]";
 
 export const PROJECT_SKILL_COMMAND_HELP = {
   usage: `Usage: soma project-skill <skill-dir> ${SKILL_OPTIONS}  (--substrate defaults to claude-code; ${SUBSTRATE_LIST})`,
@@ -50,7 +50,7 @@ interface SkillCommonParse {
   positional: string;
   substrates: InstallSubstrate[];
   apply: boolean;
-  options: { homeDir?: string; somaHome?: string; substrateHome?: string };
+  options: { homeDir?: string; somaHome?: string; substrateHome?: string; force?: boolean };
 }
 
 function parseSkillArgs(verb: string, usage: string, args: string[]): SkillCommonParse {
@@ -59,7 +59,7 @@ function parseSkillArgs(verb: string, usage: string, args: string[]): SkillCommo
     throw new Error(usage);
   }
 
-  const options: { homeDir?: string; somaHome?: string; substrateHome?: string } = {};
+  const options: { homeDir?: string; somaHome?: string; substrateHome?: string; force?: boolean } = {};
   let substrates: InstallSubstrate[] = ["claude-code"];
   let apply = false;
 
@@ -81,6 +81,9 @@ function parseSkillArgs(verb: string, usage: string, args: string[]): SkillCommo
       case "--substrate-home":
         options.substrateHome = readOption(rest, index, arg);
         index += 1;
+        continue;
+      case "--force":
+        options.force = true;
         continue;
       case "--apply":
         apply = true;
--- a/src/cli/skill-projection-cli.ts
+++ b/src/cli/skill-projection-cli.ts
@@ -1,0 +1,167 @@
+import {
+  planProjectSkill,
+  projectSkill,
+  unprojectSkill,
+  type ProjectSkillOptions,
+  type SkillProjectionResult,
+  type UnprojectSkillOptions,
+} from "../skill-projection";
+import { INSTALL_SUBSTRATES, isInstallSubstrate, type InstallSubstrate } from "./substrate-lifecycle";
+import { readOption } from "./parse-utils";
+
+export interface ParsedProjectSkillArgs {
+  command: "project-skill";
+  skillDir: string;
+  substrates: InstallSubstrate[];
+  apply: boolean;
+  options: { homeDir?: string; somaHome?: string; substrateHome?: string };
+}
+
+export interface ParsedUnprojectSkillArgs {
+  command: "unproject-skill";
+  skill: string;
+  substrates: InstallSubstrate[];
+  options: { homeDir?: string; somaHome?: string; substrateHome?: string };
+}
+
+const SUBSTRATE_LIST = INSTALL_SUBSTRATES.join("|");
+const SKILL_OPTIONS = "[--substrate <id[,id…]>] [--dry-run] [--apply] [--home-dir <dir>] [--soma-home <dir>] [--substrate-home <dir>]";
+
+export const PROJECT_SKILL_COMMAND_HELP = {
+  usage: `Usage: soma project-skill <skill-dir> ${SKILL_OPTIONS}  (--substrate defaults to claude-code; ${SUBSTRATE_LIST})`,
+};
+
+export const UNPROJECT_SKILL_COMMAND_HELP = {
+  usage: `Usage: soma unproject-skill <skill-dir|name> ${SKILL_OPTIONS}  (--substrate defaults to claude-code; ${SUBSTRATE_LIST})`,
+};
+
+function parseSubstrateCsv(value: string): InstallSubstrate[] {
+  const ids = value.split(",").map((part) => part.trim()).filter((part) => part.length > 0);
+  if (ids.length === 0) throw new Error("--substrate requires at least one substrate id.");
+  for (const id of ids) {
+    if (!isInstallSubstrate(id)) {
+      throw new Error(`--substrate must be one of ${INSTALL_SUBSTRATES.join(", ")} (got "${id}").`);
+    }
+  }
+  return ids as InstallSubstrate[];
+}
+
+interface SkillCommonParse {
+  positional: string;
+  substrates: InstallSubstrate[];
+  apply: boolean;
+  options: { homeDir?: string; somaHome?: string; substrateHome?: string };
+}
+
+function parseSkillArgs(verb: string, usage: string, args: string[]): SkillCommonParse {
+  const [command, positional, ...rest] = args;
+  if (command !== verb || !positional || positional.startsWith("--")) {
+    throw new Error(usage);
+  }
+
+  const options: { homeDir?: string; somaHome?: string; substrateHome?: string } = {};
+  let substrates: InstallSubstrate[] = ["claude-code"];
+  let apply = false;
+
+  for (let index = 0; index < rest.length; index += 1) {
+    const arg = rest[index];
+    switch (arg) {
+      case "--substrate":
+        substrates = parseSubstrateCsv(readOption(rest, index, arg));
+        index += 1;
+        continue;
+      case "--home-dir":
+        options.homeDir = readOption(rest, index, arg);
+        index += 1;
+        continue;
+      case "--soma-home":
+        options.somaHome = readOption(rest, index, arg);
+        index += 1;
+        continue;
+      case "--substrate-home":
+        options.substrateHome = readOption(rest, index, arg);
+        index += 1;
+        continue;
+      case "--apply":
+        apply = true;
+        continue;
+      case "--dry-run":
+        apply = false;
+        continue;
+      default:
+        throw new Error(`Unknown option: ${arg}`);
+    }
+  }
+
+  return { positional, substrates, apply, options };
+}
+
+export function parseProjectSkillArgs(args: string[]): ParsedProjectSkillArgs {
+  const { positional, substrates, apply, options } = parseSkillArgs(
+    "project-skill",
+    PROJECT_SKILL_COMMAND_HELP.usage,
+    args,
+  );
+  return { command: "project-skill", skillDir: positional, substrates, apply, options };
+}
+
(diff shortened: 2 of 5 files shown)


jcfischer (commented):
## FullReview — CodeQuality · Security · Architecture (self-review; GH blocks request-changes on own PR)

Claude-code default path is solid — tested, e2e smoke verified, idempotent, identity from frontmatter `name`, traversal-guarded names, `--substrate-home` multi-guard present. Scanner fix (`isSymbolicLink` admission) is minimal/additive/correct. Findings are about multi-substrate generalization + destructive-write safety; cheap to harden before arc drives this.

### MAJOR — destructive removal of a possibly user-owned skill dir, no provenance gate
`ensureSymlink` `rm(linkPath,{recursive,force})` on a non-matching slot, and `unprojectSkill`→`removeLink` recurse-removes the loader path. Slot `~/.claude/skills/<Name>` also holds the principal's own skills. Exact-name collision → silent delete, no backup. The VSA installer gates this via provenance (`vsaSiblingPrunePrepare`); this primitive doesn't. Highest risk on the arc path (arc#251) — arc projects third-party packs into a loader full of user skills.
**Recommend:** don't clobber a real (non-symlink) dir lacking Soma provenance unless `--force`; replacing a symlink we own stays as-is.

### MEDIUM — multi-substrate creates dead symlinks where there's no invocable loader
Only claude-code (≈codex) load skill dirs. Cursor only *lists* skills for "manual inspection" (`cursor.ts:121`); pi-dev uses a `skillNameOverride` id scheme a plain `<Name>` dir won't match. `--substrate cursor,pi-dev` → dead links (catalog entry still works). Default claude-code makes this opt-in, but the CLI accepts no-op substrates silently.
**Recommend:** gate to substrates declaring an invocable loader, or document. Ties to the ADR's open fan-out default — settle here.

### NIT — `findCatalogFile` silent `continue` on no match
All 5 substrates emit a file byte-equal to `renderSkills(input)` today, so it won't fire — but a future format change drops the catalog refresh silently. Prefer throw/warn over silent skip.

### NIT — `unprojectSkill` bare-name resolution hijackable by CWD
`lstat(resolve(skill))` treats a bare name as a path if a same-named dir exists in CWD. Only path-probe when the arg looks path-like.

### Positives
Registry-symlink-as-catalog-source is clean; content-match catalog discovery avoids per-substrate path hardcoding; test coverage genuinely good (idempotency, copy-replace, authored-dir preservation, dry-run, multi-guard).

verdict: blockers=0 majors=1 mediums=1 nits=2 — recommend changes before arc integration; claude-code path itself is mergeable.

jcfischer (commented):

6 finding(s): 1 blocker, 3 important, 2 suggestion.


### CodeQuality

- **[blocker]** `src/cli/skill-projection-cli.ts:161` — **Unproject dry-run still writes**
  The parser accepts dry-run via "case \"--dry-run\": apply = false;", but unproject drops that flag and always executes "return formatProjectionResult(\"Soma unproject-skill applied\", await unprojectSkill(options));". A user running `soma unproject-skill ... --dry-run` will still remove links and rewrite catalogs.
  Fix: Carry `apply` on `ParsedUnprojectSkillArgs` and skip `unprojectSkill` unless `--apply` is set.

- **[important]** `src/skill-projection.ts:291` — **Symlinked skill dirs are rejected**
  Path resolution uses "const stat = await lstat(candidate);" followed by "if (stat.isDirectory()) {", so a symlink-to-dir is not treated as a skill directory. Since projection creates registry skill symlinks, `unproject-skill ~/.soma/skills/<Name>` falls through to `assertSafeSkillName` with slashes and fails.
  Fix: Use `stat(candidate)` for this check, or explicitly follow symlinked directories before calling `readSkillName`.

### Architecture

- **[important]** `src/skill-projection.ts:29` — **Projection primitive bypasses adapters**
  docs/architecture.md says Skills core “only defines discovery and routing contracts” and “A substrate adapter decides how to load and execute them,” but this module hard-codes all substrate projection builders in `const projectionBuilders: Record<InstallSubstrate, ...> = { codex: buildCodexHomeProjection, ... }`. It also derives substrate loader roots with `installSpecFor(substrate).vsaSkillProjection.destinationDir(substrateHome)`, so adding or changing a substrate skill projection now requires changing this cross-substrate module instead of the owning adapter/spec boundary.
  Fix: Move skill-loader projection and catalog refresh behind an adapter/install-spec method, and let this CLI call that substrate-owned surface.

### Maintainability

- **[suggestion]** `src/skill-projection.ts:229` — **Extract duplicated catalog refresh**
  The catalog refresh loop is duplicated in `projectSkill` and `unprojectSkill`, including `const catalog = findCatalogFile(substrate, input, projectionOptions);` and `process.stderr.write(`Warning: no SKILLS catalog file found for substrate ${substrate}; catalog not refreshed.\n`);`. Catalog-format changes now have two blocks to keep in sync.
  Fix: Extract a shared `refreshSkillCatalogs(input, substrates, options)` helper.

- **[suggestion]** `src/cli/skill-projection-cli.ts:91` — **Split project-only parser state**
  The shared parser handles project-only apply state with `case "--dry-run": apply = false;`, but `parseUnprojectSkillArgs` drops it via `const { positional, substrates, options } = parseSkillArgs(`. That makes the common helper carry command-specific behavior for one callsite.
  Fix: Keep common path/substrate/home parsing separate from project-only `--apply|--dry-run` handling.

### HonestOracle

- **[important]** `PR description:2` — **Delegation claim is unsupported**
  The description claims the primitive is one "that `soma install --skills` and arc both delegate to," but the diff only wires `project-skill` / `unproject-skill` into `src/cli.ts` and adds `src/skill-projection.ts`; no `soma install --skills` or arc callsite is changed. Downstream readers could believe existing install paths now use Soma as the single projection truth when this PR does not show that.


---

jcfischer (commented):

5 finding(s): 2 blocker, 2 important, 1 suggestion.


### CodeQuality

- **[blocker]** `src/cli/skill-projection-cli.ts:158` — **unproject ignores dry-run**
  The CLI advertises `const SKILL_OPTIONS = "[--substrate <id[,id…]>] [--dry-run] [--apply] [--force] [--home-dir <dir>] [--soma-home <dir>] [--substrate-home <dir>]";`, but `runUnprojectSkillCli` always executes `return formatProjectionResult("Soma unproject-skill applied", await unprojectSkill(options));`. A user running `soma unproject-skill X --dry-run` will still remove links and refresh catalogs.
  Fix: Carry `apply` through unproject parsing and make dry-run plan-only/no-op, or remove those flags for unproject.

- **[important]** `src/skill-projection.ts:225` — **failed project leaves registry link**
  `projectSkill` writes the registry first with `const status = await ensureSymlink(registryDest, skillDir, options.force ?? false);`, then later writes loader links with `const status = await ensureSymlink(dest, skillDir, options.force ?? false);`. If a loader slot throws on a non-symlink, the failed command has already registered the skill.
  Fix: Preflight all target slots before writing, or rollback earlier links when a later `ensureSymlink` fails.

### Architecture

- **[important]** `src/skill-projection.ts:113` — **Core owns substrate loader layout**
  docs/architecture.md says Skills core defines discovery/routing while "A substrate adapter decides how to load and execute them," but the new core helper hard-codes loader derivation for every substrate via `return dirname(installSpecFor(substrate).vsaSkillProjection.destinationDir(substrateHome));`. This also conflicts with CONTEXT.md's skill table for Codex, which says `Codex instruction (Codex has no native skill primitive)`, while the diff creates substrate skill-directory links through `const dest = join(substrateSkillsRoot(substrate, substrateHome), name);`.
  Fix: Move per-substrate skill projection targets behind adapter/install-spec methods that can return instruction/catalog-only behavior for substrates without native skill loaders.

### Maintainability

- **[suggestion]** `src/skill-projection.ts:237` — **Extract duplicated catalog refresh**
  The loop `for (const substrate of options.substrates) {` duplicates the same refresh block later in `unprojectSkill`, including `const catalog = findCatalogFile(substrate, input, projectionOptions);` and the same `process.stderr.write(...)` warning path. Future changes to catalog discovery or write handling must be kept in sync across both functions.
  Fix: Extract a shared `refreshSkillCatalogs(options, somaHome, input)` helper returning `catalogFiles`.

### HonestOracle

- **[blocker]** `src/skill-projection.ts:296` — **Registry symlink ownership is assumed**
  The comment says, "Remove the registry symlink — only if it IS a symlink Soma created," but the code removes any registry symlink without checking its target or provenance. This misleads downstream users into believing authored registry symlinks are preserved when `unprojectSkill` will delete them.


---

jcfischer:
Addressed in b7bc34a:

- **[blocker] unproject dry-run still writes** — fixed. `unproject-skill` now defaults to dry-run and only mutates on `--apply`, symmetric with `project-skill`. Added `planUnprojectSkill` + `apply` on the parsed args; e2e-confirmed dry-run leaves links intact.
- **[important] symlinked skill dirs rejected** — fixed. Arg→name resolution now uses `stat` (follows symlinks), so `unproject-skill ~/.soma/skills/<Name>` resolves the registry symlink to its skill name. New test covers it.
- **[important] primitive bypasses adapters** — agreed, deferred to #356. Slice 1 keeps the working primitive; moving the loader-root/builder behind a `SubstrateInstallSpec` method is a focused boundary cleanup.
- **[important] PR delegation claim unsupported** — fixed. PR description now states this PR does NOT wire `install --skills`/arc; they're future slices that will call the primitive.
- **[suggestion] duplicated catalog refresh** — extracted `refreshSkillCatalogs` (shared by project + unproject) and `formatPlan`.
- **[suggestion] project-only parser state** — both verbs now carry `apply`, so the shared parser is no longer command-specific.

Full suite 1492/0, typecheck + lint clean. Ready for re-review.

jcfischer (commented):

### CodeQuality

- **[important]** `src/skill-projection.ts:344` — **Registry force is ignored**
  The option promises `/** Remove a real (non-symlink) directory at the slot; off by default. */ force?: boolean;`, but the registry branch does `} else { links.push({ scope: "registry", path: registryDest, status: "absent" }); }`. With `--force`, a real registry dir is left in place and the catalog refresh keeps listing the skill.
  Fix: Handle non-symlink registry entries with `force` the same way `removeLink(dest, force)` handles substrate entries.

- **[suggestion]** `src/skill-projection.ts:317` — **Unproject plan overstates removals**
  `links.push({ scope: "registry", path: join(registrySkillsDir(somaHome), name), target: "" });` always reports a registry removal, but apply intentionally leaves authored registry directories intact. The default dry-run can therefore show a registry path as removed when `unprojectSkill` will preserve it.
  Fix: Have `planUnprojectSkill` inspect the registry entry and only list removable links, or mark preserved real dirs distinctly.

### Maintainability

- **[suggestion]** `src/skill-projection.ts:217` — **Deduplicate link destination planning**
  The apply path builds destinations with `const registryDest = join(registryDir, name);` and `const dest = join(substrateSkillsRoot(substrate, substrateHome), name);`, while the plan path repeats them as `path: join(registryDir, name)` and `path: join(substrateSkillsRoot(substrate, substrateHome), name)`. A future loader-root or registry-slot change now has to be mirrored in both dry-run and apply code.
  Fix: Extract shared link-plan builders used by both planProjectSkill and projectSkill.

### HonestOracle

- **[important]** `src/skill-projection.ts:22` — **Delegation claim remains false**
  The comment claims "`soma install --skills`, and arc, both delegate to this" while the PR description says it "does not yet wire `soma install --skills` or arc". This misleads downstream maintainers about the current routing and overstates this slice as the existing single projection truth.


---

jcfischer (commented):

4 finding(s): 1 blocker, 1 important, 1 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/cli/skill-projection-cli.ts:145` — **Empty catalog refresh renders poorly**
  `Catalog refresh: ${plan.catalogRefresh.join(", ")}` emits `Catalog refresh: ` if callers pass an empty substrate list, which is possible through the library types even if the CLI parser usually prevents it. That makes dry-run output ambiguous rather than reporting that no catalog refresh is planned.
  Fix: Render a fallback such as `none` when `plan.catalogRefresh.length === 0`.

### Architecture

- **[important]** `src/cli/skill-projection-cli.ts:37` — **Public verb drifts from lifecycle contract**
  CONTEXT.md section "Lifecycle verbs: install, reproject, upgrade, load, uninstall" defines the public lifecycle verbs, but the diff exposes `Usage: soma unproject-skill <skill-dir|name>` as a new top-level removal verb. Once scripts depend on this command, renaming it to the documented lifecycle surface requires compatibility baggage.
  Fix: Keep `unprojectSkill` library-only for this slice, or expose removal through the documented `uninstall` surface after updating CONTEXT.md.

- **[nit]** `src/skill-projection.ts:30` — **Avoid banned sync vocabulary**
  CONTEXT.md's projection/writeback vocabulary kills `sync` as fuzzy, but the comment says `no re-sync`. This is a glossary drift in code comments.
  Fix: Use `without reprojecting`.

### HonestOracle

- **[blocker]** `src/skill-projection.ts:17` — **Delegation claim remains false**
  The comment says, "`soma install --skills`, and arc, both delegate to this" and "soma stays the single projection truth," but the PR description says this slice "does not yet wire `soma install --skills` or arc." That misleads downstream readers into treating this primitive as already enforced system behavior rather than an unwired design intent.


---

jcfischer (commented):

3 finding(s): 3 important.


### Architecture

- **[important]** `src/cli.ts:183` — **Public lifecycle verb bypasses glossary**
  The diff adds top-level CLI surface `"unproject-skill"`, but CONTEXT.md's Lifecycle verbs section says `install, reproject, upgrade, load, uninstall` are the five lifecycle verbs with `No overlap`, and defines `uninstall` as adapter removal of a projection. Shipping `unproject-skill` makes a non-canonical lifecycle verb public API, so later alignment requires a breaking command rename or permanent alias.
  Fix: Expose the removal path under an `uninstall`-aligned command or keep `unprojectSkill` library-only until the glossary adds this lifecycle verb.

### HonestOracle

- **[important]** `src/skill-projection.ts:24` — **Stale delegation claim remains**
  The comment claims "`soma install --skills`, and arc, both delegate to this" while the PR description says they are not wired yet. That misleads readers into treating this primitive as the current single projection path.

- **[important]** `src/skill-projection.ts:253` — **Catalog drift only warns**
  The comment says a catalog miss will "surface it rather than silently skipping the catalog refresh," but the code only writes "Warning: no SKILLS catalog file found" and then `continue`s. A caller can receive a successful projection result even though the skill is not listed.


---

jcfischer:
Addressed in $(git rev-parse --short HEAD):

- **[important] registry force ignored** — fixed. The registry branch now removes a real authored dir under `--force` (status `preserved` without it), symmetric with `removeLink(dest, force)`. New test covers force-removal + plan honesty.
- **[important] delegation claim in code comment** — fixed. The module doc comment no longer says install/arc delegate today; it states this slice ships the primitive only (matches the corrected PR description).
- **[suggestion] plan overstates removals** — `planUnprojectSkill` now lists the registry link only when it would actually be removed (symlink, or real dir under `--force`); a preserved authored dir is no longer shown as a removal.
- **[suggestion] dedup link planning** — extracted `skillSlots()` as the single source for registry + per-substrate loader paths; plan and apply both consume it.

Full suite 1493/0, typecheck + lint clean.

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/skill-projection.ts:179` — **Symlink ownership is asserted, not proven**
  The comment says "A symlink in the slot is Soma-owned (or a prior link) — replace freely," but the code only checks `stat.isSymbolicLink()` and never verifies target, provenance, or Soma ownership before `await rm(linkPath, { recursive: true, force: true })`. This misleads downstream readers into believing user-created symlinks are protected when they are silently replaced.


---

jcfischer (commented):

1 finding(s): 1 suggestion.


### HonestOracle

- **[suggestion]** `PR description:23` — **Verification claim lacks evidence**
  The description claims "typecheck PASS · lint clean · 11 unit tests · full suite **1492/0** · e2e: dry-run/apply/force/unproject all confirmed" without attaching command output, CI links, or reproducible logs. Downstream reviewers cannot distinguish actual verification from an unchecked status line.


---

jcfischer (commented):

### Performance

- **[suggestion]** `src/skill-projection.ts:217` — **Avoid repeated catalog rendering**
  `const expected = renderSkills(input);` runs inside `findCatalogFile`, which is called from `for (const substrate of substrates) {`, so multi-substrate refresh allocates the full SKILLS catalog once per substrate.
  Fix: Compute `renderSkills(input)` once in `refreshSkillCatalogs` and pass it into `findCatalogFile`.

- **[suggestion]** `src/skill-projection.ts:264` — **Parallelize catalog writes**
  `for (const substrate of substrates) {` performs `const written = await writeProjection(...)` serially, so multi-substrate apply latency becomes the sum of independent filesystem writes.
  Fix: Create catalog write tasks per substrate and `await Promise.all(...)`.

### Maintainability

- **[suggestion]** `src/cli/skill-projection-cli.ts:16` — **Extract shared option shape**
  The option literal `options: { homeDir?: string; somaHome?: string; substrateHome?: string; force?: boolean };` is repeated for project, unproject, and parser state, so adding another shared flag requires multiple edits.
  Fix: Introduce a `SkillProjectionCliOptions` interface and reuse it.

### HonestOracle

- **[important]** `PR description:29` — **Verification claim has no evidence**
  The PR states "typecheck PASS · lint clean · 11 unit tests · full suite **1492/0** · e2e: dry-run/apply/force/unproject all confirmed" but provides no command output, CI link, transcript, or reproducible command list. Downstream reviewers are being asked to rely on a surrogate assertion rather than inspectable verification evidence.


---

jcfischer:
Addressed in `3135f05`: **[important] symlink ownership asserted not proven** — reworded the `ensureSymlink` comment to match the code. It no longer claims Soma-ownership; it states the slot is owned by project-skill and any symlink (including a user's) is replaced — which loses no data (the link's target dir is untouched), whereas real dirs are the data-loss case guarded by `--force`. No behavior change; the comment was the overclaim.

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/skill-projection.ts:20` — **Unrelated user skills claim is false**
  The comment claims "unrelated user skills are never touched," but `ensureSymlink` replaces "Any symlink in the slot" for the same derived name, including a user-created symlink pointing elsewhere. A downstream reader would rely on a stronger safety guarantee than the code enforces.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/skill-projection.ts:151` — **Name parsing escapes frontmatter**
  The code claims skill identity is frontmatter, but `const match = /^name:\s*["']?(.+?)["']?\s*$/m.exec(content);` scans the entire SKILL.md body. A body line or code block containing `name:` can override the basename even when the frontmatter omits a name.
  Fix: Parse only the YAML frontmatter block before matching `name`.

### Maintainability

- **[suggestion]** `src/skill-projection.ts:92` — **Split create and remove plan shapes**
  `links: { scope: "registry" | "substrate"; substrate?: InstallSubstrate; path: string; target: string }[];` forces unproject plans to use sentinel values like `links.push({ scope: "substrate", substrate, path, target: "" });`. That makes consumers infer intent from an empty string instead of the type carrying the operation.
  Fix: Use separate plan link variants, e.g. `{ action: "link"; target: string } | { action: "remove" }`.

### HonestOracle

- **[important]** `PR description:6` — **Identity claim omits fallback**
  The description says "Identity = frontmatter `name`.", but `src/skill-projection.ts` implements "falling back to the dir basename." This misleads downstream callers about the actual identity contract for skills without a `name` field.


---

jcfischer (commented):

### HonestOracle

- **[important]** `src/skill-projection.ts:21` — **Replacement claim contradicts force guard**
  The module comment claims "a pre-existing copy at that name ... is replaced", but `ensureSymlink` refuses non-symlink paths unless `force` is set. That misleads API readers into expecting automatic migration of hand-made copies when the default behavior is refusal.

- **[suggestion]** `src/skill-projection.ts:318` — **Absent-path fallback is overstated**
  The comment says "Path-like but absent — fall back to treating it as a bare name", but a path-like missing arg such as `./MyTool` still reaches `assertSafeSkillName(name)` with a slash and errors. The documented fallback is not what the shown code does.


---

jcfischer:
Addressed in `b336041`: **[important] 'unrelated user skills' overclaim** — reworded the module comment to draw the exact line: a differently-named skill is never touched; at the same `<name>` slot, a symlink is replaced (no data loss) and a real dir is preserved unless `--force`. Matches the `ensureSymlink` comment and the actual behavior.

Issue #370, opened 2026-07-02
Provenance headers and doctor drift checks for all substrate projections


jcfischer:
## Problem

Generated projection files are indistinguishable from hand-maintained ones, so drift is silent. The 2026-07 migration repeatedly found stale claims in substrate instruction files that a doctor check would have caught.

## Proposal

1. Every generated projection file (all substrates) carries a provenance header: generated-by, source of truth, refresh command, "not authoritative".
2. `soma doctor --substrate <s>` warns on: stale projections (source newer than projection), unmanaged drift (projection edited outside overlay), and installed-package vs source-checkout divergence.

## Coordination

The memory plan's M4 (Plans/2026-07-02-memory-subsystem-implementation-plan-v2.md) adds `rules/soma/MEMORY.md` to the claude-code rules bundle and depends on byte-idempotent projection. Provenance headers must not break that idempotency: no wall clock timestamps in projected bytes; derive from stored content or make the stamp stable per source state.

## Acceptance

- Reprojecting twice with unchanged sources stays byte identical, headers included.
- `soma doctor --substrate claude-code` flags a deliberately hand-edited projection and a deliberately stale one in a test fixture.
- Doctor exit code distinguishes clean / warnings / errors for CI use.

Issue #368, opened 2026-07-02
Claude Code adapter: generate CLAUDE.md as a projection


jcfischer:
## Problem

`~/.claude/CLAUDE.md` is hand maintained today. It drifts from Soma source, silently loses content, and duplicates what the adapter already projects into `~/.claude/rules/soma/`. The 2026-07 harness migration converted its content to Soma sources; the file itself must become a generated projection.

## Proposal

The claude-code adapter writes `~/.claude/CLAUDE.md` from Soma source with:

1. A provenance header: generated-by, source of truth `~/.soma`, generation input hash, refresh command, and a "this file is not authoritative, edit ~/.soma instead" line.
2. A preserved substrate-local overlay section (clearly delimited, e.g. `<!-- soma:overlay:start -->` ... `<!-- soma:overlay:end -->`) that survives regeneration byte for byte.

## Acceptance

- Reprojecting twice with unchanged sources is byte idempotent (no wall clock timestamps in projected bytes).
- A hand edit inside the overlay section survives reprojection.
- A hand edit outside the overlay is either reverted with a warning printed by `soma install claude-code`, or rejected with a diff shown.
- Greenfield: installing into a directory with no existing CLAUDE.md produces a valid file.

Part of the 2026-07 one-portable-assistant migration (Phase 4 depends on this).

PR #377, opened 2026-07-02
feat(adapters): provenance headers + doctor unmanaged-edit check (#370)

jcfischer:
Closes part of #370 (provenance primitive + doctor drift). Foundation for #368.

## What

- **`src/adapters/shared/provenance.ts`** — shared primitive: a byte-stable `Generated by Soma` header (no timestamp, so reprojection stays byte-identical), overlay markers for hand-maintained regions, and helpers (`withProvenance`, `stripProvenance`, `hasProvenanceHeader`, `extractOverlay`, `renderOverlay`, `hasUnmanagedEdit`), re-exported from the `shared` barrel.
- **claude-code home projection** wraps each `rules/soma` skeleton file in the header. The byte-portable `active-vsa` artifact is deliberately excluded (cross-substrate byte-equality + leading frontmatter).
- **skill-projection catalog finder** is header-tolerant via `stripProvenance`, so `SKILLS.md` resolves while carrying provenance.
- **claude-code doctor** gains `claude-code-projection-unmanaged-edit`: a present, non-stale projection that lost its header was hand-replaced and would be silently overwritten on reproject — warn with a remediation command.

## Tests

- New `test/provenance.test.ts` (header stability, overlay round-trip, stripProvenance round-trip, unmanaged-edit compare).
- Doctor fixture updated to write a real header-carrying projection; positive unmanaged-edit test added.
- Validation: this PR's CI checks (Deterministic portability, confidentiality-gate) are green, and the provenance + doctor tests are included in the diff. Locally `bun test` was 1524 pass / 0 fail and `tsc` clean at author time; the CI checks are the reproducible record.

## Scope notes

- Other substrates adopt the same header in follow-up (this PR wires claude-code, the migration's substrate).
- The `installed-package vs source-checkout divergence` doctor check from #370 is deferred pending a version signal; noted on the issue.

## Byte-idempotency

Header carries no wall-clock timestamp, so the AC-4 idempotency tests (`install` twice = identical bytes) still hold — verified green.


jcfischer (commented):

### CodeQuality

- **[important]** `src/adapters/shared/provenance.ts:56` — **stripProvenance drops leading content newlines**
  `return content.slice(headerEnd + 3).replace(/^\n+/, "");` removes every newline after the header, including newlines that belong to the original content. That breaks the documented round-trip for inputs like `withProvenance("claude-code", "\n# Body")`, which would strip to `# Body` instead of `\n# Body`.
  Fix: Remove only the separator inserted by `withProvenance`, e.g. slice `headerEnd + 5` when the header is followed by `\n\n`, or otherwise preserve the remainder exactly.

### Architecture

- **[suggestion]** `src/adapters/shared/provenance.ts:36` — **Use reproject vocabulary**
  The generated header says `run \`soma install ${substrate} --apply\` to refresh`, but CONTEXT.md's `Lifecycle verbs: install, reproject, upgrade, load, uninstall` section reserves `install` for first-time setup and kills `refresh` in favor of `reproject`. This makes a cross-substrate generated header teach the wrong lifecycle boundary.
  Fix: Use `soma reproject ${substrate}` and say `to reproject`.

### Maintainability

- **[suggestion]** `src/adapters/shared/provenance.ts:75` — **Defer unused overlay API**
  The diff adds overlay API that no production path calls: `export const OVERLAY_BEGIN`, `export function extractOverlay(existing: string): string | null`, `export function renderOverlay(body: string | null): string`, and `export function hasUnmanagedEdit(content: string, expectedManaged: string): boolean`. Re-exporting these follow-up-only helpers creates public surface and tests for behavior this PR cannot exercise.
  Fix: Land only the provenance helpers used now, or keep overlay helpers private until #368 wires them in.

### HonestOracle

- **[important]** `src/adapters/shared/provenance.ts:38` — **Substring is not header evidence**
  The doc claims `True when content still carries a Soma provenance header.`, but the implementation is only `return content.includes(PROVENANCE_MARKER);`. A file with the header removed but the words `Generated by Soma` anywhere in user content will be treated as managed, so the doctor’s missing-header drift signal can fail open.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/adapters/claude-code/doctor.ts:55` — **Doctor only checks one managed file**
  The new projection wraps every skeleton file with provenance via `...skeleton.map((file) => ({ ...file, content: withProvenance("claude-code", file.content) }))`, but doctor only reads `const contextRaw = await readFileOrNull(contextPath);` before deciding whether the projection lost its header. A hand-replaced `SKILLS.md` or other generated skeleton file will still be silently overwritten without the intended warning.
  Fix: Check all managed `rules/soma` skeleton files for missing provenance, or narrow the projection change so only the context file is treated as doctor-managed.

### Maintainability

- **[suggestion]** `test/onboarding-doctor.test.ts:346` — **Use provenance helper in fixture**
  The fixture reimplements the helper contract with "`${provenanceHeader("claude-code")}\n\n# Soma Claude Code Context\n`", duplicating the separator convention owned by `withProvenance`. If the wrapping format changes, this fixture can drift from production rendering.
  Fix: Use `withProvenance("claude-code", "# Soma Claude Code Context\n")`.

### HonestOracle

- **[important]** `src/adapters/claude-code/doctor.ts:54` — **Missing header is sold as hand edit**
  The comment says "a present-but-header-less projection was replaced by hand," and the warning says "Claude Code projection was edited by hand," but the check only proves `!hasProvenanceHeader(contextRaw)` for `.claude/rules/soma/CONTEXT.md`. That misleads downstream users because header absence can also come from an older projection or partial migration, while the doctor reports a specific human edit as fact.


---

jcfischer (commented):

### Maintainability

- **[suggestion]** `src/adapters/shared/provenance.ts:43` — **Deduplicate header parsing**
  `hasProvenanceHeader` parses the leading comment with `const trimmed = content.trimStart();` while `stripProvenance` repeats a similar parser with `if (!content.startsWith("<!--")) return content;`. Keeping two slightly different parsers raises future-change cost for the provenance format.
  Fix: Extract a small `findProvenanceHeader(content)` helper returning the header end/index and reuse it in both functions.

### HonestOracle

- **[important]** `src/adapters/shared/provenance.ts:17` — **Overlay preservation is only asserted**
  The comment claims "Files that support an overlay ... keep everything between the markers across regeneration; edits outside the markers are unmanaged drift," but the diff only adds `extractOverlay`/`renderOverlay` and tests them in isolation. No projection code in this PR reads an existing file, extracts the overlay, or reinserts it on reprojection, so the documented preservation guarantee is not shown.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `PR description:14` — **Unsubstantiated validation claim**
  The description states, "Full suite: 1524 pass, 0 fail. `tsc` clean. Touched files eslint-clean." No CI link, command transcript, or reproducible evidence is included in the PR data, so reviewers are asked to rely on an asserted validation result rather than evidence.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### Performance

- **[suggestion]** `src/adapters/claude-code/doctor.ts:67` — **Parallelize doctor file reads**
  The loop `for (const relativePath of PROVENANCE_MANAGED_RULES_FILES)` awaits `const raw = await readFileOrNull(join(substrateHome, relativePath));` one file at a time, so doctor latency becomes the sum of each filesystem read. These files are independent, so the check can run them concurrently without changing behavior.
  Fix: Use `Promise.all(PROVENANCE_MANAGED_RULES_FILES.map(async (relativePath) => ...))` and filter the missing-header results.

### HonestOracle

- **[suggestion]** `src/adapters/shared/provenance.ts:82` — **Substrate header contract overstates**
  The contract says "Prefix `content` with the provenance header for `substrate`", but the implementation returns any existing Soma header unchanged via "if (hasProvenanceHeader(content)) return content;". A caller passing content with a different substrate header keeps the wrong refresh command while believing this helper normalized it.


---

PR #393, opened 2026-07-03
feat(memory): write+verify CLI with recall-first refusal (M1)

jcfischer:
M1 — soma memory write|verify for durable notes (semantic+procedural), plan v2 §M1.

Governance: trust DERIVED from --trigger (no --trust flag); principal-correction additionally requires --principal-authority (deliberate, logged escalation — not crypto auth; soma has no principal-auth primitive). Mutating an existing principal note (merge/supersede) needs the same escalation. import→quarantined, consolidation→assistant. tool/import provenance refused under principal-correction (MINJA). Recall-first refusal: exact-body-hash OR Jaccard≥0.6 over active notes, --force overrides. Ids globally unique across types. One mutation → one event, rolled back if the append fails.

Verification (local; repo runs no CI on this PR): 'bun test test/memory-write.test.ts' → 26 pass (deterministic, reproducible). tsc + eslint clean on changed files. Full 'bun test' suite green except (a) grok-hook tests that flake on a 5s per-test timeout under heavy parallel machine load — all 50 pass at --timeout 30000 — and (b) pre-existing fixtures that read a real ~/work/PAI tree. Neither touches the memory subsystem.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -104,8 +104,8 @@ The seven compartments:
 
 Within the **Memory** compartment, the `work/knowledge/learning/relationship/state`
 stores hold curated free-form material. The **memory-note subsystem** is a
-further store *within* Memory (not a new compartment): schema-governed, atomic,
-trust- and decay-tracked notes. Mechanics (paths, frontmatter, the
+further store *within* Memory (not a new compartment): schema-governed,
+single-fact, trust- and decay-tracked notes. Mechanics (paths, frontmatter, the
 `soma memory write|verify` surface) live in `docs/architecture.md`.
 
 **Why:** Coheres with the cellular metaphor (a cell body has compartments — nucleus, mitochondria, etc., each doing its own job, not stacked or ordered). Captures the right shape: seven peer subdivisions, each with its own contract, no implied ordering. `layer` implies a stack (wrong); `store` implies storage and breaks for Policy and Learning; `domain` collides with DDD vocabulary already in use across the codebase.
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -103,13 +103,10 @@ The seven compartments:
 **Not synonyms:** Do not use `layer`, `store`, `domain`, `component`, `module`, `section`, `part` as glossary terms for these. `compartment` is the only word.
 
 Within the **Memory** compartment, the `work/knowledge/learning/relationship/state`
-stores hold curated free-form material. Alongside them lives the **memory-note
-subsystem** (plan v2, milestones M0–M7): a schema-governed durable store of
-atomic notes under `memory/{semantic,procedural,episodic}/`, each with
-bi-temporal frontmatter (trust, provenance, `valid_until`, `last_verified`,
-`resurface_count`). It is a store *within* Memory, not a new compartment; its
-sole write path is `soma memory write|verify` (trust derived from the write
-trigger, dedup-gated, one event per mutation). See `docs/architecture.md`.
+stores hold curated free-form material. The **memory-note subsystem** is a
+further store *within* Memory (not a new compartment): schema-governed, atomic,
+trust- and decay-tracked notes. Mechanics (paths, frontmatter, the
+`soma memory write|verify` surface) live in `docs/architecture.md`.
 
 **Why:** Coheres with the cellular metaphor (a cell body has compartments — nucleus, mitochondria, etc., each doing its own job, not stacked or ordered). Captures the right shape: seven peer subdivisions, each with its own contract, no implied ordering. `layer` implies a stack (wrong); `store` implies storage and breaks for Policy and Learning; `domain` collides with DDD vocabulary already in use across the codebase.
 
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -102,6 +102,15 @@ The seven compartments:
 
 **Not synonyms:** Do not use `layer`, `store`, `domain`, `component`, `module`, `section`, `part` as glossary terms for these. `compartment` is the only word.
 
+Within the **Memory** compartment, the `work/knowledge/learning/relationship/state`
+stores hold curated free-form material. Alongside them lives the **memory-note
+subsystem** (plan v2, milestones M0–M7): a schema-governed durable store of
+atomic notes under `memory/{semantic,procedural,episodic}/`, each with
+bi-temporal frontmatter (trust, provenance, `valid_until`, `last_verified`,
+`resurface_count`). It is a store *within* Memory, not a new compartment; its
+sole write path is `soma memory write|verify` (trust derived from the write
+trigger, dedup-gated, one event per mutation). See `docs/architecture.md`.
+
 **Why:** Coheres with the cellular metaphor (a cell body has compartments — nucleus, mitochondria, etc., each doing its own job, not stacked or ordered). Captures the right shape: seven peer subdivisions, each with its own contract, no implied ordering. `layer` implies a stack (wrong); `store` implies storage and breaks for Policy and Learning; `domain` collides with DDD vocabulary already in use across the codebase.
 
 ---
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -131,8 +131,11 @@ memory/
 
 Each note is one file: strict frontmatter (id, type, trust, provenance,
 bi-temporal `valid_until`, `last_verified`, `resurface_count`, links) plus a
-markdown body. `soma memory write|verify` (M1) is the only write path — trust is
-derived from the write trigger, writes are dedup-gated (recall-first refusal),
+markdown body. `soma memory write|verify` (M1) is the only *governed* write path
+(a convention, not a filesystem-enforced guarantee — nothing stops out-of-band
+edits to the markdown, which is why every recalled note carries a verification
+banner). Through it, trust is derived from the write trigger, writes are
+dedup-gated (recall-first refusal),
 and each mutation appends one event to the **existing** `memory/STATE/events.jsonl`
 stream (the same journal the Observability section reads — note mutations do not
 fork a second event stream). The write/event coupling is best-effort, not
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -115,8 +115,9 @@ MEMORY/
 The initial version should avoid requiring a vector database. Search can start
 with filenames, frontmatter, ripgrep, and small deterministic indexes.
 
-The `MEMORY/*` compartments above are the legacy tier (free-form markdown per
-compartment, searched by `soma memory search`). The **memory-note subsystem**
+The `MEMORY/*` stores above are the legacy tier (free-form markdown per store,
+searched by `soma memory search`) — sub-stores within the single Memory
+compartment, not peer Soma compartments. The **memory-note subsystem**
 (plan v2, milestones M0–M7) is a *separate*, schema-governed durable store that
 lives alongside them under lowercase `memory/`:
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -133,11 +133,15 @@ Each note is one file: strict frontmatter (id, type, trust, provenance,
 bi-temporal `valid_until`, `last_verified`, `resurface_count`, links) plus a
 markdown body. `soma memory write|verify` (M1) is the only write path — trust is
 derived from the write trigger, writes are dedup-gated (recall-first refusal),
-and every mutation appends exactly one event to the **existing**
-`memory/STATE/events.jsonl` stream (the same journal the Observability section
-reads — note mutations do not fork a second event stream). This taxonomy is intentionally
-distinct from the `MEMORY/*` stores: those stores hold curated free-form
-material; the note store holds atomic, governed, decay-tracked notes.
+and each mutation appends one event to the **existing** `memory/STATE/events.jsonl`
+stream (the same journal the Observability section reads — note mutations do not
+fork a second event stream). The write/event coupling is best-effort, not
+crash-atomic: an event-append *failure* rolls the file mutation back, but a hard
+process crash in the window between the two can still orphan a file from its
+event (a documented gap reconciled by the M7 audit; soma has no WAL/2PC). This
+taxonomy is intentionally distinct from the `MEMORY/*` stores: those stores hold
+curated free-form material; the note store holds single-fact, governed,
+decay-tracked notes.
 
 Cross-machine Soma state uses **Home replication**, not projection refresh or
 substrate writeback. The design is Git-backed first, policy-gated per scope,
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -126,14 +126,15 @@ memory/
   procedural/<id>.md     # playbooks / how-to   (dedup-gated, verifiable)
   episodic/…             # session digests + action log (M5)
   INDEX.md               # earned-inclusion index (M3)
-  state/events.jsonl     # append-only mutation journal
 ```
 
 Each note is one file: strict frontmatter (id, type, trust, provenance,
 bi-temporal `valid_until`, `last_verified`, `resurface_count`, links) plus a
 markdown body. `soma memory write|verify` (M1) is the only write path — trust is
 derived from the write trigger, writes are dedup-gated (recall-first refusal),
-and every mutation appends exactly one event. This taxonomy is intentionally
+and every mutation appends exactly one event to the **existing**
+`memory/STATE/events.jsonl` stream (the same journal the Observability section
+reads — note mutations do not fork a second event stream). This taxonomy is intentionally
 distinct from the `MEMORY/*` compartments: the compartments hold curated
 free-form material; the note store holds atomic, governed, decay-tracked notes.
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -136,8 +136,8 @@ derived from the write trigger, writes are dedup-gated (recall-first refusal),
 and every mutation appends exactly one event to the **existing**
 `memory/STATE/events.jsonl` stream (the same journal the Observability section
 reads — note mutations do not fork a second event stream). This taxonomy is intentionally
-distinct from the `MEMORY/*` compartments: the compartments hold curated
-free-form material; the note store holds atomic, governed, decay-tracked notes.
+distinct from the `MEMORY/*` stores: those stores hold curated free-form
+material; the note store holds atomic, governed, decay-tracked notes.
 
 Cross-machine Soma state uses **Home replication**, not projection refresh or
 substrate writeback. The design is Git-backed first, policy-gated per scope,
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -115,6 +115,28 @@ MEMORY/
 The initial version should avoid requiring a vector database. Search can start
 with filenames, frontmatter, ripgrep, and small deterministic indexes.
 
+The `MEMORY/*` compartments above are the legacy tier (free-form markdown per
+compartment, searched by `soma memory search`). The **memory-note subsystem**
+(plan v2, milestones M0–M7) is a *separate*, schema-governed durable store that
+lives alongside them under lowercase `memory/`:
+
+```text
+memory/
+  semantic/<id>.md      # durable facts        (dedup-gated, verifiable)
+  procedural/<id>.md     # playbooks / how-to   (dedup-gated, verifiable)
+  episodic/…             # session digests + action log (M5)
+  INDEX.md               # earned-inclusion index (M3)
+  state/events.jsonl     # append-only mutation journal
+```
+
+Each note is one file: strict frontmatter (id, type, trust, provenance,
+bi-temporal `valid_until`, `last_verified`, `resurface_count`, links) plus a
+markdown body. `soma memory write|verify` (M1) is the only write path — trust is
+derived from the write trigger, writes are dedup-gated (recall-first refusal),
+and every mutation appends exactly one event. This taxonomy is intentionally
+distinct from the `MEMORY/*` compartments: the compartments hold curated
+free-form material; the note store holds atomic, governed, decay-tracked notes.
+
 Cross-machine Soma state uses **Home replication**, not projection refresh or
 substrate writeback. The design is Git-backed first, policy-gated per scope,
 and only auto-merges stores with deterministic merge semantics. See
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -101,33 +101,32 @@ registries, while the personal Soma home remains owned by one principal. See
 
 ### Memory
 
-Memory is structured as files first:
+Memory is structured as files first, under a single lowercase `memory/` root
+(one canonical root — the layout never relies on distinguishing `memory` from
+`MEMORY`, which case-insensitive filesystems collapse):
 
 ```text
-MEMORY/
-  WORK/
+memory/
+  WORK/            # legacy free-form stores (searched by `soma memory search`)
   KNOWLEDGE/
   LEARNING/
   RELATIONSHIP/
-  STATE/
+  STATE/           # events.jsonl lives here
+  semantic/<id>.md   # memory-note subsystem (M0–M7): durable facts   (dedup-gated)
+  procedural/<id>.md #   playbooks / how-to                            (dedup-gated)
+  episodic/…         #   session digests + action log (M5)
+  INDEX.md           #   earned-inclusion index (M3)
 ```
 
 The initial version should avoid requiring a vector database. Search can start
 with filenames, frontmatter, ripgrep, and small deterministic indexes.
 
-The `MEMORY/*` stores above are the legacy tier (free-form markdown per store,
-searched by `soma memory search`) — sub-stores within the single Memory
-compartment, not peer Soma compartments. The **memory-note subsystem**
-(plan v2, milestones M0–M7) is a *separate*, schema-governed durable store that
-lives alongside them under lowercase `memory/`:
-
-```text
-memory/
-  semantic/<id>.md      # durable facts        (dedup-gated, verifiable)
-  procedural/<id>.md     # playbooks / how-to   (dedup-gated, verifiable)
-  episodic/…             # session digests + action log (M5)
-  INDEX.md               # earned-inclusion index (M3)
-```
+The uppercase-named legacy stores (`WORK`, `KNOWLEDGE`, …) hold free-form
+markdown; the **memory-note subsystem** (plan v2, milestones M0–M7) is a
+*separate*, schema-governed durable store whose lowercase-named directories
+(`semantic`, `procedural`, `episodic`) sit as siblings under the same `memory/`
+root. Both are sub-stores within the single Memory compartment, not peer Soma
+compartments.
 
 Each note is one file: strict frontmatter (id, type, trust, provenance,
 bi-temporal `valid_until`, `last_verified`, `resurface_count`, links) plus a
--- a/src/cli/memory.ts
+++ b/src/cli/memory.ts
@@ -4,7 +4,7 @@ import {
   verifyMemoryNote,
   writeMemoryNote,
 } from "../index";
-import { SOMA_MEMORY_WRITE_TRIGGERS } from "../types";
+import { WRITABLE_NOTE_TYPES, isWritableNoteType } from "../memory-write";
 import type {
   SomaMemoryNoteType,
   SomaMemoryPromotionOptions,
@@ -60,14 +60,15 @@ export const MEMORY_COMMAND_HELP: { usage: string; subcommands: Record<MemoryAct
     search: "Usage: soma memory search [query] [--query <text>] [--limit <n>] [--home-dir <dir>] [--soma-home <dir>]",
     promote: "Usage: soma memory promote --from-run <run-id> --store <learning|knowledge|relationship|work> --title <text> [--lesson <text>] [--applies-when <text>]",
     write:
-      "Usage: soma memory write --trigger <principal-correction|import|consolidation> --body <text> " +
+      "Usage: soma memory write --trigger <principal-correction|import> --body <text> " +
       "(create: --id <slug> --type <semantic|procedural> [--force]) " +
       "(--merge <id> | --supersede <id>) " +
       "[--principal-authority] [--project <key>] [--source-of-truth <ref>] [--links a,b] " +
       "[--recall-trigger <text>] [--review <text>] " +
       "[--provenance <import|tool:name>] [--substrate <s>] [--home-dir <dir>] [--soma-home <dir>]. " +
       "Trust is DERIVED from --trigger; there is no --trust flag. principal-correction requires " +
-      "--principal-authority. The consolidation trigger is an internal (M6) SDK path, not available on the public CLI.",
+      "--principal-authority. (The consolidation trigger, which mints assistant trust, is an internal " +
+      "M6 SDK path and is not accepted on the public CLI.)",
     verify:
       "Usage: soma memory verify <id> [--id <id>] [--principal-authority] [--substrate <s>] [--home-dir <dir>] [--soma-home <dir>]. " +
       "Verifying a principal-trust note requires --principal-authority (assistant-trust notes are an internal SDK path).",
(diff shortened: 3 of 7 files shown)


jcfischer (commented):

7 finding(s): 3 blocker, 1 important, 2 suggestion, 1 nit.


### CodeQuality

- **[blocker]** `src/memory-write.ts:344` — **Supersede rollback leaves old note closed**
  On event append failure, `supersedeNote` only runs `await unlink(newPath).catch(() => undefined);` after it has already written `await writeNoteFile(old.path, closed, "w");`. That leaves the target note with `valid_until` set and a forward link even though the supersede operation throws, corrupting memory state.
  Fix: Before writing `closed`, keep the original note content and restore `old.path` in the event-append catch.

### Security

- **[blocker]** `src/memory-write.ts:73` — **Note id allows path traversal**
  The diff builds the output path with user-controlled id: `return join(typeDir(somaHome, type), `${id}.md`);`. Because `--id` is only checked as non-empty before `writeFile`, an id containing `../` can write outside `memory/semantic` or `memory/procedural`.
  Fix: Reject ids that are not safe filename slugs before calling `memoryNotePath`, or resolve and assert the final path stays under `typeDir`.

### Architecture

- **[important]** `src/index.ts:448` — **Internal memory helpers become public API**
  The root export promotes implementation details as stable surface: "findDuplicateCandidates, memoryNotePath, MEMORY_DEDUP_JACCARD_THRESHOLD". This freezes the current file layout and dedup threshold for consumers, making future Memory storage or recall changes breaking API work.
  Fix: Export only writeMemoryNote/verifyMemoryNote and their types; keep path/dedup helpers module-private or test via internal imports.

- **[nit]** `src/types.ts:2339` — **Bare agent term leaks into src**
  CONTEXT.md bans bare agent in docs and code, but the new comment says "so an agent cannot smuggle `principal` trust". Use a qualified substrate term or "assistant" depending on the intended layer.
  Fix: Replace bare "agent" with "assistant" or "substrate-side caller".

### Performance

- **[suggestion]** `src/memory-write.ts:128` — **Duplicate scan is fully sequential**
  `for (const { path, type, note } of await collectDurableNotes(somaHome))` depends on `collectDurableNotes`, which does `const content = await readFile(path, "utf8").catch(() => undefined);` inside nested loops. A create against a large memory corpus waits for every note read/parse before it can write, even though semantic and procedural files are independent.
  Fix: Read files per directory with bounded parallelism or maintain an indexed body-hash/token cache for duplicate checks.

### Maintainability

- **[suggestion]** `src/cli/memory.ts:222` — **Extract shared memory CLI options**
  `parseMemoryWriteArgs` and `parseMemoryVerifyArgs` repeat the same option cases: `case "--home-dir":`, `case "--soma-home":`, and `case "--substrate":` each assign with `readOption(args, index, arg); index += 1; break;`. Keeping this duplicated across subcommands makes every common option addition or behavior change require multiple switch edits.
  Fix: Extract a small helper for common memory CLI options that returns whether it consumed the current arg and the next index.

### HonestOracle

- **[blocker]** `src/memory-write.ts:31` — **Event guarantee is false**
  The diff claims "One mutating call → exactly one events journal line," but `mergeNote`, `verifyMemoryNote`, and the old-note side of `supersedeNote` persist file mutations before appending the event and do not roll them back if event append fails. That misleads downstream readers into treating the event journal as complete when state can change silently.


---

jcfischer (commented):

5 finding(s): 1 blocker, 4 important.


### CodeQuality

- **[important]** `src/memory-write.ts:307` — **Duplicate ids become ambiguous**
  Create only reserves the per-type path via `const path = memoryNotePath(somaHome, note.type as WritableType, note.id);`, but later lookups return the first corpus match with `if (loaded.note.id === id) return loaded;`. A semantic and procedural note can share an id with different bodies, making verify/merge/supersede mutate whichever type is collected first.
  Fix: Reject any existing `note.id === newNote.id` across `collectDurableNotes()` before create/supersede, or require type for id-based mutations.

### Security

- **[blocker]** `src/memory-write.ts:266` — **Principal trust is caller-assertable**
  The diff assigns trust from caller-controlled input with `trust: SOMA_MEMORY_TRIGGER_TRUST[options.trigger]`, while the public CLI accepts `--trigger <principal-correction|import|consolidation>`. Any caller can pass `principal-correction` and create a `principal` note without a human/principal authorization gate.
  Fix: Require an authenticated principal-correction authority signal before mapping the trigger to principal trust.

### Architecture

- **[important]** `src/memory-write.ts:39` — **Memory layout bypasses documented stores**
  docs/architecture.md says Memory is structured as `MEMORY/WORK`, `MEMORY/KNOWLEDGE`, `MEMORY/LEARNING`, `MEMORY/RELATIONSHIP`, and `MEMORY/STATE`, but the new module states writes stay under `memory/semantic/` + `memory/procedural/` and `WRITABLE_TYPE_DIRS` maps durable notes there. This creates a parallel durable Memory taxonomy instead of using the documented compartment stores.
  Fix: Route semantic/procedural note files through the documented Memory store layout, or update the architecture contract in the same PR.

### Performance

- **[important]** `src/memory-write.ts:242` — **Targeted writes scan the whole corpus**
  `loadNoteById` calls `for (const loaded of await collectDurableNotes(somaHome))`, and `collectDurableNotes` reads and parses every `.md` file before the loop can return. This makes `verify`, `merge`, and `supersede` O(corpus size) I/O and allocation for a single known id.
  Fix: Probe `semantic/<id>.md` and `procedural/<id>.md` directly, then parse only the matching file.

### HonestOracle

- **[important]** `src/types.ts:2339` — **Trigger mapping claim contradicts implementation**
  The comment claims "The five write triggers of design §7 map onto write-CLI actions", but the exported `SOMA_MEMORY_WRITE_TRIGGERS` immediately exposes only `"principal-correction", "import", "consolidation"`. This misleads downstream implementers into believing all design §7 triggers are represented by the write CLI when `session-end` and `consequential-action` are explicitly absent.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-write.ts:306` — **Malformed cross-type ids can collide**
  `noteIdExists` treats every `loadNoteById` rejection as absence: `() => false`. Because `loadNoteById` parses existing files via `return { path, type, note: parseMemoryNote(raw), raw };`, a malformed `semantic/foo.md` lets `procedural/foo.md` be created even though ids are meant to be globally unique.
  Fix: Only return false for the not-found case; propagate parse/read errors for existing files.

### Security

- **[blocker]** `src/memory-write.ts:383` — **Merge bypasses principal authority**
  `body: `${note.body}\n\n**Update (${today}):** ${options.body.trim()}`` appends caller-controlled text to the existing note without checking `principalAuthority`. A tool or import caller can merge into a `principal` note and preserve its trust, bypassing the documented principal-authority gate.
  Fix: For `note.trust === "principal"`, require `options.trigger === "principal-correction" && options.principalAuthority === true` before merge.

### Architecture

- **[important]** `docs/architecture.md:128` — **Do not fork the event stream**
  The diff adds `state/events.jsonl     # append-only mutation journal`, but docs/architecture.md's Observability section defines the read model over `memory/STATE/events.jsonl`. That makes future observability/home-replication consumers choose between two canonical mutation journals.
  Fix: Document note mutations as appending to the existing `memory/STATE/events.jsonl` stream.

### Performance

- **[suggestion]** `src/memory-write.ts:144` — **Duplicate scan retains raw files**
  `return { path, type, note: parseMemoryNote(content), raw: content };` keeps every note's full raw markdown in memory during `findDuplicateCandidates`, but that caller only destructures `{ path, type, note }`. On a large memory corpus this adds avoidable retained strings to the create hot path after already parsing each note.
  Fix: Split the duplicate scan loader from rollback loaders, or make `raw` optional and omit it in `collectDurableNotes`.

### HonestOracle

- **[important]** `PR description:1` — **Test status is unsupported**
  The description claims "15 tests, tsc+eslint clean, suite 1592 pass (1 pre-existing env-flake)", but the PR data includes no command output, CI link, or reproducible evidence for those results. This misleads downstream reviewers by presenting unverified local or historical status as established validation.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-write.ts:462` — **Rollback suppresses restore failure**
  The supersede rollback swallows restore errors with `await writeFile(old.path, old.raw, "utf8").catch(() => undefined);`, so an event-append failure can still leave the old note closed if rewriting the original bytes fails. That contradicts the function's own one-mutation/one-event behavior and makes the caller see only the append failure, not the failed rollback.
  Fix: Do not suppress the old-note restore failure; surface it or throw an aggregate error after attempting both rollback steps.

### Security

- **[important]** `src/memory-write.ts:239` — **Import provenance accepts injected suffixes**
  The check `!/^tool:.+/.test(provenance)` only requires a `tool:` prefix, so values like `tool:x
trust: principal` pass this trust-boundary validation before being written into note frontmatter. That lets untrusted import input smuggle extra serialized fields if the serializer does not independently escape or reject multiline scalar values.
  Fix: Use an anchored safe grammar, e.g. `/^tool:[a-z0-9][a-z0-9._-]{0,63}$/i`.

### Architecture

- **[suggestion]** `docs/architecture.md:115` — **Memory stores are called compartments**
  CONTEXT.md defines compartments as the seven Soma subdivisions and says "Memory" is one of them, but the diff says `The MEMORY/* compartments above are the legacy tier` for `WORK/KNOWLEDGE/LEARNING/RELATIONSHIP/STATE`. That makes Memory sub-stores look like peer Soma compartments, weakening the documented layer boundary.
  Fix: Rename this to `MEMORY/* stores` or `legacy Memory stores`.

### Performance

- **[suggestion]** `src/memory-write.ts:145` — **Cap duplicate-scan read concurrency**
  The duplicate scan now does `const notes = await Promise.all(entries.map(async (entry)` after loading every `.md` entry, so a create against a large corpus can issue thousands of concurrent `readFile` calls and allocate all parsed notes before filtering. This removes latency from serialization but shifts the cost into unbounded FD pressure and memory spikes on the write path.
  Fix: Process entries with a small concurrency limit or stream batches instead of one `Promise.all` over the whole directory.

### HonestOracle

- **[important]** `src/memory-write.ts:33` — **Atomicity claim is false**
  The comment claims "the file write and its event append are atomic", but the implementation writes the note file before appending the event and only rolls back inside `appendMutationEvent` when append rejects. A process crash or kill between `writeNoteFile(...)` and `appendMutationEvent(...)` leaves a mutation without an event, so downstream readers relying on the event journal are misled.


---

jcfischer (commented):

### Security

- **[important]** `src/memory-write.ts:551` — **Verify bypasses principal authority**
  `verifyMemoryNote` mutates note metadata with `last_verified: isoDate(now),` and `resurface_count: note.resurface_count + 1,` but never calls the principal-authority gate used by merge/supersede. A tool caller can refresh a `principal` note's decay signals without `--trigger principal-correction` or `--principal-authority`.
  Fix: Require principal authority before verifying principal-trust notes, or make verify a non-principal mutation path with an explicit lower-trust event.

### Architecture

- **[suggestion]** `docs/architecture.md:137` — **Do not call stores compartments**
  CONTEXT.md defines `compartment` as one of the seven Soma subdivisions and lists Memory as a single compartment; the diff says `distinct from the MEMORY/* compartments: the compartments hold curated free-form material`, making Memory sub-stores peer compartments by wording.
  Fix: Say `distinct from the MEMORY/* stores: those stores hold curated free-form material; the note store holds...`

### Performance

- **[suggestion]** `src/memory-write.ts:345` — **Check id before corpus scan**
  `const candidates = await findDuplicateCandidates(somaHome, note.body);` runs before `if (await noteIdExists(somaHome, note.id))`, so a create with an existing id reads and tokenizes the whole durable corpus before failing on a cheap path-existence check.
  Fix: Move `noteIdExists(somaHome, note.id)` before the dedup scan.

### Maintainability

- **[suggestion]** `src/memory-write.ts:190` — **Extract writable type iteration**
  The same casted enumeration appears in multiple helpers: "for (const type of Object.keys(WRITABLE_TYPE_DIRS) as WritableType[])". Repeating that pattern in `collectDurableNotes`, `loadNoteById`, and `noteIdExists` makes future writable-type changes depend on remembering each loop shape.
  Fix: Add `const WRITABLE_TYPES = Object.keys(WRITABLE_TYPE_DIRS) as WritableType[];` and reuse it.

### HonestOracle

- **[blocker]** `src/memory-write.ts:27` — **Import merge stays trusted**
  The claim "`import` is always `quarantined`" is false for merge: `mergeNote` builds the replacement from `{ ...note,` and changes only `last_verified` and `body`, so an import merge into an assistant note preserves assistant trust. That misleads downstream audit readers into treating imported merged content as quarantined when the persisted note remains trusted.

- **[important]** `src/memory-write.ts:25` — **Escalation is not logged**
  The claim "principalAuthority escalation" is "logged" is unsupported for principal merge/supersede events: merge records only `metadata: { id: merged.id, type }`, and supersede records only `metadata: { id: newNote.id, supersededId: closed.id }`. A downstream audit cannot prove the required principal-authority escalation from the event stream.


---

jcfischer (commented):

5 finding(s): 3 important, 2 nit.


### CodeQuality

- **[important]** `src/memory-write.ts:377` — **Rollback can write to wrong target**
  `writeNoteFile` creates parent directories with `await mkdir(join(path, ".."), { recursive: true });`, but `join(path, "..")` treats the note filename as a directory segment and resolves to the parent only by path normalization. In rollback paths that bypass `writeNoteFile`, such as `() => writeFile(path, raw, "utf8")`, a missing parent directory is not recreated, so rollback can fail after an external directory removal even though normal writes recreate parents.
  Fix: Use `dirname(path)` consistently, and route rollback restores through a helper that recreates the parent before writing.

### Architecture

- **[nit]** `src/types.ts:2352` — **Use principal, not human**
  The comment says `the human said` and `documented human-authority gate`; CONTEXT.md:356 says `principal` is the only word for this role in docs/code contexts.
  Fix: Use `principal said` and `principal-authority gate`.

- **[nit]** `src/memory-write.ts:386` — **Qualify agent reference**
  The comment says `an automated/agent invocation defaults safe`, introducing a bare `agent` reference; CONTEXT.md:550-560 requires `agent` to be always qualified.
  Fix: Use `automated assistant invocation` or name the substrate agent.

### HonestOracle

- **[important]** `docs/architecture.md:136` — **Event guarantee overstates crash safety**
  The doc says "every mutation appends exactly one event" and line 140 calls the note store "atomic", but src/memory-write.ts admits "A hard crash there can still leave a file without an event". That misleads downstream audit readers into treating the event journal as complete when the implementation only covers append rejection, not crash interruption.

- **[important]** `src/memory-write.ts:227` — **Dedup gate silently skips malformed notes**
  The PR claims "writes are dedup-gated (recall-first refusal)", but this scanner says "skip unreadable/malformed" and returns undefined for parse failures. A malformed active note becomes invisible to duplicate detection, so the gate fails open without surfacing that the corpus was only partially checked.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/cli/memory.ts:326` — **Conflicting verify ids are ignored**
  The parser accepts both forms but `options.id ??= positionalId;` silently keeps `--id` when a different positional id was also supplied. A command like `soma memory verify old --id new` will verify `new` while ignoring `old`, which is surprising for a mutating command.
  Fix: Reject when both ids are present and differ.

### Performance

- **[suggestion]** `src/memory-write.ts:344` — **Forced writes still scan corpus**
  `const { candidates, unreadable } = await findDuplicateCandidates(somaHome, note.body);` runs unconditionally, and the comment says it scans `even under --force`. That makes `--force` creates pay full O(corpus) directory reads, file reads, parsing, hashing, and tokenization despite bypassing duplicate refusal.
  Fix: For `options.force`, use a cheaper unreadable audit path or skip duplicate scanning unless audit metadata is required.

### Maintainability

- **[suggestion]** `src/memory-write.ts:507` — **Extract event envelope construction**
  Each mutation repeats the same event scaffold: "timestamp: now.toISOString(),", "substrate: options.substrate ?? \"custom\",", "artifactPaths", and "metadata". Keeping that envelope in four call sites makes future journal-shape changes easy to apply inconsistently.
  Fix: Extract a small `appendNoteMutationEvent` helper that fills timestamp, substrate, paths, and shared metadata.

### HonestOracle

- **[important]** `src/memory-write.ts:236` — **Unreadable directories disappear silently**
  The comment claims unreadable notes are "NOT silently dropped", but the directory walk uses `readdir(dir).catch(() => [] as string[])`, so an unreadable durable directory is treated as empty and never reported in `unreadable`. That makes the dedup blind spot silent despite the PR's recall-first refusal claim.

- **[important]** `src/memory-write.ts:618` — **Verify overclaims correctness proof**
  The verify doc says "a resurfaced note that proved correct", but `verifyMemoryNote` accepts only `id`, `now`, substrate/home options, and optional `principalAuthority`; it records no evidence, source, or proof. Downstream readers may treat `last_verified` as correctness-backed when this API only records an assertion.

- **[suggestion]** `src/types.ts:2430` — **Result field doc invents state**
  The public type comment says `supersededId` is "the closed note's new path/state; merge: the pre-edit note", but the field is a `string` and supersede returns only `closed.id` while merge returns no pre-edit note. The API documentation promises data the result cannot carry.


---

jcfischer (commented):

### Performance

- **[suggestion]** `src/memory-write.ts:340` — **Avoid double body scans per note**
  The loop does `const exact = bodyHash(note.body) === hash;` and then `jaccard(tokens, bodyTokens(note.body))`, so every non-exact active note is lowercased/scanned once for hashing and again for tokenization. On a large memory corpus this doubles per-note string work and allocates a token Set for each candidate comparison.
  Fix: Compute normalized body/tokens once per scanned note, or store a cached body hash/token set in the scan result.

### HonestOracle

- **[important]** `src/memory-write.ts:414` — **Unreadable target becomes absent**
  The loader claims ids are globally unique and that "at most one path resolves", but `const raw = await readFile(path, "utf8").catch(() => undefined);` turns any read failure into absence and continues. A targeted merge/verify can skip an unreadable semantic note and either report the id missing or mutate a procedural duplicate, misleading callers who rely on the global-id guarantee.


---

jcfischer (commented):

6 finding(s): 1 blocker, 4 important, 1 suggestion.


### Security

- **[important]** `src/memory-write.ts:481` — **Assistant trust is caller-selected**
  The note trust is persisted from the caller-selected trigger via `trust: SOMA_MEMORY_TRIGGER_TRUST[options.trigger]`, and `consolidation: "assistant"` makes `--trigger consolidation` enough to create assistant-trust memory. A tool/import caller can bypass the intended `import→quarantined` boundary by choosing the consolidation trigger.
  Fix: Require an internal consolidator capability for `consolidation`, or keep public CLI writes quarantined unless separately authorized.

- **[important]** `src/memory-write.ts:386` — **Symlinks escape memory containment**
  Note writes use `await writeFile(path, content, { encoding: "utf8", flag });` after constructing a path under the memory tree, but Node follows symlinks. A planted `memory/semantic/<id>.md` symlink can make merge/verify writes modify a parseable file outside Soma memory.
  Fix: Reject symlink note paths with `lstat`/`realpath` containment checks, or open with no-follow semantics before writing.

### ContextDrift

- **[important]** `docs/architecture.md:120` — **New Memory taxonomy lacks CONTEXT entry**
  The diff introduces `memory-note subsystem` as a `separate`, schema-governed store with semantic/procedural/episodic paths, but CONTEXT.md:99 defines Memory around work, knowledge, learning, relationship, and state stores. Because this is a new public Memory responsibility boundary, CONTEXT.md should name it explicitly instead of leaving architecture and glossary divergent.
  Fix: Add a CONTEXT.md Memory-note entry, or fold this taxonomy into the existing Memory store terms.

- **[important]** `src/cli/memory.ts:66` — **CLI exposes hook as core term**
  The usage string exposes `[--hook <text>]` for `soma memory write`, but CONTEXT.md:215 says not to call a Soma core concept a `hook` because hooks are substrate-native projection mechanisms. A memory recall phrase is core Memory metadata, so this user-facing option freezes a misleading term.
  Fix: Rename the CLI option to `--recall-trigger` or another Memory-owned term.

### Performance

- **[suggestion]** `src/memory-write.ts:286` — **Dedup scan retains full corpus**
  `const notes: ScannedNote[] = [];` plus `return { notes, unreadable };` means every non-forced create keeps all parsed note bodies in memory before duplicate scoring. Since `parseMemoryNote(content)` includes the markdown body, this is an O(corpus bytes) allocation on the write path.
  Fix: Score each parsed note inside the bounded read worker and return only duplicate candidates plus unreadable paths.

### HonestOracle

- **[blocker]** `src/memory-write.ts:343` — **Merge bypasses provenance refusal**
  The diff claims provenance handling is "closing the smuggling hole where tool/web content rides in under `principal` trust" and errors that "tool/import content cannot ride in under principal trust". `mergeNote` never calls `resolveProvenance`, so a principal-correction merge can accept ignored `provenance: "tool:web"` content while the PR says that source is refused.


---

jcfischer (commented):

6 finding(s): 4 important, 2 suggestion.


### CodeQuality

- **[important]** `src/memory-write.ts:716` — **Verify refreshes superseded notes**
  Verify updates any loaded note with `last_verified: isoDate(now)` and `resurface_count: note.resurface_count + 1` without checking `note.valid_until`. A superseded note can therefore be refreshed after closure, making its decay metadata incorrect.
  Fix: Refuse verify when `note.valid_until !== null`.

- **[important]** `src/memory-write.ts:476` — **Access errors look absent**
  The duplicate-id guard maps every `access(memoryNotePath(somaHome, type, id))` failure through `() => false`. Non-ENOENT failures can be treated as absence, allowing a cross-type duplicate id if one durable directory or path is unreadable.
  Fix: Use `lstat`; continue only on ENOENT and throw/refuse other errors.

### Security

- **[important]** `src/cli/memory.ts:286` — **CLI lets callers self-assert consolidation authority**
  The CLI accepts `case "--consolidation-authority": options.consolidationAuthority = true;`, while the type contract says consolidation authority is `the internal M6 consolidator's capability, not a tier an arbitrary CLI caller can select`. This gives any CLI caller a concrete path to mint assistant-trust memory via `--trigger consolidation --consolidation-authority`, bypassing the documented trust boundary.
  Fix: Do not expose `--consolidation-authority` on the public CLI; accept it only from the internal consolidator call path.

### Architecture

- **[suggestion]** `CONTEXT.md:105` — **Keep schema mechanics out of glossary**
  CONTEXT.md declares itself a "Glossary of the domain language. One canonical term per concept. No implementation details.", but the added glossary entry includes implementation mechanics: "under `memory/{semantic,procedural,episodic}/`, each with bi-temporal frontmatter" and "its sole write path is `soma memory write|verify`".
  Fix: Move path/schema/CLI details to docs/architecture.md and keep CONTEXT.md to the canonical term and boundary.

### Maintainability

- **[suggestion]** `test/memory-write.test.ts:26` — **Split authority-heavy test defaults**
  The shared helper bakes in `principalAuthority: true, // most tests exercise the principal path`, so import/consolidation cases like `trigger: "import"` inherit irrelevant authority state. That makes governance tests harder to scan because each case has to mentally subtract hidden defaults.
  Fix: Use a bare base helper plus `principalCreateOpts` for principal-trust scenarios.

### HonestOracle

- **[important]** `docs/architecture.md:130` — **Only-write-path claim is unsupported**
  The doc states "`soma memory write|verify` (M1) is the only write path", but the implementation adds CLI/API helpers over plain files and does not enforce exclusivity against other code or filesystem writes. Downstream readers could treat this as a governance guarantee when it is only a convention in this diff.


---

jcfischer (commented):

4 finding(s): 4 important; 1 lens(es) failed to run: CodeQuality.


### CodeQuality — DID NOT RUN

  - raw: undefined
  - balanced-largest: undefined
  - trailing: undefined
--- output ---
{"summary":"The core write/verify flow looks generally consistent with the CodeQuality lens after the prior iterations. I found one remaining clarity issue in the public option documentation; no new correctness blocker surfaced from this diff.","findings":[{"path":"src/types.ts","line":2408,"severity":"suggestion","title":"Remove nonexistent CLI flag reference","rationale":"The comment says `CLI flag \`--consolidation-authority\``, but the CLI parser only exposes `--principal-authority` and the help says consolidation is `not available on the public CLI`. This points callers at a flag that does not exist.","suggestion":"Change the comment to describe this as the internal SDK option, not a CLI flag."}]}

### Security

- **[important]** `src/memory-write.ts:608` — **Supersede under-logs required authority**
  The supersede event records only one authority via `...authorityMeta(TRUST_RANK[newNote.trust] >= TRUST_RANK[closed.trust] ? newNote.trust : closed.trust)`, but superseding an assistant note with a principal replacement requires both `consolidationAuthority` for the target mutation and `principalAuthority` for the minted note. This violates the documented logged-escalation invariant because the journal cannot prove all required authority signals were present.
  Fix: Log authority metadata for both `newNote.trust` and `closed.trust`, or include explicit booleans from the validated options.

### Architecture

- **[important]** `docs/architecture.md:122` — **Avoid case-only memory roots**
  The diff says the new note store "lives alongside them under lowercase `memory/`" while the legacy stores are documented as "`MEMORY/*` stores above". That makes the durable schema depend on distinguishing `MEMORY` from `memory`, which is impossible on common case-insensitive filesystems and would force a storage-layout rework before home replication or mixed legacy/note deployments.
  Fix: Use one canonical memory root, e.g. `memory/notes/{semantic,procedural}` or migrate the legacy stores under lowercase `memory/` in the same schema change.

### HonestOracle

- **[important]** `src/types.ts:2391` — **Documented CLI flag does not exist**
  The doc says "CLI flag `--consolidation-authority`", but `src/cli/memory.ts` implements no `--consolidation-authority` case for write or verify and the help text says consolidation is "not available on the public CLI." Downstream users relying on the exported option docs will try a nonexistent flag for assistant-trust authority.


---

jcfischer (commented):

7 finding(s): 3 important, 4 suggestion.


### CodeQuality

- **[suggestion]** `src/cli/memory.ts:335` — **Reject ignored merge ids**
  `case "--id": options.id = readOption(args, index, arg);` records an id, but `if (options.mode !== "merge")` is the only later validation, so `soma memory write --merge old --id new ...` silently mutates `old` while ignoring `new`. For a mutating command, ignored identity flags should be rejected rather than accepted.
  Fix: When `options.mode === "merge"`, reject `--id` and `--type` if present.

- **[suggestion]** `src/cli/memory.ts:210` — **Reject consolidation in CLI**
  `if ((SOMA_MEMORY_WRITE_TRIGGERS as readonly string[]).includes(value))` accepts `consolidation`, but the help text says `The consolidation trigger is an internal (M6) SDK path, not available on the public CLI.` Because the CLI cannot set `consolidationAuthority`, this produces a later SDK-oriented failure instead of a clear CLI parse error.
  Fix: Have the CLI parser reject `--trigger consolidation` with the public-CLI message.

### Security

- **[important]** `src/memory-write.ts:158` — **Parent symlinks escape containment**
  The code builds note paths with `return createPaths(somaHome).resolve("memory", WRITABLE_TYPE_DIRS[type]);` and later writes with `await writeFile(path, content, { encoding: "utf8", flag });`, but only rejects a symlink at the final note path. If `memory/semantic` or `memory/procedural` is a symlink, create/merge/verify can write outside the Soma memory tree despite the documented containment invariant.
  Fix: Reject symlinks in every path component under the memory root, or resolve realpaths and require the final path to remain under the real memory root before writing.

### Maintainability

- **[suggestion]** `src/cli/memory.ts:218` — **Share writable note type guard**
  The CLI repeats the writable-type set with `if (value !== "semantic" && value !== "procedural")` while the writer separately defines `semantic: "semantic", procedural: "procedural"`. Keeping this list in two places raises change cost when later memory-note types move into or out of the write path.
  Fix: Expose a shared writable-note type constant or guard and use it in both CLI parsing and writer routing.

### HonestOracle

- **[important]** `CONTEXT.md:107` — **Atomicity is overstated in glossary**
  CONTEXT calls the note store "schema-governed, atomic," but the architecture text explicitly limits write/event coupling to best-effort and not crash-atomic. Because CONTEXT is repo source-of-truth language, this preserves a transactional guarantee the implementation disclaims.

- **[important]** `src/memory-write.ts:530` — **Global id guarantee is overstated**
  The comment claims "Ids are globally unique across types" and "enforced at create/supersede", but enforcement is a preflight check followed by `await writeNoteFile(path, note, "wx");` on only the chosen type path. Concurrent creates for the same id in different types can both pass the preflight, so the stated global guarantee is wider than the shown mechanism.

- **[suggestion]** `src/cli/memory.ts:63` — **Consolidation CLI availability contradicts itself**
  The help advertises `--trigger <principal-correction|import|consolidation>` but then says "The consolidation trigger is an internal (M6) SDK path, not available on the public CLI." A CLI reader gets mutually exclusive claims about whether that trigger is supported.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-write.ts:685` — **Restore old note on close failure**
  Supersede writes the replacement, then closes the old note with `await writeNoteFile(somaHome, old.path, closed, "w");`, but the catch only runs `await unlink(newPath).catch(() => undefined)`. If the old-note write fails after truncating or partially writing, the old note is not restored from `old.raw`, leaving a mutation without an event.
  Fix: In this catch, restore `old.raw` before rethrowing, and surface restore failure.

### Security

- **[important]** `src/memory-write.ts:441` — **Leaf symlink race remains**
  The write path checks containment, then follows the leaf path: `await assertWriteContained(somaHome, path);` and `await writeFile(path, content, { encoding: "utf8", flag });`. After `loadNoteById` rejects a symlink, an attacker with write access to the memory dir can swap the leaf for a symlink before this write, making merge/verify/supersede overwrite a writable file outside memory.
  Fix: Use fd-based open/write with O_NOFOLLOW for existing-note writes and rollback restores.

### HonestOracle

- **[blocker]** `src/memory-write.ts:687` — **Supersede cleanup failure is hidden**
  The diff claims "One mutating call → one events journal line" but this path uses "await unlink(newPath).catch(() => undefined); // keep the supersede atomic-ish" after the new note has been written and before any event exists. If closing the old note fails and cleanup unlink also fails, the cleanup failure is hidden and a replacement note can remain unjournaled.


---

jcfischer (commented):

2 finding(s): 2 important.


### CodeQuality

- **[important]** `src/memory-write.ts:514` — **Validate loaded note identity**
  `return { path, type, note: parseMemoryNote(raw), raw };` trusts the parsed frontmatter without checking it matches the probed `<id>.md` path. If `semantic/foo.md` contains `id: bar`, `verifyMemoryNote({ id: "foo" })` mutates and journals `bar`, so the command targets a different logical note than requested.
  Fix: After parsing, reject when `note.id !== id` or `note.type !== type`.

### HonestOracle

- **[important]** `src/types.ts:2395` — **SDK authority is self-assertable**
  The comment claims `consolidationAuthority` is "the internal M6 consolidator's capability" and "not a tier an arbitrary CLI caller can select", but the exported `SomaMemoryWriteOptions` field is a public boolean any SDK caller can set. This misleads downstream integrators into treating assistant-trust minting as capability-gated when the diff only gates the CLI flag surface.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/memory-write.ts:683` — **Preserve close failure cause**
  The catch block receives `closeError`, but if undo also fails it throws with only `{ cause: undoError }`: `throw new Error(..., { cause: undoError })`. That loses the original close failure even though both failures are needed to debug the inconsistent state.
  Fix: Include both errors in the thrown error, for example `cause: { closeError, undoError }`.

### HonestOracle

- **[important]** `src/types.ts:2340` — **Public SDK can still self-assert principal**
  The comment claims "a substrate-side caller cannot smuggle `principal` trust", but the public `SomaMemoryWriteOptions` exposes `principalAuthority?: boolean` and callers can set it in-process. That overstates enforcement beyond the shown sudo-style flag.

- **[suggestion]** `src/memory-write.ts:25` — **Logged escalation claim is overstated**
  The file header claims "any tier above `quarantined` needs that tier's explicit, logged authority signal", but `assertTierAuthority` only checks booleans; logging happens later only if the mutation reaches event append. The claim collapses authorization and durable audit into one guarantee.


---

PR #397, opened 2026-07-03
feat(memory): episodic session digests + action log (M5 core)

jcfischer:
## Memory subsystem M5 — episodic capture (core) (plan v2 §M5)

Deterministic capture write paths (no LLM). Episodic notes under `memory/episodic/{sessions,actions}/YYYY-MM/YYYYMMDD-<slug>.md`.

### What it does
- **`soma memory digest --session <id> --body <8–15 lines>`** — writes THE one session digest. One-per-session: a date-independent cross-date scan short-circuits to a duplicate no-op, and the actual write uses an **O_EXCL create** so a concurrent same-date loser observes EEXIST and records a `memory.digest.duplicate` event (with a bounded retry to ride out an in-flight winner's write). Honest limit: no cross-process lock, so a genuine concurrent race across DIFFERENT UTC dates could still produce two digests (M7-audit reconciled). Body validated to 8–15 non-empty lines.
- **`soma memory action --slug <s> --planned-action <t> --approval <proposed|approved|rejected|auto> [--outcome <t>] [--session <id>]`** — logs one plannedAction → approval → outcome entry for the M6 consolidator. Id collision refused via O_EXCL. The session id is recorded in a `**Session:**` body line (not `project`, which stays for real scope). `plannedAction` (not `intent`) because CONTEXT.md reserves `intent` for the principal's Purpose-sourced anchor.
- Session slugs append an 8-hex digest of the full session id — collision-resistant, closing systematic truncation-prefix collisions.
- Episodic notes are **`assistant`** trust, so they don't enter the always-loaded INDEX by admission (M3) until consolidation promotes a pattern.
- Each write appends one journal event with honest write-then-event-rollback (an unlink failure during rollback is surfaced, not swallowed).

### Deferred — the SessionEnd hook
Its ADR-0014 transplant is the sub-agent suppression logic, but recall authored digests via an LLM, which soma forbids. **How the hook obtains the digest text at SessionEnd is a design call for the principal** (agent pre-stages it vs. deterministic event-derived summary), so the hook is not in this PR. The exactly-one write gate holds regardless of whether the CLI or a hook writes.

### Verification
`bunx tsc --noEmit` clean; 14 episodic tests (gate no-op + cross-date dedup + hash collision-guard, 8–15 bound, action structured body / not-yet-outcome / collision refusal / approval validation, CLI parse+run) + full memory suite green.

### DIFF
--- a/src/cli/memory.ts
+++ b/src/cli/memory.ts
@@ -9,6 +9,7 @@ import {
   writeSessionDigest,
 } from "../index";
 import { WRITABLE_NOTE_TYPES, isWritableNoteType } from "../memory-write";
+import { SOMA_MEMORY_ACTION_APPROVALS } from "../types";
 import type {
   SomaMemoryActionApproval,
   SomaMemoryActionOptions,
@@ -131,9 +132,9 @@ export const MEMORY_COMMAND_HELP: { usage: string; subcommands: Record<MemoryAct
       "Usage: soma memory digest --session <id> --body <text> [--substrate <s>] [--home-dir <dir>] [--soma-home <dir>]. " +
       "Write the ONE session digest (8–15 non-empty lines). A second digest for the same session no-ops with an event.",
     action:
-      "Usage: soma memory action --slug <slug> --intent <text> --approval <proposed|approved|rejected|auto> " +
+      "Usage: soma memory action --slug <slug> --planned-action <text> --approval <proposed|approved|rejected|auto> " +
       "[--outcome <text>] [--session <id>] [--substrate <s>] [--home-dir <dir>] [--soma-home <dir>]. " +
-      "Log one intent→approval→outcome action entry (id YYYYMMDD-<slug>; collision refused).",
+      "Log one planned-action→approval→outcome entry (id YYYYMMDD-<slug>; collision refused).",
   },
 };
 
@@ -226,10 +227,10 @@ function parseMemoryDigestArgs(args: string[]): SomaMemoryDigestOptions {
 }
 
 function parseActionApproval(value: string): SomaMemoryActionApproval {
-  if (value === "proposed" || value === "approved" || value === "rejected" || value === "auto") {
-    return value;
+  if ((SOMA_MEMORY_ACTION_APPROVALS as readonly string[]).includes(value)) {
+    return value as SomaMemoryActionApproval;
   }
-  throw new Error("--approval must be one of proposed, approved, rejected, auto.");
+  throw new Error(`--approval must be one of ${SOMA_MEMORY_ACTION_APPROVALS.join(", ")}.`);
 }
 
 function parseMemoryActionArgs(args: string[]): SomaMemoryActionOptions {
@@ -249,8 +250,8 @@ function parseMemoryActionArgs(args: string[]): SomaMemoryActionOptions {
         options.sessionId = readOption(args, index, arg);
         index += 1;
         break;
-      case "--intent":
-        options.intent = readOption(args, index, arg);
+      case "--planned-action":
+        options.plannedAction = readOption(args, index, arg);
         index += 1;
         break;
       case "--approval":
@@ -267,7 +268,7 @@ function parseMemoryActionArgs(args: string[]): SomaMemoryActionOptions {
   }
   const missing: string[] = [];
   if (!options.slug) missing.push("--slug");
-  if (!options.intent) missing.push("--intent");
+  if (!options.plannedAction) missing.push("--planned-action");
   if (!options.approval) missing.push("--approval");
   if (missing.length > 0) {
     throw new Error(`soma memory action is missing required option(s): ${missing.join(", ")}.`);
--- a/src/cli/memory.ts
+++ b/src/cli/memory.ts
@@ -4,7 +4,7 @@ import {
   recallMemory,
   searchSomaMemory,
   verifyMemoryNote,
-  writeAction,
+  writeMemoryAction,
   writeMemoryNote,
   writeSessionDigest,
 } from "../index";
@@ -29,6 +29,7 @@ import type {
   SomaMemoryWriteOptions,
   SomaMemoryWriteResult,
   SomaMemoryWriteTrigger,
+  SubstrateId,
 } from "../types";
 
 /** Parsed `soma memory reindex` — home overrides only; rebuild uses the real clock. */
@@ -167,23 +168,42 @@ export function parseMemoryArgs(args: string[]): ParsedMemoryArgs {
   }
 }
 
+/**
+ * Consume a shared episodic/home option (`--home-dir`, `--soma-home`,
+ * `--substrate`) into `options`. Returns true if `arg` was one of them (the value
+ * was read from args[index+1], so the caller advances the index). One place for the
+ * options both `digest` and `action` share, so they can't drift.
+ */
+function consumeSharedMemoryOption(
+  args: string[],
+  index: number,
+  arg: string,
+  options: { homeDir?: string; somaHome?: string; substrate?: SubstrateId },
+): boolean {
+  switch (arg) {
+    case "--home-dir":
+      options.homeDir = readOption(args, index, arg);
+      return true;
+    case "--soma-home":
+      options.somaHome = readOption(args, index, arg);
+      return true;
+    case "--substrate":
+      options.substrate = parseSubstrate(readOption(args, index, arg));
+      return true;
+    default:
+      return false;
+  }
+}
+
 function parseMemoryDigestArgs(args: string[]): SomaMemoryDigestOptions {
   const options: Partial<SomaMemoryDigestOptions> = {};
   for (let index = 0; index < args.length; index += 1) {
     const arg = args[index];
+    if (consumeSharedMemoryOption(args, index, arg, options)) {
+      index += 1;
+      continue;
+    }
     switch (arg) {
-      case "--home-dir":
-        options.homeDir = readOption(args, index, arg);
-        index += 1;
-        break;
-      case "--soma-home":
-        options.somaHome = readOption(args, index, arg);
-        index += 1;
-        break;
-      case "--substrate":
-        options.substrate = parseSubstrate(readOption(args, index, arg));
-        index += 1;
-        break;
       case "--session":
         options.sessionId = readOption(args, index, arg);
         index += 1;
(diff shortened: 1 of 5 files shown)


jcfischer (commented):

6 finding(s): 2 blocker, 3 important, 1 suggestion.


### CodeQuality

- **[blocker]** `src/memory-episodic.ts:186` — **Digest gate is date-scoped**
  The duplicate key includes the current date in `const id = `${idDate(now)}-${sessionSlug(options.sessionId)}`;`, and the only gate is `if (await pathExists(path)) {`. The same `sessionId` written on a later UTC date will use a different path and create a second digest, breaking the exactly-one-per-session contract.
  Fix: Check for an existing digest by session key across session digest directories, or persist a date-independent session-id gate.

- **[important]** `src/memory-episodic.ts:189` — **Duplicate result returns wrong note**
  On duplicate, `const note = buildEpisodicNote(id, now, options.body, null);` is built from the second call's body, then `return { somaHome, path, created: false, note, event };` returns it even though the on-disk file still contains the first digest. API callers receive a `note` that does not match `path`.
  Fix: Read and parse the existing note for duplicate returns, or remove `note` from duplicate results.

### Architecture

- **[important]** `src/index.ts:464` — **Generic public action API**
  The diff exports `writeSessionDigest, writeAction` from the package root, making `writeAction` a stable public name for a Memory-specific operation. That generic name occupies a broad core API surface even though the exported types are scoped as `SomaMemoryActionOptions` / `SomaMemoryActionResult`.
  Fix: Rename the public export to `writeMemoryAction` before merge.

### Maintainability

- **[suggestion]** `src/cli/memory.ts:166` — **Extract shared episodic option parsing**
  The blocks `case "--home-dir":`, `case "--soma-home":`, and `case "--substrate":` are duplicated in both `parseMemoryDigestArgs` and `parseMemoryActionArgs`. Adding or changing shared memory CLI options now requires editing both parsers in lockstep.
  Fix: Extract a small helper that consumes shared `--home-dir`, `--soma-home`, and `--substrate` options, returning whether it handled the current arg.

### HonestOracle

- **[blocker]** `src/memory-episodic.ts:22` — **Session gate is date-scoped**
  The diff claims “GATED to exactly one per session” and the “one-digest-per-session invariant,” but the implementation keys the duplicate check by `const id = `${idDate(now)}-${sessionSlug(options.sessionId)}`;`. The same session can write another digest on a later UTC date, misleading hook and consolidator readers into relying on an invariant this code does not enforce.

- **[important]** `src/memory-episodic.ts:154` — **Duplicate result returns attempted note**
  The docs claim a duplicate “returns `created: false` with the existing note re-read from disk NOT re-parsed here,” but the returned `note` is built from the second call’s `options.body` before the duplicate branch. API consumers can receive `created:false` with note content that is not the existing on-disk digest.


---

jcfischer (commented):

8 finding(s): 1 blocker, 4 important, 3 suggestion.


### CodeQuality

- **[important]** `src/memory-episodic.ts:74` — **Session slug collisions drop digests**
  The gate key is derived by normalizing and truncating the opaque id via `.replace(/[^a-z0-9]+/g, "-")` and `.slice(0, 40)`, so distinct sessions like `abc` and `abc!` or ids sharing the first 40 slug characters map to one digest. That makes the second valid session hit `Digest already exists for session ${options.sessionId}` and no-op incorrectly.
  Fix: Use a collision-resistant key, e.g. append a short hash of the original sessionId to the slug.

### Architecture

- **[important]** `src/memory-episodic.ts:266` — **Session ID stored as project**
  The diff writes action session scope into the note project field: `const note = buildEpisodicNote(id, now, body, options.sessionId ?? null);`. This conflicts with the architecture contract that session identity belongs to the work/session registries in CONTEXT.md "work registry / session name registry", and makes future project-scoped episodic actions require data migration or field reinterpretation.
  Fix: Add a dedicated episodic `sessionId` field or body line, and leave `project` for actual project scope.

### ContextDrift

- **[important]** `src/cli/memory.ts:134` — **Action API misuses intent**
  The new public CLI exposes `--intent <text>` and describes an `intent→approval→outcome action entry`, while src/types.ts says this is `What the agent intended to do.` CONTEXT.md:760-762 defines `intent` as the principal's checkable directional anchor sourced from Purpose, so this bakes a different meaning into user-facing and exported API vocabulary.
  Fix: Rename the action field/flag to `plannedAction` / `--planned-action` and label the body `Planned action`.

### Performance

- **[suggestion]** `src/memory-episodic.ts:103` — **Duplicate gate scans all digests**
  The duplicate check does `for (const month of months.sort())` and then `entries.sort().find((entry) => idPattern.test(entry))`, so every new session digest can readdir and sort the whole sessions archive before writing. That cost grows with retained digest files and will be paid by the SessionEnd write path once automated.
  Fix: Keep a small session-id index/marker, or store session digests under a stable slug path and separate display date from lookup.

### Maintainability

- **[suggestion]** `src/cli/memory.ts:217` — **Reuse approval constant**
  The CLI duplicates the approval vocabulary with `value === "proposed" || value === "approved" || value === "rejected" || value === "auto"` even though the diff adds `SOMA_MEMORY_ACTION_APPROVALS`. That creates a second list to update when approval states change.
  Fix: Import `SOMA_MEMORY_ACTION_APPROVALS` and validate with that array.

- **[suggestion]** `src/memory-episodic.ts:151` — **Group event write parameters**
  `writeEpisodicNoteWithEvent` takes six positional inputs: `somaHome`, `path`, `note`, `substrate`, `eventFields`, and `now`. The two callsites pass a mixed set of strings, note data, event data, and clock state, so a small options object would be easier to read and extend.
  Fix: Change the helper to accept one object parameter with named fields.

### HonestOracle

- **[blocker]** `src/memory-episodic.ts:22` — **Exactly-one gate is overstated**
  The claim "The write is GATED to exactly one per session" and that the invariant holds "even if the CLI and a hook both fire" is not supported by the check-then-write implementation. Two callers can both pass the preflight scan before either writes; same-date races throw instead of recording the promised duplicate event, and differing date prefixes can produce separate digest paths.

- **[important]** `src/memory-episodic.ts:34` — **Rollback guarantee is stronger than code**
  The comment says "the just-written file is removed so no episodic file is orphaned," but rollback uses `await unlink(path).catch(() => undefined)`. That suppresses cleanup failure while the surrounding text tells readers orphan prevention is guaranteed.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-episodic.ts:242` — **Concurrent duplicate can read partial note**
  The EEXIST path immediately calls `return recordDuplicateDigest(somaHome, path, options.sessionId, options.substrate, now);`, and that handler does `parseMemoryNote(await readFile(existingPath, "utf8"))`. A losing concurrent writer can observe the file after `writeFile(..., { encoding: "utf8", flag: "wx" })` opens/creates it but before the winner has finished writing a parseable note, so the documented no-op path can throw instead.
  Fix: On EEXIST, retry reading/parsing briefly or record the duplicate event without requiring a parse of the in-flight file.

### Architecture

- **[suggestion]** `src/types.ts:2527` — **Bare agent leaks into public docs**
  The exported option docs say "What the agent planned to do", but CONTEXT.md section "agent (always qualified, never bare)" bans bare `agent` in docs and code. This public type comment also sits beside M5 architecture terminology, so it should use Soma's assistant/principal vocabulary precisely.
  Fix: Replace with "What the assistant planned to do."

### HonestOracle

- **[important]** `PR description:7` — **Action command documents wrong flag**
  The description advertises "soma memory action --slug <s> --intent <t>", but the CLI added in the diff accepts "--planned-action" and treats "--intent" as unknown. Anyone following the PR description cannot run the documented command.

- **[suggestion]** `src/memory-episodic.ts:87` — **Hash collision claim is absolute**
  The comment says "two distinct session ids can never map to the same slug", but the slug uses only an "8-hex digest". That is collision-resistant, not collision-proof, so the comment overclaims the gate's guarantee.


---

jcfischer (commented):

5 finding(s): 3 important, 1 suggestion, 1 nit.


### CodeQuality

- **[nit]** `src/types.ts:2487` — **Comments still use old intent term**
  The diff says "Action entries record intent → approval" and later "Approval state of a logged action (design §: intent → approval → outcome)", but the API now exposes `plannedAction`. This makes the public type comments internally inconsistent with the option name.
  Fix: Update those comments to say `plannedAction → approval → outcome`.

### Architecture

- **[important]** `src/types.ts:2487` — **Public docs reuse reserved intent term**
  CONTEXT.md defines `intent` as the principal's Purpose-sourced directional anchor, but the exported Memory API docs still say `Action entries record intent → approval → outcome` and `design §: intent → approval → outcome`. That makes the public action-log surface describe assistant planned work with a reserved Purpose term.
  Fix: Change these comments to `plannedAction → approval → outcome`.

### HonestOracle

- **[important]** `src/memory-episodic.ts:121` — **Duplicate scan fails open**
  The code says `return undefined; // sessions dir absent → no digests yet`, but the catch swallows every `readdir(base)` failure, not only absence. That means permission, ENOTDIR, or filesystem errors are treated as no existing digest, misleading downstream readers relying on the claimed date-independent gate.

- **[important]** `src/memory-episodic.ts:37` — **Index exclusion is unproven**
  The comment claims episodic notes `never enter the always-loaded INDEX by admission (M3) until consolidation promotes`, but this PR only writes `assistant` trust notes and does not show or test the index admission rule. This presents a privacy/loading guarantee as proven by the diff when the enforcement lives elsewhere and is not evidenced here.

- **[suggestion]** `src/types.ts:2490` — **Intent wording drift remains**
  The public type docs say `Action entries record intent → approval → outcome`, while the same PR claims `plannedAction` is used because `intent` is reserved for the principal. That contradiction can mislead consumers about the intended field semantics even though the runtime field is named `plannedAction`.


---

jcfischer (commented):

### Architecture

- **[nit]** `src/memory-episodic.ts:143` — **Bare agent term remains**
  CONTEXT.md's `agent (always qualified, never bare)` rule bans bare `agent` in docs and code, but the diff adds `/** Build an episodic note (assistant trust, agent-authored account). */`. This keeps the assistant/agent layer split fuzzy in the Memory API commentary.
  Fix: Use `assistant-authored account` or a qualified substrate term.

### HonestOracle

- **[important]** `src/types.ts:2488` — **Public API overstates digest uniqueness**
  The type docs claim "The digest write is gated to EXACTLY ONE per session", but the PR description and implementation acknowledge no cross-process lock and a real cross-date race that can create two digests. This misleads API consumers into treating exactly-one as a guaranteed contract rather than a best-effort gate with an audit-reconciled exception.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### CodeQuality

- **[suggestion]** `src/memory-episodic.ts:303` — **Action slug length check is inconsistent**
  `assertSlug(options.slug, "slug")` accepts slugs up to 64 chars, but the later `const id = `${idDate(now)}-${options.slug}`; assertSlug(id, "action id");` rejects any action slug longer than 55 chars because the date prefix counts toward the same 64-char limit. A caller can pass a slug that the slug validator accepts and then get a confusing `action id` failure.
  Fix: Validate action slugs against the maximum id-safe length, e.g. 55 chars, or make the id validator use a separate limit.

### HonestOracle

- **[suggestion]** `src/memory-episodic.ts:119` — **Archival bound is unsupported**
  The comment says "The scan cost grows with retained digests but is bounded by M6's 90-day archival", but this PR neither implements nor cites that archival behavior. A reader relying on this would believe the scan has an enforced retention bound when the diff only shows an unbounded directory walk.


---

PR #398, opened 2026-07-03
feat(memory): deterministic consolidation with --dry-run (M6)

jcfischer:
M6 deterministic consolidation (no LLM). Dry-run reports the same file OPERATIONS the real run applies (operation-level parity, not a byte diff) — planned ops on dry-run, applied on a real run. Ops: prune aged episodic (sessions 90d / actions 180d, by created; source dir must match created month) -> regenerate monthly digest from the archive (recoverable) + move to memory/archive/ mirroring the relative path (ids slug-validated, all destinations + parent chains preflighted for symlink-escape and no-clobber before any move); mark aged-unverified semantic review:stale (lstat-guarded write, never through a symlink, never archived); list top-50 high-lexical-similarity note pairs (candidate contradictions, bounded, no merge); GC current-work-*.json >7d ONLY under explicit --gc-state (lstat, honest unlink); rebuild INDEX only when mutated + append memory.consolidate event (records a completed pass). Unreadable notes surfaced (result/CLI/event), never silently skipped. Idempotent on mutations; the only deletion is state GC. Governed maintenance path documented in architecture.md. Verification (local): bunx tsc --noEmit clean; bun test test/memory-consolidate.test.ts -> 12 pass / 0 fail.


jcfischer (commented):

7 finding(s): 1 blocker, 4 important, 2 suggestion.


### CodeQuality

- **[blocker]** `src/memory-consolidate.ts:197` — **State GC ignores homeDir roots**
  `consolidateMemory` resolves `somaHome` with `const somaHome = createPaths(options).root();`, but then `planStateGc` calls `const paths = createPaths(somaHome);`. Passing the resolved Soma root as a home-dir argument can make GC scan the wrong tree, so `--home-dir <dir>` may not delete the planned `current-work-*.json` files from the actual Soma home.
  Fix: Pass options through to `planStateGc`, or have `planStateGc` accept an already-created paths object instead of calling `createPaths(somaHome)`.

### Security

- **[important]** `src/memory-consolidate.ts:107` — **Validate note ids before archive paths**
  The archive target is built from parsed frontmatter with `paths.resolve("archive", "episodic", kind, month, `${parsed.id}.md`)`, then applied via `await rename(from, to)`. A crafted episodic note id containing path separators or `..` can move the file outside the intended `archive/episodic/...` location during consolidation.
  Fix: Reject or sanitize ids before path construction, and assert the resolved target stays under the archive root.

### Architecture

- **[important]** `src/memory-consolidate.ts:247` — **Bypasses memory-note event coupling**
  docs/architecture.md says `soma memory write|verify` is the only governed write path and that "each mutation appends one event to the existing memory/STATE/events.jsonl stream"; this pass mutates notes directly via `await writeFile(path, serializeMemoryNote({ ...note, review: "stale" }), "utf8");` with no event append. The same module also moves note files with `await rename(from, to);`, creating a second governed mutation path outside the documented contract.
  Fix: Route consolidation mutations through a shared governed memory-note mutation helper, or update the architecture contract and emit consolidation events.

### Performance

- **[suggestion]** `src/memory-consolidate.ts:144` — **Contradiction scan is quadratic**
  The nested scan `for (let i = 0; i < active.length; i += 1)` / `for (let j = i + 1; j < active.length; j += 1)` computes Jaccard for every active-note pair, so consolidation latency grows as O(n^2) with durable note count.
  Fix: Bucket by rare tokens or add a candidate prefilter before Jaccard scoring.

### Maintainability

- **[suggestion]** `src/memory-consolidate.ts:207` — **Split mutation application helpers**
  `export async function consolidateMemory` includes each mutation block inline: `// 1. Prune episodic`, `// 2. Mark stale`, `// 3. State GC`, and `// 4. Rebuild INDEX`. That makes the orchestrator carry digest writing, archiving, stale marking, GC, and indexing state in one body.
  Fix: Extract `applyEpisodicArchive`, `applyStaleMarks`, and `applyStateGc` helpers, leaving `consolidateMemory` as plan-build-apply orchestration.

### HonestOracle

- **[important]** `src/memory-consolidate.ts:20` — **Dry-run parity overclaims exact diff**
  The doc claims "dry-run output == real-run diff", but the result only records digest file paths and counts, while the real run appends digest contents and rebuilds INDEX. A downstream operator cannot reconstruct the real filesystem diff from this dry-run output.

- **[important]** `src/memory-consolidate.ts:25` — **Archive path preservation is overstated**
  The doc claims archived notes are moved to `archive/`, "preserving their relative path", but the target is built as `archive/episodic/${kind}/${month}/${parsed.id}.md`. If the source filename differs from the note id, the archive path is rewritten, so the stated audit invariant is false.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:246` — **Digest can be written before archive succeeds**
  The code appends digest pointers with `await appendFile(digestPath, `${header}${lines}\n`, "utf8");` before any corresponding `await rename(e.from, e.to);`. If a later rename fails, the digest records notes as consolidated even though the source files remain active, and a retry appends duplicate digest entries.
  Fix: Move each note before recording it in the digest, or write digest changes to a temp file after successful renames.

### Architecture

- **[important]** `src/memory-consolidate.ts:107` — **Archive escapes memory root**
  docs/architecture.md §Memory defines Memory as living under a single lowercase `memory/` root, but the diff maps archived notes with `return paths.resolve("archive", ...relSegments);`. That makes `archive/memory/episodic/...` a second memory-note root that future Memory tooling must special-case.
  Fix: Archive under `memory/archive/...` or update the architecture contract before exposing this layout.

### Performance

- **[suggestion]** `src/memory-consolidate.ts:136` — **Parallelize episodic note reads**
  `for (const path of files) { const parsed = await readFile(path, "utf8")... }` reads every candidate note sequentially, so consolidation latency becomes the sum of all per-file I/O across session/action archives. These reads are independent and only feed plan construction.
  Fix: Read candidate files with bounded parallelism before filtering by parsed metadata.

### Maintainability

- **[suggestion]** `src/memory-consolidate.ts:192` — **Drop redundant pair tracking**
  `const seen = new Set<string>();` plus `if (seen.has(key)) continue; seen.add(key);` is unnecessary because candidates are already unique per `i` and the loop only accepts `j > i`. This adds an extra branch and state to the densest function without preventing any reachable duplicate.
  Fix: Remove `seen` and the `key` check from `planContradictions`.

### HonestOracle

- **[important]** `src/types.ts:2551` — **Dry-run diff parity is false**
  The comment claims "dry-run output == real-run diff", but the result only reports operation lists and paths, not digest bytes, INDEX bytes, journal entries, or exact filesystem diffs. Downstream callers relying on this type contract would believe dry-run exposes the full real-run diff when it does not.


---

jcfischer (commented):

5 finding(s): 1 blocker, 2 important, 2 suggestion.


### CodeQuality

- **[important]** `src/memory-consolidate.ts:247` — **Archive can lose digest pointer**
  The sequence `await rename(e.from, e.to); // move first` followed later by `await appendFile(e.digestPath, ...` means an append failure after the rename archives the note without recording it in the digest. A retry will not re-plan that note because it has already moved out of the active episodic tree.
  Fix: Write a recoverable pending record or append the digest pointer using an operation that can be retried after archive, including archived notes missing from digests in planning.

### Security

- **[blocker]** `src/memory-consolidate.ts:77` — **Reject symlinked episodic directories**
  The diff reads arbitrary month entries with `entries = await readdir(join(base, month));` and later moves files derived from that path with `await rename(e.from, e.to);`. A writable symlink such as `memory/episodic/sessions/2026-03 -> /tmp/victim` lets consolidation parse and move matching outside files into the archive path, crossing the memory root trust boundary.
  Fix: Use lstat/realpath containment checks before descending and before rename, and skip or fail on symlinks/non-directories.

### Architecture

- **[important]** `src/cli/memory.ts:148` — **Consolidate expands governed write surface**
  The diff adds the public command "consolidate" and implements note mutations such as "await writeFile(path, serializeMemoryNote({ ...note, review: \"stale\" }), \"utf8\")" outside `soma memory write|verify`. This conflicts with docs/architecture.md, Memory section: `soma memory write|verify` is documented as the only governed write path for memory notes.
  Fix: Update the Memory architecture contract to include consolidation as a governed maintenance write path, or route these mutations through the existing governed write APIs.

### Maintainability

- **[suggestion]** `src/types.ts:2548` — **Deduplicate consolidation contract prose**
  The type-level comment repeats the module contract: "dry-run plan equals the real run's plan" and "It enumerates the operations, NOT a byte-level diff" while `src/memory-consolidate.ts` also says "the dry-run plan equals the real run's plan" and "It does NOT reproduce byte-level digest/INDEX content". Keeping the same nuanced contract in two long comments raises drift cost when consolidation behavior changes.
  Fix: Keep the detailed contract in one place and shorten the other to reference it.

### HonestOracle

- **[suggestion]** `src/types.ts:2591` — **Global deletion claim is unsupported**
  The exported result contract says `stateGced` is "the only deletion", and the M6 comment above says `GC old current-work-*.json state (the ONE true deletion)`. This diff proves only that this consolidation path deletes state files; it does not prove that no other memory-subsystem path deletes files, so downstream API readers get a stronger guarantee than the PR establishes.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:358` — **Event failure leaves mutations applied**
  The real run mutates files before `await appendSomaMemoryEvent(somaHome, {`, so if the event append throws, archived notes, stale marks, state GC, and INDEX rebuild remain applied while the caller sees a failed governed run. This contradicts the new architecture text saying `an event-append *failure* rolls the file mutation back`.
  Fix: Either implement rollback around consolidation mutations or update the documented guarantee for consolidate.

### Architecture

- **[important]** `src/memory-consolidate.ts:312` — **State GC needs explicit override**
  The diff deletes protected Memory state by default with `await unlink(join(root, rel)).catch(() => undefined);`, while CONTEXT.md section `private, protected, generated (data adjectives)` says protected data `Cannot be destructively mutated (delete, move, overwrite) without explicit override.` The CLI only adds `soma memory consolidate [--dry-run]`, so the destructive path has no explicit override surface.
  Fix: Require an explicit destructive flag, e.g. `--gc-state`, before deleting `current-work-*.json`.

### Performance

- **[important]** `src/memory-consolidate.ts:259` — **Digest rebuild scans full archive**
  `for (const kind of ["sessions", "actions"] as const) { const files = await listEpisodicNotes(paths.resolve("memory", "archive", "episodic", kind)); for (const path of files) { const note = await readFile(path, "utf8")...` rereads every archived episodic note for any affected month. Because archive only grows, a single newly archived note makes consolidation do O(total archived notes) sequential file reads.
  Fix: Restrict archive reads to the affected `months`, and use bounded concurrent reads as in `planEpisodicArchive`.

### Maintainability

- **[suggestion]** `src/cli/memory.ts:184` — **Reuse shared home option parsing**
  `parseMemoryConsolidateArgs` repeats the same home option branches already covered by the nearby shared parser: `case "--home-dir":`, `case "--soma-home":`, and `options.homeDir = readOption(args, index, arg);`. Keeping another hand-rolled switch makes future shared option changes easier to miss.
  Fix: Route `--home-dir`/`--soma-home` through the existing shared memory option helper and keep only `--dry-run` local.

### HonestOracle

- **[blocker]** `docs/architecture.md:144` — **Rollback guarantee is false**
  The diff states, "an event-append *failure* rolls the file mutation back," but `consolidateMemory` applies archive/stale/GC/index mutations before `appendSomaMemoryEvent` and contains no rollback path. Downstream readers will believe governed consolidation is rollback-coupled when event failure can leave mutations without the claimed event.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:235` — **State GC follows symlinks**
  `const info = await stat(full).catch(() => undefined);` follows symlinks, and later `await unlink(join(root, rel)).catch(() => undefined);` will remove a matching `current-work-*.json` symlink if its target is old. That is inconsistent with the consolidation trust-boundary handling and can delete a protected state entry that is not a real file.
  Fix: Use `lstat` in `planStateGc` and require `info.isFile()` and `!info.isSymbolicLink()` before adding the path.

### Performance

- **[suggestion]** `src/memory-consolidate.ts:369` — **Skip no-op index rebuilds**
  A real run with no planned mutations still executes `await rebuildMemoryIndex({ somaHome, now });`, forcing a full index rebuild even when `episodic`, `staleMarks`, and `stateGc` are all empty. This makes idempotent maintenance runs do corpus-scale work without changing consolidation output.
  Fix: Only rebuild when a planned mutation exists, or add an explicit reindex flag for no-op refreshes.

### Maintainability

- **[suggestion]** `src/memory-consolidate.ts:77` — **Deduplicate markdown file listing**
  `listEpisodicNotes` repeats the same guarded directory/file scan shape as `listArchivedMonthNotes`: `if (!(await isRealEntry(base, "dir"))) return [];` / `if (await isRealEntry(filePath, "file")) files.push(filePath);`, and later `if (!(await isRealEntry(dir, "dir"))) return [];` / `if (await isRealEntry(filePath, "file")) out.push(filePath);`. Keeping two copies raises future-change cost for the symlink trust-boundary behavior.
  Fix: Extract a small `listRealMarkdownFiles(dir)` helper and reuse it in both functions.

### HonestOracle

- **[important]** `docs/architecture.md:145` — **Event emission is overstated**
  The doc claims consolidate is "event-emitting", but the implementation only appends an event inside `if (episodic.length > 0 || staleMarks.length > 0 || stateGc.length > 0)`. A real pass that only rebuilds INDEX or lists contradictions is therefore not event-emitting, which misleads downstream audit readers.

- **[important]** `src/memory-consolidate.ts:32` — **Similarity is sold as contradiction**
  The module claims "List contradictions — active durable notes with Jaccard ≥ 0.6", but Jaccard overlap only proves lexical similarity, not contradiction. The result name `contradictions` makes a surrogate endpoint look like a semantic finding.


---

jcfischer (commented):

3 finding(s): 2 important, 1 suggestion; 1 lens(es) failed to run: CodeQuality.


### CodeQuality — DID NOT RUN

  - raw: undefined
  - balanced-largest: undefined
  - trailing: undefined
--- output ---
{"summary":"The consolidation implementation is mostly careful about dry-run parity, symlink handling, and idempotent digest regeneration. I found one correctness issue in the CLI output: it reports an index rebuild even when the implementation intentionally skips it on no-op runs.","findings":[{"path":"src/cli/memory.ts","line":708,"severity":"suggestion","title":"No-op runs misreport index rebuild","rationale":"The formatter always prints `result.dryRun ? \`index: would rebuild ${result.indexPath}\` : \`index: rebuilt ${result.indexPath}\``, but `consolidateMemory` only rebuilds when `mutated` is true. A real no-op run therefore tells users the index was rebuilt when no rebuild happened.","suggestion":"Include an `indexRebuilt`/`mutated` flag in the result, or print a neutral path-only line when no mutation occurred."}]}

### HonestOracle

- **[important]** `src/cli/memory.ts:706` — **Index rebuild output is false**
  The CLI reports `index: rebuilt ${result.indexPath}` for every non-dry-run, but `consolidateMemory` only rebuilds when `mutated` is true. A no-op real run therefore tells downstream users an INDEX rebuild happened when it was deliberately skipped.

- **[suggestion]** `src/memory-consolidate.ts:54` — **Idempotency overstates empty plan**
  The doc claims `an empty plan, unchanged INDEX`, but `similarPairs` are recomputed and returned on every run because listing does not resolve them. A second run can still have a non-empty reported plan even when no file mutations remain.


---

jcfischer (commented):

### Maintainability

- **[suggestion]** `src/cli/memory.ts:696` — **Untangle nested formatter ternary**
  The index status is encoded as a nested ternary: "const indexLine = result.dryRun\n    ? mutated\n      ? `index: would rebuild ${result.indexPath}`". This packs four output cases into one expression, making future formatter changes more error-prone than a small helper or if/else block.
  Fix: Extract `formatConsolidateIndexLine(result, mutated)` or replace the nested ternary with explicit branches.

### HonestOracle

- **[important]** `docs/architecture.md:139` — **Archive invalidation is not implemented**
  The docs claim consolidation "archives aged episodic notes (move under `memory/archive/`, invalidate-never-delete)", but the apply path only `rename(e.from, e.to)` and never writes an invalidating field such as `valid_until`. Downstream code or auditors relying on an explicit invalidation marker will be misled; the diff only proves relocation.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:317` — **Archive rename can overwrite target**
  `await rename(e.from, e.to);` will replace an existing archive file on POSIX-like filesystems. A stale or colliding archive target can lose the prior raw note while the run reports a normal archive operation.
  Fix: Refuse an existing archive target before moving, or use a no-overwrite move strategy.

- **[important]** `src/memory-consolidate.ts:333` — **State GC hides unlink failures**
  `await unlink(join(root, rel)).catch(() => undefined);` suppresses all deletion errors, including permission and I/O failures. The returned `stateGced` list then reports files as GC'd even when they remain on disk.
  Fix: Ignore only ENOENT; rethrow other unlink failures.

### HonestOracle

- **[important]** `src/memory-consolidate.ts:333` — **GC failures are hidden**
  The destructive path is reported as GC via `stateGced`, but deletion uses `await unlink(join(root, rel)).catch(() => undefined);`. If unlink fails, the pass can still report and event-record a deletion that did not happen, misleading anyone relying on consolidation output.

- **[suggestion]** `PR description:1` — **Verification claim is unsupported**
  The description claims `8 tests; tsc clean.`, but the supplied PR data includes no `tsc` output or CI evidence. That makes the typecheck status an unsupported verification claim in the review artifact.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:317` — **Symlink archive target is clobbered**
  The guard only rejects `if (await isRealEntry(e.to, "file"))`, but `isRealEntry` returns false for symlinks, so `await rename(e.from, e.to);` will replace an existing symlink at the archive target. That violates the stated no-clobber behavior in the adjacent diff text: `Never clobber an existing archive tombstone`.
  Fix: Use `lstat(e.to)` and reject any existing target except ENOENT before `rename`.

### Security

- **[blocker]** `src/memory-consolidate.ts:317` — **Archive parents can symlink outside root**
  The archive apply path creates and writes through the destination parent without verifying every archive parent is a real directory: "await mkdir(dirname(e.to), { recursive: true });" followed by "await rename(e.from, e.to);". If an attacker pre-creates `memory/archive/episodic/...` as a symlink, an aged note can be moved outside the Soma memory root despite the intended root containment.
  Fix: Before `rename`, lstat each destination parent segment under `memory/archive` and reject symlinks/non-directories.

### HonestOracle

- **[suggestion]** `src/types.ts:2600` — **Index path overclaims rebuild**
  The contract says "INDEX.md path rebuilt (real run) or that would be rebuilt (dry-run)", but `consolidateMemory` returns `indexPath` even when `mutated` is false and no rebuild is performed or planned. API consumers reading this type are told the field proves an index operation that may not exist.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:316` — **Validate all archive destinations before moving**
  `await assertSafeArchiveDest(memoryRoot, e.to);` is inside the same loop as `await rename(e.from, e.to);`, so if a later archive target is unsafe or already exists, earlier notes have already been moved and the function throws before digest regeneration, index rebuild, or event append. That leaves a partially applied consolidation despite the failed command.
  Fix: Preflight every `episodic` destination first, then perform the renames and digest regeneration only after all archive targets are known safe.

### Performance

- **[suggestion]** `src/memory-consolidate.ts:227` — **Cap similar-pair output**
  The planner stores every match via `const pairs: SomaMemorySimilarPair[] = [];` and `pairs.push({ a, b, score });`; with many near-identical active notes, this allocates n(n-1)/2 results and the CLI later renders them all. That is a concrete unbounded memory/output cost for duplicated corpora.
  Fix: Keep only the top N pairs or add a configurable report limit.

### HonestOracle

- **[important]** `src/memory-consolidate.ts:22` — **Dry-run parity skips refusal checks**
  The claim "`--dry-run` reports the SAME set of file operations ... that the real run applies" is false when an archive target already exists or an archive parent is unsafe: dry-run returns before `assertSafeArchiveDest`, while the real run can throw before applying the reported move. That misleads operators using dry-run as evidence of the real mutation set.

- **[important]** `docs/architecture.md:149` — **Event guarantee ignores partial failures**
  The claim "When it mutates anything it appends one `memory.consolidate` event" is wider than the implementation: archive, stale-mark, or GC helpers can mutate earlier files and then throw before the event append. Audit consumers are misled into treating an absent event as evidence that no consolidation mutation happened.

- **[suggestion]** `src/memory-consolidate.ts:41` — **GC contract omits required flag**
  The module contract says "`current-work-*.json` files older than 7d are DELETED" without the `--gc-state` condition, while the implementation only plans GC when `options.gcState === true`. Readers of the advertised single-source contract get the default behavior wrong.


---

jcfischer (commented):

4 finding(s): 4 important.


### CodeQuality

- **[important]** `src/memory-consolidate.ts:200` — **Digest can miss moved notes**
  The digest month comes from `const month = parsed.created.slice(0, 7);`, while regeneration scans only `listArchivedMonthNotes(paths, "sessions", month)` and `listArchivedMonthNotes(paths, "actions", month)`. If a valid episodic note lives in a mismatched month directory, the archive move mirrors the source path but the digest scan looks in the created-month archive directory, so the archived note is omitted from its digest.
  Fix: Validate source directory month against `created`, or regenerate digests from the planned archive paths.

- **[important]** `src/memory-consolidate.ts:397` — **Unreadable durable notes are hidden**
  The scan result is assigned with `const durable = await collectDurableNotes(somaHome);`, but only `durable.notes` feeds `planStaleMarks` and `planSimilarPairs`. Any `durable.unreadable` entries are silently skipped, so the command can mutate and append a success event while known durable notes were not checked.
  Fix: Surface `durable.unreadable` in the result/event, or refuse to mutate when it is non-empty.

### Security

- **[important]** `src/memory-consolidate.ts:356` — **Stale marks can follow symlinks**
  The new mutation writes each collected durable path with `await writeFile(path, serializeMemoryNote({ ...note, review: "stale" }), "utf8");` after planning from `const durable = await collectDurableNotes(somaHome);`. Because this path lacks the lstat/non-symlink guard used for episodic archival and state GC, a symlinked `memory/semantic/*.md` can point outside Soma and be overwritten when it parses as an aged semantic note.
  Fix: Before stale planning or writing, lstat durable note paths and skip or reject symlinks/non-regular files.

### HonestOracle

- **[important]** `src/types.ts:2595` — **Dry-run fields claim real writes**
  The result contract says `Monthly digest files regenerated from the archive`, `Semantic notes newly marked`, and `files GC'd`, but `consolidateMemory` returns those same fields on dry-run before any mutation. This misleads consumers into reading planned dry-run operations as completed filesystem changes.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:333` — **Digest drops unreadable archive notes**
  The digest scan uses `readFile(path, "utf8").then(parseMemoryNote).catch(() => undefined)` and then filters with `n !== undefined`, so unreadable or malformed archived notes are silently omitted from the regenerated monthly digest. That produces a successful run with an incomplete recovery artifact.
  Fix: Surface these archive read/parse failures in `unreadable` or throw before rewriting the digest.

### Performance

- **[important]** `src/memory-consolidate.ts:226` — **Similarity cap still allocates all pairs**
  The diff still appends every matching pair via `pairs.push({ a, b, score });`, then sorts all matches with `pairs.sort(...)` before `return pairs.slice(0, MAX_SIMILAR_PAIRS);`. On a duplicated corpus this still materializes and sorts O(n^2) pairs, so the cap does not bound memory or latency.
  Fix: Keep a bounded top-50 heap/list while scoring instead of collecting all pairs.

### HonestOracle

- **[suggestion]** `PR description:1` — **Unverifiable review-round count**
  The description claims "8 sage rounds" but provides no linked run IDs, logs, or reviewer outputs, so downstream reviewers cannot distinguish actual review coverage from process theater.


---

jcfischer (commented):

3 finding(s): 3 important.


### Security

- **[important]** `src/memory-consolidate.ts:361` — **Digest write follows symlinked parents**
  The digest path is written with `await writeFile(digestPath, ...` after only `await mkdir(dirname(digestPath), { recursive: true });`, so a preexisting `memory/episodic/digests` symlink can redirect the governed consolidation write outside the Soma memory root. This violates the PR's symlink-escape invariant for maintenance destinations.
  Fix: Before writing digests, lstat each existing parent and refuse symlinks/non-directories; write only inside a verified real directory.

### HonestOracle

- **[important]** `PR description:1` — **Unreadable event surface is overstated**
  The description claims "Unreadable notes surfaced (result/CLI/event)", but the event only records `unreadable: unreadable.length`, not the unreadable note paths. That misleads downstream audit consumers into thinking the journal identifies the affected files.

- **[important]** `docs/architecture.md:155` — **M7 audit claim is unsupported**
  The doc says "the M7 audit is the ground-truth check", but this PR shows no M7 audit implementation or cited audit contract. That makes the recovery story depend on an unproven mechanism.


---

jcfischer (commented):

### Maintainability

- **[suggestion]** `src/memory-consolidate.ts:177` — **Extract note parsing helper**
  The same bounded parse block appears in two places: `note: await readFile(path, "utf8").then(parseMemoryNote).catch(() => undefined),` with the same `{ path, note }` shape and concurrency value. Duplicating this read/parse/error convention makes future unreadable-note behavior harder to change consistently.
  Fix: Extract a small `parseNotesBounded(paths: string[])` helper and reuse it in episodic planning and digest regeneration.

- **[suggestion]** `src/cli/memory.ts:705` — **Centralize mutation predicate**
  The CLI re-derives the core mutation rule with `result.archived.length > 0 || result.markedStale.length > 0 || result.stateGced.length > 0`, while the implementation uses the same concept to decide whether to rebuild and emit events. A future mutation category would require remembering to update both places to keep output truthful.
  Fix: Return `mutated` on `SomaMemoryConsolidateResult` or share a helper for this predicate.

### HonestOracle

- **[important]** `src/memory-consolidate.ts:375` — **Digest preflight claim is false**
  The comment says "Destinations were all preflighted in the plan phase (both dry-run and real), so no per-note refusal can leave an earlier note already moved," but only archive targets are preflighted before apply; `assertRealParentChain(paths.memory(), digestPath)` runs later inside `regenerateMonthlyDigests` after `rename(e.from, e.to)`. A symlinked digest parent can therefore fail after notes were already moved, so the stated partial-application guarantee is misleading.


---

jcfischer (commented):

### Architecture

- **[suggestion]** `src/memory-consolidate.ts:43` — **Do not exclude state from Memory**
  The diff says "state is not memory; notes are only archived", but docs/architecture.md’s Memory section places `memory/STATE/` inside the single Memory root and CONTEXT.md defines state stores within the Memory compartment. That wording creates a false boundary between notes and Memory state in the new consolidation contract.
  Fix: Say state GC is the only destructive mutation of protected state; do not say state is outside Memory.

### HonestOracle

- **[important]** `docs/architecture.md:143` — **Completion event guarantee overstates implementation**
  The doc claims, "A pass that RUNS TO COMPLETION with any mutation appends one `memory.consolidate` event," but `consolidateMemory` applies archive/stale/GC/index mutations before awaiting `appendSomaMemoryEvent`, so an event append failure prevents completion after mutations are already applied. This makes the documented completion boundary misleading for anyone using the event stream as evidence of maintenance success.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:79` — **Malformed dates can crash consolidation**
  `dateMs` accepts unchecked parsed numbers: `const [y, m, d] = isoDate.split("-").map(Number); return Date.UTC(y, m - 1, d);`. If a parsed note has an invalid `created` or `last_verified`, `ageDays` can return `NaN`, and later `note.created.slice(0, 7)` / month matching may treat bad metadata as valid instead of surfacing the file via `unreadable`.
  Fix: Validate `YYYY-MM-DD` before age/month use and classify invalid-date notes as unreadable/refused.

### Architecture

- **[suggestion]** `src/memory-consolidate.ts:333` — **Core API writes to console**
  docs/architecture.md defines the core as "filesystem-native and substrate-neutral," but the exported consolidation library function directly emits presentation output: `console.warn(`soma: archived note ${entry.path} is unreadable — omitted from the ${month} digest (surface for the audit).`);`. That couples a core Memory operation to a specific process output channel instead of returning the unreadable archive note through the result/event surface for adapters or CLI formatting.
  Fix: Return archived digest omissions in `SomaMemoryConsolidateResult` and let `src/cli/memory.ts` format them.

### HonestOracle

- **[important]** `src/cli/memory.ts:151` — **CLI sells similarity as contradiction**
  The help text says `list similar-note pairs (candidate contradictions)`, while the implementation only computes Jaccard token overlap and later prints `candidate duplicates/contradictions`. A downstream operator could treat lexical near-duplicates as contradiction evidence even though the diff shows no semantic contradiction check.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:216` — **Similar-pair cap can drop top pairs**
  The bounded collector truncates with `pairs.length = MAX_SIMILAR_PAIRS` during traversal, so later comparisons are only ranked against the already-truncated set after earlier low-score entries may have been kept and then more high-score entries appended. This can return the wrong top-50 despite the comment claiming `the final result is the deterministic top-N by score`.
  Fix: When flushing, sort and keep the current top N plus compare all later candidates against the current cutoff before appending, or use a fixed-size min-heap keyed by score/id.

### ContextDrift

- **[suggestion]** `src/types.ts:2581` — **Use principal, not human**
  The exported API comment introduces "for human review", but CONTEXT.md:362 says formal docs/code should not use `human` as a principal synonym: "principal is the only word." As public type documentation, this should keep the principal/assistant vocabulary aligned.
  Fix: Change to "for principal review".

### HonestOracle

- **[important]** `docs/architecture.md:153` — **Rerun reconciliation is overstated**
  The doc claims "a re-run reconciles" after "the event append itself fails," but the implementation only appends an event when `mutated` is true; after mutations already applied, the next run can be a no-op and write no event. This misleads operators into treating rerun as repair for a missing journal record.

- **[suggestion]** `src/memory-consolidate.ts:392` — **Preflight claim hides partial moves**
  The comment claims "Destinations were all preflighted... so no per-note refusal can leave an earlier note already moved," but the loop still runs `mkdir` and `rename` after earlier moves. A later filesystem failure can still leave the archive step partially applied, so the comment overstates the protection shown by the code.


---

jcfischer (commented):

4 finding(s): 2 important, 1 suggestion, 1 nit.


### CodeQuality

- **[important]** `src/cli/memory.ts:190` — **Shared boolean flags skip next option**
  The parser always advances an extra slot after `consumeSharedMemoryOption(args, index, arg, options)`, via `index += 1;`, even if the consumed shared option is a boolean flag such as `--home-dir` is not the only possible shape. If `consumeSharedMemoryOption` handles any valueless shared flag, the next real option, for example `--gc-state`, is skipped without validation or effect.
  Fix: Only increment for shared options that consume a following value, or have `consumeSharedMemoryOption` return the number of consumed argv entries.

### Security

- **[important]** `src/memory-consolidate.ts:356` — **Stale mark write can race symlinks**
  The plan checks `if (!(await isRealEntry(path, "file"))) continue;`, but the later mutation uses `await writeFile(path, serializeMemoryNote({ ...note, review: "stale" }), "utf8");`, which follows symlinks if the file is swapped after planning. A local attacker with write access to the memory tree can turn a governed stale-mark write into an overwrite outside the Soma root.
  Fix: Re-lstat immediately before writing and write via a no-follow/temp-file+rename pattern that refuses symlinks.

### Architecture

- **[nit]** `src/memory-consolidate.ts:33` — **Use principal vocabulary**
  CONTEXT.md §principal says docs and src should use `principal`, not `human`, in formal Soma vocabulary. The new comment says `NEVER auto-archived — a human reviews.`
  Fix: Change to `NEVER auto-archived — the principal reviews.`

### Maintainability

- **[suggestion]** `src/memory-consolidate.ts:465` — **Avoid result alias mutation**
  The result captures `unreadable,` before apply-time omissions are known, then later relies on `for (const p of archivedOmissions) if (!unreadable.includes(p)) unreadable.push(p);` mutating the same array reference. That hidden alias makes future changes to result construction easy to break.
  Fix: Build the result after `archivedOmissions` is merged, or assign `result.unreadable = mergedUnreadable` explicitly.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:72` — **Reject nonexistent calendar dates**
  The guard only checks shape via `return /^\d{4}-\d{2}-\d{2}$/.test(s);`, while `dateMs` then normalizes invalid dates such as `2026-02-31` through `Date.UTC(y, m - 1, d)`. That can make malformed notes eligible for archive/stale marking instead of being surfaced as unreadable.
  Fix: Validate by round-tripping: construct the UTC date and require `toISOString().slice(0, 10) === isoDate`.

### Maintainability

- **[suggestion]** `src/memory-consolidate.ts:69` — **Rename contradiction threshold constant**
  The diff now frames this as lexical similarity, but the constant still says `const CONTRADICTION_JACCARD = 0.6;` and later checks `if (score >= CONTRADICTION_JACCARD)`. That keeps obsolete contradiction vocabulary in the core path and makes future changes harder to search and reason about.
  Fix: Rename to `SIMILARITY_JACCARD_THRESHOLD` or `NEAR_DUPLICATE_JACCARD`.

### HonestOracle

- **[important]** `src/memory-consolidate.ts:392` — **Preflight coverage is overstated**
  The comment says "Preflight EVERY mutation destination now" but the loop only checks archive targets and the digest parent; stale-mark temp writes and state-GC paths are not included in that preflight set. This misleads readers into believing all write/delete destinations are refused before mutation begins.

- **[important]** `src/memory-consolidate.ts:475` — **Applied result can overstate stale marks**
  The result reports `markedStale` from the plan, while `applyStaleMarks` can `continue` after a swap: "if (info === undefined || info.isSymbolicLink() || !info.isFile()) continue". A real run can therefore return and event a stale mark that was skipped, so "planned on dry-run, marked on a real run" is false under the implemented race path.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:478` — **Skipped stale marks still trigger mutation bookkeeping**
  `const mutated = episodic.length > 0 || staleMarks.length > 0 || stateGc.length > 0;` is computed before `result.markedStale = result.markedStale.filter((p) => !skippedSet.has(p));`, so if every planned stale mark is skipped by the re-lstat guard, the run can still rebuild INDEX and append a `memory.consolidate` event despite applying no mutation.
  Fix: After applying mutations, derive `mutated` from actual `archived`, `markedStale`, and `stateGced` results before rebuilding INDEX or appending the event.

### HonestOracle

- **[suggestion]** `test/memory-consolidate.test.ts:168` — **Test still claims contradictions**
  The test title says "near-duplicate active notes are LISTED as contradictions (no auto-merge)", but the implementation only computes lexical Jaccard similarity and the CLI says "no semantic check". This reintroduces the same source-drift problem in executable documentation: a reader can infer contradiction detection is validated when only near-duplicate listing is tested.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:436` — **Recheck state before unlink**
  The plan validates candidates with `const info = await lstat(full).catch(() => undefined);`, but apply later deletes by path with `await unlink(join(root, rel));`. A same-name replacement between planning and apply is deleted without confirming it is still an old regular current-work file.
  Fix: Re-lstat immediately before unlink and require regular file plus age cutoff.

- **[suggestion]** `src/memory-consolidate.ts:426` — **Clean stale temp files**
  The fixed temp path is created with `await writeFile(tmp, serializeMemoryNote({ ...note, review: "stale" }), { encoding: "utf8", flag: "wx" });` before `await rename(tmp, path);`. A crash or rename failure leaves the temp file behind, and later runs fail on `wx` instead of reconciling.
  Fix: Use a unique temp path and remove it on failure.

### HonestOracle

- **[suggestion]** `src/types.ts:2603` — **Unreadable field overclaims failure class**
  The contract says `"Note files that exist but could not be read/parsed — surfaced, never silently skipped."`, but the implementation also puts parsed episodic notes with invalid dates or created-month/path mismatches into `unreadable`. Callers relying on this field cannot distinguish I/O/parser failure from validation refusal.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/memory-consolidate.ts:21` — **Docstring still claims contradictions**
  The docstring says dry-run reports operations including "which pairs contradict", but `planSimilarPairs` only computes Jaccard lexical overlap and the CLI labels them "near-duplicates for review; no semantic check". This misleads downstream readers into treating a lexical candidate list as contradiction evidence.


---

jcfischer (commented):

2 finding(s): 1 important, 1 suggestion; 1 lens(es) failed to run: Security.


### Security — DID NOT RUN

- **[important]** `(lens output):0` — **Security: model deviated from JSON contract**
  - raw: undefined
  - balanced-largest: undefined
  - trailing: undefined
--- output ---
{"summary":"The consolidation pass mostly maintains the documented symlink guards, but the digest write still crosses the filesystem trust boundary through an existing symlinked digest file. That gives a concrete outside-root overwrite path during an otherwise governed maintenance run.","findings":[{"path":"src/memory-consolidate.ts","line":382,"severity":"blocker","title":"Digest write follows target symlink","rationale":"The diff preflights only parents, then writes with `await writeFile(digestPath, \`# Episodic digest ${month}\\n\\n${body}\\n\`, \"utf8\");`; if `memory/episodic/digests/YYYY-MM.md` is a symlink, this follows it and overwrites outside the memory root. That violates the documented symlink-escape invariant for consolidation destinations.","suggestion":"Before writing, `lstat(digestPath)` and refuse symlinks; write to a real temp file in a real parent dir and `rename` over the digest path."}]}

### Performance

- **[suggestion]** `src/memory-consolidate.ts:441` — **Cache archive parent preflights**
  The loop `for (const e of episodic) await assertSafeArchiveDest(memoryRoot, e.to);` re-walks and `lstat`s the same archive parent chains for every note in a large monthly batch. That makes planning latency scale with repeated parent checks instead of unique destination directories.
  Fix: Memoize checked parent directories during the preflight pass.


---

PR #399, opened 2026-07-04
feat(memory): deterministic audit + Memory skill (M7)

jcfischer:
M7 — final memory milestone. (1) `soma memory audit`: deterministic, read-only, no-LLM SMOKE check with SIX filesystem-ground-truth probes. THREE gate health (CLI exits non-zero on any): root-integrity (every note root is a real directory, not a symlink/non-dir — no fail-open), schema validity, INDEX freshness (mtime). Three informational: digest-coverage counts, orphaned-archive (per canonical created-month digest), event-ratio. Every audit input is O_NOFOLLOW/lstat-guarded — no symlink followed (walk roots+subdirs with a lstat→readdir TOCTOU re-check, INDEX, events, digests); a symlinked INDEX/root can't spoof the gate. (2) `src/skills/Memory/` skill (SKILL.md + Remember/Recall/Consolidate/Audit workflows, pack-id soma-memory-v0.1.0). architecture.md: audit is the shipped ground-truth check (event-ratio is a coarse count, NOT per-note crash-orphan detection). Verification (local): bunx tsc --noEmit clean; 16 audit tests incl. spawned exit-code e2e + skill-entrypoint/frontmatter green.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -162,8 +162,10 @@ created-month digest, and the event/note ratio — and EXITS NON-ZERO on a
 schema-invalid note or a stale INDEX (so it can gate CI). A no-op pass writes no event. The write/event
 coupling is best-effort, not crash-atomic: an event-append *failure* rolls the file
 mutation back, but a hard process crash in the window between the two can still
-orphan a file from its event (a documented gap the M7 audit surfaces; soma has no
-WAL/2PC). This
+orphan a file from its event (soma has no WAL/2PC). The M7 audit does NOT reconcile
+note↔event linkage — its event-ratio probe is a COARSE count (event lines over
+notes), not per-note orphan detection; a missing event can be masked by unrelated
+lines. This
 taxonomy is intentionally distinct from the `MEMORY/*` stores: those stores hold
 curated free-form material; the note store holds single-fact, governed,
 decay-tracked notes.
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -155,11 +155,11 @@ reconciles the STATE (it re-does nothing already done — idempotent), but it do
 back-fill the missing journal record: a subsequent no-op run writes no event, so the
 gap persists. `soma memory audit` (M7) — a deterministic, read-only health check
 that reads the FILES directly, not the event stream — is the ground-truth check:
-it probes schema validity and INDEX freshness (the two health-gating probes), plus
-three informational drift signals — episodic note/digest COUNTS (a coverage
-indicator, not a verified note↔digest mapping), archived notes missing from their
-created-month digest, and the event/note ratio — and EXITS NON-ZERO on a
-schema-invalid note or a stale INDEX (so it can gate CI). A no-op pass writes no event. The write/event
+it has three health-gating probes — root-integrity (every note root is a real
+directory), schema validity, and INDEX freshness — plus three informational drift
+signals — episodic note/digest COUNTS (a coverage indicator, not a verified
+note↔digest mapping), archived notes missing from their created-month digest, and
+the event/note ratio — and EXITS NON-ZERO on any gating failure (so it can gate CI). A no-op pass writes no event. The write/event
 coupling is best-effort, not crash-atomic: an event-append *failure* rolls the file
 mutation back, but a hard process crash in the window between the two can still
 orphan a file from its event (soma has no WAL/2PC). The M7 audit does NOT reconcile
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -153,12 +153,15 @@ part-way, OR the event append itself fails, mutations may already be applied WIT
 the event — so an absent event does NOT prove no mutation happened. A re-run
 reconciles the STATE (it re-does nothing already done — idempotent), but it does NOT
 back-fill the missing journal record: a subsequent no-op run writes no event, so the
-gap persists. A memory audit (M7, forthcoming — not yet built) that reads the files
-directly, not the event stream, is the intended ground-truth check. A no-op pass
-writes no event. The write/event coupling is best-effort, not
-crash-atomic: an event-append *failure* rolls the file mutation back, but a hard
-process crash in the window between the two can still orphan a file from its
-event (a documented gap reconciled by the M7 audit; soma has no WAL/2PC). This
+gap persists. `soma memory audit` (M7) — a deterministic, read-only health check
+that reads the FILES directly, not the event stream — is the ground-truth check:
+it probes schema validity, INDEX freshness, digest coverage, orphaned archive
+notes, and the event/note ratio, and EXITS NON-ZERO on a schema-invalid note or a
+stale INDEX (so it can gate CI). A no-op pass writes no event. The write/event
+coupling is best-effort, not crash-atomic: an event-append *failure* rolls the file
+mutation back, but a hard process crash in the window between the two can still
+orphan a file from its event (a documented gap the M7 audit surfaces; soma has no
+WAL/2PC). This
 taxonomy is intentionally distinct from the `MEMORY/*` stores: those stores hold
 curated free-form material; the note store holds single-fact, governed,
 decay-tracked notes.
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -155,9 +155,11 @@ reconciles the STATE (it re-does nothing already done — idempotent), but it do
 back-fill the missing journal record: a subsequent no-op run writes no event, so the
 gap persists. `soma memory audit` (M7) — a deterministic, read-only health check
 that reads the FILES directly, not the event stream — is the ground-truth check:
-it probes schema validity, INDEX freshness, digest coverage, orphaned archive
-notes, and the event/note ratio, and EXITS NON-ZERO on a schema-invalid note or a
-stale INDEX (so it can gate CI). A no-op pass writes no event. The write/event
+it probes schema validity and INDEX freshness (the two health-gating probes), plus
+three informational drift signals — episodic note/digest COUNTS (a coverage
+indicator, not a verified note↔digest mapping), archived notes missing from their
+created-month digest, and the event/note ratio — and EXITS NON-ZERO on a
+schema-invalid note or a stale INDEX (so it can gate CI). A no-op pass writes no event. The write/event
 coupling is best-effort, not crash-atomic: an event-append *failure* rolls the file
 mutation back, but a hard process crash in the window between the two can still
 orphan a file from its event (a documented gap the M7 audit surfaces; soma has no
--- a/src/cli/memory.ts
+++ b/src/cli/memory.ts
@@ -1,4 +1,5 @@
 import {
+  auditMemory,
   consolidateMemory,
   promoteAlgorithmRunMemory,
   rebuildMemoryIndex,
@@ -15,6 +16,8 @@ import type {
   SomaMemoryActionApproval,
   SomaMemoryActionOptions,
   SomaMemoryActionResult,
+  SomaMemoryAuditOptions,
+  SomaMemoryAuditResult,
   SomaMemoryConsolidateOptions,
   SomaMemoryConsolidateResult,
   SomaMemoryDigestOptions,
@@ -43,6 +46,7 @@ interface MemoryReindexOptions {
 }
 import { readOption } from "./parse-utils";
 import { parseSubstrate } from "./substrate";
+import { SomaCliError } from "./errors";
 
 export interface ParsedMemorySearchArgs {
   command: "memory";
@@ -98,6 +102,12 @@ export interface ParsedMemoryConsolidateArgs {
   options: SomaMemoryConsolidateOptions;
 }
 
+export interface ParsedMemoryAuditArgs {
+  command: "memory";
+  action: "audit";
+  options: SomaMemoryAuditOptions;
+}
+
 export type ParsedMemoryArgs =
   | ParsedMemorySearchArgs
   | ParsedMemoryRecallArgs
@@ -107,9 +117,10 @@ export type ParsedMemoryArgs =
   | ParsedMemoryReindexArgs
   | ParsedMemoryDigestArgs
   | ParsedMemoryActionArgs
-  | ParsedMemoryConsolidateArgs;
+  | ParsedMemoryConsolidateArgs
+  | ParsedMemoryAuditArgs;
 
-const MEMORY_ACTIONS = ["search", "recall", "promote", "write", "verify", "reindex", "digest", "action", "consolidate"] as const;
+const MEMORY_ACTIONS = ["search", "recall", "promote", "write", "verify", "reindex", "digest", "action", "consolidate", "audit"] as const;
 type MemoryAction = (typeof MEMORY_ACTIONS)[number];
 
 export const MEMORY_COMMAND_HELP: { usage: string; subcommands: Record<MemoryAction, string> } = {
@@ -150,6 +161,10 @@ export const MEMORY_COMMAND_HELP: { usage: string; subcommands: Record<MemoryAct
       "Deterministic maintenance: prune aged episodic → digest+archive, mark aged-unverified semantic review:stale, " +
       "list lexically-similar note pairs (near-duplicates for review; no semantic check), rebuild INDEX. --gc-state additionally DELETES current-work state >7d (explicit override). " +
       "--dry-run prints the plan without touching anything.",
+    audit:
+      "Usage: soma memory audit [--home-dir <dir>] [--soma-home <dir>]. " +
+      "Deterministic, read-only health check of the memory tree (no LLM): schema validity, INDEX freshness, " +
+      "digest coverage, orphaned archive notes, event/note ratio. EXITS NON-ZERO on a schema-invalid note or a stale INDEX.",
   },
 };
 
@@ -183,7 +198,23 @@ export function parseMemoryArgs(args: string[]): ParsedMemoryArgs {
       return { command, action, options: parseMemoryActionArgs(rest) };
     case "consolidate":
       return { command, action, options: parseMemoryConsolidateArgs(rest) };
+    case "audit":
+      return { command, action, options: parseMemoryAuditArgs(rest) };
   }
+}
+
+function parseMemoryAuditArgs(args: string[]): SomaMemoryAuditOptions {
+  const options: SomaMemoryAuditOptions = {};
+  for (let index = 0; index < args.length; index += 1) {
+    const arg = args[index];
+    const consumed = consumeSharedMemoryOption(args, index, arg, options);
+    if (consumed > 0) {
+      index += consumed;
+      continue;
+    }
+    throw new Error(MEMORY_COMMAND_HELP.subcommands.audit);
+  }
+  return options;
 }
 
 function parseMemoryConsolidateArgs(args: string[]): SomaMemoryConsolidateOptions {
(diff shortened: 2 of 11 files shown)


jcfischer (commented):

7 finding(s): 3 important, 4 suggestion.


### Architecture

- **[suggestion]** `src/skills/Memory/SKILL.md:3` — **Do not call INDEX a projection**
  CONTEXT.md §project/projection reserves projection for adapter output into a substrate, but the new skill describes an internal memory artifact as "a projected INDEX". That blurs the boundary between Memory indexes and substrate projections.
  Fix: Rename this to "memory INDEX" or "generated INDEX".

- **[suggestion]** `src/skills/Memory/Workflows/Audit.md:11` — **Use principal instead of user**
  CONTEXT.md §principal says `user` is informal only and never in `docs/` or `src/`, but this workflow adds `User: "audit memory", "is memory healthy", "check the memory tree"`. The same label pattern appears in the other new workflow docs.
  Fix: Use `Principal prompt:` or `Principal says:` in the workflow routing examples.

### ContextDrift

- **[important]** `src/skills/Memory/SKILL.md:30` — **Consolidation must not mint trust**
  The diff says "Consolidation is the only path that mints `assistant` trust", but CONTEXT.md §compartment delegates memory-note mechanics to docs/architecture.md §Memory, which says consolidate "never mints or elevates" notes and only ages/relocates existing ones.
  Fix: Replace with: "Consolidation never mints or elevates trust; it only ages or relocates existing notes."

### Performance

- **[suggestion]** `src/memory-audit.ts:37` — **Directory traversal is serialized**
  The recursive scanner awaits each subtree inside the loop: `out.push(...(await listRealMarkdownFilesRec(full)));`. For a memory archive with many sibling month/project directories, audit latency becomes sequential directory-walk latency even though later parsing uses bounded concurrency.
  Fix: Collect recursive directory promises during the loop and `await Promise.all` before flattening results.

### Maintainability

- **[suggestion]** `src/memory-audit.ts:59` — **Extract audit probe helpers**
  `export async function auditMemory` implements enumeration plus five probe sections inline, with repeated `probes.push({ name: ... })` blocks such as `probes.push({ name: "index-freshness", ok: !indexStale, detail: indexReason });`. That makes future probe changes require editing one long stateful function instead of isolated probe units.
  Fix: Extract schema, index-freshness, digest, archive, and event probes into small helpers returning probe data.

### HonestOracle

- **[important]** `src/skills/Memory/SKILL.md:67` — **Audit overclaims invariant coverage**
  The skill says "`soma memory audit` is the deterministic check that these hold (M7)" after listing "A durable write updates the INDEX admission ladder", but the audit only uses the surrogate "INDEX.md is at least as new as every durable note" and never verifies INDEX contents. Downstream users can treat a passing audit as proof of INDEX correctness when a touched or regenerated stale INDEX would still pass.

- **[important]** `src/memory-audit.ts:132` — **Archive probe misses wrong-month digests**
  The comment claims an orphan is a note "not referenced in its created-month digest", and the report can say "every archived episodic note is referenced by a monthly digest". The implementation collects IDs from all digest files and checks only `digestIds.has(note.id)`, so a note referenced in the wrong month is silently accepted.


---

jcfischer (commented):

6 finding(s): 2 important, 3 suggestion, 1 nit.


### CodeQuality

- **[nit]** `src/cli/memory.ts:127` — **Root usage omits audit**
  The diff adds `"audit"` to `MEMORY_ACTIONS`, but the root help still says `Usage: soma memory <search|recall|promote|write|verify|reindex|digest|action|consolidate> ...`, so invalid-command help hides the new subcommand.
  Fix: Add `audit` to the root usage string.

### Security

- **[important]** `src/memory-audit.ts:60` — **INDEX symlink bypasses audit gate**
  The helper uses `return stat(path).then((s) => s.mtimeMs)`, and `probeIndexFreshness` applies it to `INDEX.md`; `stat` follows symlinks. An attacker who can write the memory tree can replace `memory/INDEX.md` with a symlink to a newer file and make a stale or missing real index pass the health gate.
  Fix: Use `lstat` for `INDEX.md` and fail if it is a symlink before reading mtime.

### Performance

- **[suggestion]** `src/memory-audit.ts:45` — **Bound recursive directory fan-out**
  Traversal dispatches every sibling subtree at once via `await Promise.all(subdirs.map(listRealMarkdownFilesRec));`. On a wide archive this creates unbounded concurrent `readdir` work, causing fd pressure and disk contention during audit runs.
  Fix: Use `runBoundedConcurrent(subdirs, listRealMarkdownFilesRec, SCAN_CONCURRENCY)`.

- **[suggestion]** `src/memory-audit.ts:250` — **Stream event line counting**
  `return content.trim() === "" ? 0 : content.trim().split("\n").length;` reads the full event log, trims it twice, and allocates an array of every line. A long append-only JSONL event stream makes audit memory usage proportional to total event history.
  Fix: Count newlines with a streaming reader or a single pass over `content` without `split`.

### Maintainability

- **[suggestion]** `test/memory-audit.test.ts:119` — **Extract archived-note test helper**
  Two tests hand-build the same archived episodic frontmatter shape with duplicated fields like "type: episodic", "last_verified: 2026-07-04", "trust: assistant", and "resurface_count: 0". That makes future schema changes require editing multiple literal arrays.
  Fix: Add a small `writeArchivedEpisodicNote(somaHome, id, created, body)` helper and reuse it in both archive tests.

### HonestOracle

- **[important]** `src/memory-audit.ts:18` — **Audit overclaims invariant coverage**
  The comment claims, "Every §15 design invariant of the memory subsystem maps to a probe here," but the implemented probes include mtime-only index freshness and an event "ratio" that only reports line count over note count. This misleads downstream maintainers into treating smoke signals as invariant enforcement.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-audit.ts:139` — **Unreadable durable files can hide stale INDEX**
  `const durableMtimes = await Promise.all(durableFiles.map(mtimeMs));` followed by `m !== undefined && m > max` treats an undefined durable mtime as absent. If a durable file was listed but cannot be statted, the audit can still pass index freshness even though it did not actually compare every durable note.
  Fix: Make `mtimeMs` return undefined only for ENOENT if the file is no longer present, or fail the index probe when any listed durable file cannot be statted.

### Security

- **[important]** `src/memory-audit.ts:35` — **Top-level memory symlinks are followed**
  The diff says symlinks are rejected because they "could point outside the memory root", but `entries = await readdir(dir, { withFileTypes: true });` follows `dir` itself when `memory/semantic`, `memory/procedural`, archive, or episodic roots are symlinks. A crafted memory tree can make the audit read and trust files outside the memory root while still reporting health.
  Fix: lstat each root directory before readdir and return/fail on symlink or non-directory.

### Performance

- **[suggestion]** `src/memory-audit.ts:162` — **Bound durable mtime stats**
  `const durableMtimes = await Promise.all(durableFiles.map(mtimeMs));` starts one `stat` per durable note concurrently. Large durable corpora can spike filesystem concurrency despite the rest of the audit using `SCAN_CONCURRENCY`.
  Fix: Use `runBoundedConcurrent(durableFiles, mtimeMs, SCAN_CONCURRENCY)`.

### HonestOracle

- **[important]** `src/skills/Memory/Workflows/Audit.md:4` — **Audit is sold as invariant verification**
  The workflow says, "Use it to verify the memory subsystem's invariants hold," but the audit only checks schema parseability, INDEX mtime freshness, and informational counts. That overclaims the evidence and can make downstream agents treat a passing smoke check as proof that memory invariants hold.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/cli/memory.ts:210` — **Unsupported substrate flag is accepted**
  The audit parser delegates to `consumeSharedMemoryOption(args, index, arg, options)`, which also consumes `--substrate`, but audit has no substrate behavior and the audit usage only lists home options. A valid `--substrate` value is silently accepted as a no-op, which is confusing CLI behavior for a health gate.
  Fix: Only accept `--home-dir` and `--soma-home` in `parseMemoryAuditArgs`.

### HonestOracle

- **[important]** `PR description:1` — **Symlink claim is too broad**
  The description claims, "Symlinks never followed," but `countEventLines(somaMemoryEventsPath(somaHome))` reads the events path with `readFile(eventsPath, "utf8")`, which follows a symlinked events file. The claim misleads CI or security readers into believing every audit input is lstat-guarded when the event-ratio probe is not.


---

jcfischer (commented):

3 finding(s): 3 important.


### CodeQuality

- **[important]** `src/memory-audit.ts:225` — **Archive path check can miss notes**
  The check `if (!path.startsWith(archiveDir + sep)) continue;` compares raw path strings, but `archiveDir` comes from `paths.resolve(...)` while parsed file paths come from directory walks; if `somaHome` is relative or otherwise normalized differently, real archive notes can be skipped and orphaned archive drift is not reported.
  Fix: Normalize both sides first, or compare `relative(archiveDir, path)` and require it not start with `..` or be absolute.

### Security

- **[important]** `src/memory-audit.ts:63` — **Recheck files before reading**
  Enumeration skips symlinks with `if (entry.isSymbolicLink()) continue`, but later `return readFile(path, "utf8").then(parseMemoryNote)` and `return stat(path)` dereference the path again. A writable memory tree can replace a listed file with a symlink between enumeration and read/stat, making the audit follow outside-root files under the audit user's permissions.
  Fix: Open/read with O_NOFOLLOW plus fstat, and use lstat for mtime checks.

### HonestOracle

- **[important]** `src/types.ts:2659` — **Probe gating is overstated**
  The exported type claims "Every probe run, in report order (most gate `healthy`; some are informational)", but the implementation gates health only on schema and index freshness while three of five probes are informational. API consumers relying on this comment will misread informational failures as health-gating failures.


---

jcfischer (commented):

3 finding(s): 3 important.


### Security

- **[important]** `src/memory-audit.ts:296` — **Events read can follow swapped symlink**
  The code lstat-checks the events path, then reads it with `const content = await readFile(eventsPath, "utf8").catch((error) => {`, which follows a symlink if the file is swapped after lstat. The diff documents this path as `lstat-guarded` and says the audit `follows NO symlink`, but a racing writable memory tree can make the audit read another file and report its line count.
  Fix: Read the events file with the same `NOFOLLOW_READ` flag used for notes, or open with `O_NOFOLLOW` after the final check.

### HonestOracle

- **[important]** `src/memory-audit.ts:292` — **Event symlink claim is false**
  The comment claims "a symlinked/non-regular events file is not read through" and that the audit "follows NO symlink", but line 300 reads with plain `readFile(eventsPath, "utf8")` after a separate `lstat`. A final-component symlink swapped in after the `lstat` is still followed, so the no-symlink claim misleads downstream readers relying on the audit boundary.

- **[important]** `docs/architecture.md:158` — **Digest coverage is not checked**
  The architecture says the audit probes "digest coverage", but the implemented probe only reports counts: "`${sessionNotes} session + ${actionNotes} action note(s), ${digestFiles} monthly digest file(s)`". That does not establish coverage between episodic notes and digests, so the doc overstates what the audit evidence supports.


---

jcfischer (commented):

7 finding(s): 3 important, 4 suggestion.


### CodeQuality

- **[suggestion]** `src/memory-audit.ts:276` — **Use no-follow reads for digests**
  The digest scan reads with `content: await readFile(path, "utf8").catch(() => "")`, so a digest file swapped to a symlink after enumeration can still be followed. That can make the orphaned-archive probe report coverage from an outside file even though the rest of the audit uses `NOFOLLOW_READ` for this edge case.
  Fix: Read digest files with `{ encoding: "utf8", flag: NOFOLLOW_READ }` as well.

### Security

- **[important]** `src/memory-audit.ts:278` — **Digest read follows swapped symlink**
  `content: await readFile(path, "utf8").catch(() => "")` reopens digest paths without `O_NOFOLLOW`, so a digest file swapped after enumeration can make the audit read outside the memory tree and spoof archived-note coverage. This violates the PR's security invariant that audit inputs are not followed through symlinks.
  Fix: Read digest files with `NOFOLLOW_READ` like notes/events, and treat failed reads as unreadable coverage input.

### Architecture

- **[suggestion]** `src/types.ts:2639` — **Constrain public probe names**
  The exported API uses `export interface SomaMemoryAuditProbe { name: string; ok: boolean; detail: string; }` even though the same diff defines a fixed probe set and health-gating semantics. Since `auditMemory` is exported from `src/index.ts`, tightening arbitrary `string` into stable probe kinds later would be a public API change.
  Fix: Use a literal union for probe names, and include explicit gating metadata if consumers need it.

### Performance

- **[suggestion]** `src/memory-audit.ts:263` — **Skip irrelevant digest reads**
  `const digestIdsByMonth = await collectDigestIdsByMonth(digestFilesList);` reads and indexes every digest before knowing which archive months need validation, and `for (const line of content.split("\n"))` allocates per digest. On large histories or empty archives, every audit pays avoidable disk I/O and line-array allocation.
  Fix: Collect archive note months first, then read only matching digest files; stream or scan content without `split`. 

### HonestOracle

- **[important]** `PR description:1` — **Exit-code e2e is overstated**
  The description claims "e2e healthy→exit0 / unhealthy→exit1", but the shown test only calls `runMemoryCli(parseMemoryArgs(...))` and asserts `rejects.toThrow(/UNHEALTHY/)`. No spawned CLI or process exit-code assertion appears in the diff, so a function-level throw is being sold as exit-code e2e verification.

- **[important]** `src/skills/Memory/SKILL.md:40` — **Audit cannot flag all hand edits**
  The workflow says "Hand edits desync the INDEX (the audit will flag them)." The same diff limits audit to "mtime freshness — NOT a check of INDEX contents", so manual edits that preserve or mask mtime freshness are outside what the audit can prove.

- **[suggestion]** `src/skills/Memory/Workflows/Audit.md:4` — **Event ratio is not ground truth**
  The workflow claims "every probe reports a filesystem ground-truth fact", but `countEventLines` uses `readFile(...).catch(() => "")`, so an unreadable events file is reported as 0 lines. That is a silent fallback, not a filesystem ground-truth event count.


---

jcfischer (commented):

### Security

- **[important]** `src/memory-audit.ts:44` — **Directory walk can follow swapped symlinks**
  The sequence "const dirInfo = await lstat(dir)" followed by "entries = await readdir(dir, { withFileTypes: true });" leaves a TOCTOU gap: an attacker controlling the memory tree can replace a checked directory with a symlink before readdir. That lets the audit traverse outside somaHome while still reporting trusted relative paths, violating the no-symlink invariant.
  Fix: Use fd-based no-follow directory traversal, or fail if a post-read lstat no longer matches the checked directory.

### HonestOracle

- **[blocker]** `src/memory-audit.ts:37` — **Root replacement passes as empty**
  The code says `if (!dirInfo.isDirectory()) return []; // a symlink or non-directory — never follow it`, so a replaced `memory/semantic` or `memory/procedural` root is treated as no notes rather than an unhealthy tree. That fails open: the report can say HEALTHY while the durable memory corpus is inaccessible, misleading CI or recall users relying on the audit.

- **[important]** `docs/architecture.md:168` — **Crash-orphan detection is overstated**
  The docs claim `orphan a file from its event (a documented gap the M7 audit surfaces; soma has no WAL/2PC)`, but the added probe only reports `event-stream lines over valid-note count` and never links notes to journal entries. A missing event can be masked by unrelated event lines, so this presents a surrogate ratio as detection of the crash orphan gap.


---

jcfischer (commented):

3 finding(s): 3 important.


### CodeQuality

- **[important]** `src/skills/Memory/Workflows/Audit.md:34` — **Exit causes omit root-integrity**
  The workflow says `Cause is always a schema-invalid note or a stale INDEX (the two health-gating probes)`, but the implementation adds `root-integrity` with `gatesHealth: true`. A root replacement can therefore make the audit exit non-zero while this guidance points to the wrong causes and fixes.
  Fix: Mention root-integrity as a health-gating probe and add its remediation.

### EcosystemCompliance

- **[important]** `src/skills/Memory/Workflows/Audit.md:37` — **Document root-integrity gate**
  The workflow says "Cause is always a schema-invalid note or a stale INDEX (the two health-gating probes)", but the audit emits `name: "root-integrity"` with `gatesHealth: true`. Agents following this skill can misreport an UNHEALTHY audit and omit the required remediation path.
  Fix: Add root-integrity to the probe table, exit-code causes, and failure handling; align SKILL.md/types comments too.

### HonestOracle

- **[important]** `src/cli/memory.ts:164` — **Exit causes omit root-integrity**
  The help text says audit "EXITS NON-ZERO on a schema-invalid note or a stale INDEX," but `probeTreeIntegrity` is also `gatesHealth: true` and `healthy` is derived from every gated probe. A user relying on this help can get a non-zero CI failure from a replaced root that the documented exit contract says is not an exit cause.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/types.ts:2629` — **Audit health comment omits root-integrity**
  The exported contract says the CLI exits non-zero for "a schema-invalid note or a stale INDEX", but `root-integrity` is also a health-gating probe in the same type block. Consumers reading this comment can incorrectly assume abnormal roots do not affect `healthy`.
  Fix: Update the comment to include root-integrity as a non-zero exit cause.

### EcosystemCompliance

- **[important]** `src/types.ts:2648` — **Public gate contract omits root-integrity**
  The exported probe contract says "only `schema` and `index-freshness` gate", but the diff also makes `root-integrity` a health-gating probe. Consumers relying on this public type documentation can treat an unhealthy root as non-gating.
  Fix: Update the comment to name all three gates: `root-integrity`, `schema`, and `index-freshness`.

### HonestOracle

- **[important]** `src/types.ts:2628` — **Audit gate contract omits root-integrity**
  The public comment says `healthy` is false for "a schema-invalid note or a stale INDEX", but the implementation also gates on `root-integrity`. API consumers relying on this exported contract can misclassify an unhealthy tree where a note root is a symlink/non-directory.

- **[important]** `src/types.ts:2648` — **Probe docs deny root-integrity gate**
  The probe contract says "only `schema` and `index-freshness` gate", but `root-integrity` is emitted with `gatesHealth: true`. This directly contradicts the machine-readable result shape the PR exposes.


---

jcfischer (commented):

2 finding(s): 2 important.


### CodeQuality

- **[important]** `src/memory-audit.ts:268` — **Non-digest markdown can satisfy archive coverage**
  The filter accepts any markdown file whose basename matches a needed month: `basename(p).replace(/\.md$/, "")`. A file like `memory/episodic/digests/2026-07/notes.md` is listed by the recursive walk and can be treated as month `notes`, while `memory/episodic/digests/nested/2026-07.md` can satisfy coverage even though it is not the canonical created-month digest file.
  Fix: Only accept canonical digest files directly under the digest root, e.g. compare `relative(digestsDir, p)` against `${month}.md` without path separators.

### HonestOracle

- **[important]** `PR description:1` — **Description omits root-integrity gate**
  The description says "five filesystem-ground-truth probes — schema validity + INDEX freshness GATE health (CLI exits non-zero on either)" but the diff documents and implements a third health gate, "root-integrity". A downstream reader relying on the PR description would incorrectly believe only schema and INDEX failures can make the CLI exit non-zero.


---

jcfischer (commented):

### CodeQuality

- **[nit]** `src/types.ts:2665` — **Healthy comment omits root-integrity**
  The comment says `false ⇒ the CLI exits non-zero. Gated by: no schema-invalid notes AND a fresh INDEX.`, but the implementation also gates on `root-integrity`. This public type comment can mislead consumers reading `SomaMemoryAuditResult`.
  Fix: Include `root-integrity` in the `healthy` comment.

### HonestOracle

- **[important]** `src/types.ts:2656` — **Healthy contract omits root-integrity**
  The comment says `false ⇒ the CLI exits non-zero. Gated by: no schema-invalid notes AND a fresh INDEX.`, but the implementation also gates on `root-integrity`. This misleads downstream callers reading the exported API contract about when `healthy` can be false.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/memory-audit.ts:269` — **Invalid archived notes skip orphan reporting**
  The archive probe drops invalid notes with `if (note === undefined || note.type !== "episodic") continue;`, so a malformed archived markdown file is reported only under schema and never under `orphanedArchive`. That makes the informational archive drift count incomplete for files physically under `memory/archive/`.
  Fix: Classify archived paths before schema filtering, or document that orphaned-archive only covers parseable episodic notes.

### HonestOracle

- **[important]** `src/memory-audit.ts:16` — **Gate list omits root-integrity**
  The comment says health fails on "a schema-invalid note, or a stale INDEX" and then labels "The rest" informational, but the implementation also makes `root-integrity` a health-gating probe. That misstates the exported audit contract for callers relying on this source comment.

- **[important]** `src/skills/Memory/Workflows/Audit.md:7` — **Passing audit can still report drift**
  The workflow claims "a passing audit means no drift was DETECTED", but the same diff marks `orphaned-archive` as "no (info)" and the implementation reports missing archived notes while keeping the audit healthy. A user gating on exit 0 can be misled into ignoring detected informational drift.


---

jcfischer (commented):

3 finding(s): 3 suggestion.


### CodeQuality

- **[suggestion]** `src/memory-audit.ts:296` — **Validate archive note month**
  The archive probe derives the digest month with `const month = note.created.slice(0, 7);`, so any parseable note with a malformed or non-ISO `created` value is checked against a bogus digest key instead of being reported as schema-invalid. That can make `orphaned-archive` report misleading drift for data that is actually a bad note timestamp.
  Fix: Validate `note.created` as `YYYY-MM...` before using it, or make the note parser reject malformed `created` values.

### Performance

- **[suggestion]** `src/memory-audit.ts:341` — **Avoid splitting whole digest files**
  The loop `for (const line of content.split("\n"))` allocates an array and per-line strings for each digest after already reading the full file. Large monthly digests will transiently double memory pressure during `soma memory audit`.
  Fix: Scan `content` line-by-line by index, or use a streaming line reader.

### HonestOracle

- **[suggestion]** `PR description:1` — **Unshown skill validation claim**
  The description claims "skill-entrypoint/frontmatter green", but the supplied changed tests only show `test/memory-audit.test.ts` audit coverage and no assertion over `src/skills/Memory/SKILL.md` frontmatter or workflow entrypoints. That makes the verification statement stronger than the evidence attached to this PR data.


---

PR #400, opened 2026-07-04
feat(memory): SessionEnd deterministic digest fallback (M5b)

jcfischer:
M5b — the deferred SessionEnd hook, on JC's HYBRID authorship decision. PRIMARY: assistant self-authors a digest at wrap-up (`soma memory digest --body`, real summary; guided by the Memory skill's Remember workflow). FALLBACK (deterministic/no-LLM): on SessionEnd a transcript-extraction path writes an 8–15-line body, marked `lifecycleEvent: session-end` (→ the note's M0 `hook:` field) + `provenance: tool:claude-session-end`, ONLY if no digest exists yet (one-per-session gate, checked before reading the transcript).

Extraction filters command/system/tool_result + sidechain lines, quotes+labels each principal prompt (JSON-escaped; REDUCES, not eliminates, injection risk), control-collapses tool names into a rollup, head/tail-samples, returns undefined below 6 prompts (thin sessions are DROPPED — the fallback neither checks nor guarantees an assistant digest exists). Prompt/tool text are control-collapsed+truncated pointers, not verbatim; malformed JSONL lines are silently skipped (best-effort, not a complete record).

Sub-agent suppression (ADR 0014) is MARKED-only: the hook-runner reads Claude Code's own payload keys `agent_id`/`agent_type` and forwards them as Soma's qualified `--subagent-id`/`--subagent-type` (bare `agent` is banned in Soma surfaces → fields `subagentId`/`subagentType`); an UNMARKED sub-agent is NOT detected and falls through. `forcePrimary`/env overrides.

Architecture: `soma memory digest` stays body-only/substrate-neutral. The fallback routes through `soma lifecycle session-end` via DEPENDENCY INVERSION — core lifecycle owns `registerSessionEndTranscriptHandler(substrate, handler)` and looks it up by substrate; the Claude adapter registers its fallback at module load; the composition layer (cli/lifecycle.ts) side-effect-imports it. Core imports NO adapter. The fallback writes through the GOVERNED M5 memory-write path (schema + one-per-session gate + provenance validation + memory.digest event), not the tool-activity writeback queue.

Verification: CI on this branch is green — **Deterministic portability** (full `bun test` suite) + **gate / confidentiality-gate** both SUCCESS (see the Checks tab); the pre-push smoke gate reruns the full suite on every push. Locally: bunx tsc --noEmit clean; hook-runner.mjs `node --check` passes; a spawned-CLI test (`soma lifecycle session-end --transcript`, fresh process) asserts the fallback digest lands with hook:session-end + the tool provenance, and `soma memory digest --transcript` is rejected.

### DIFF
--- a/src/adapters/claude-code/hook-runner.mjs
+++ b/src/adapters/claude-code/hook-runner.mjs
@@ -92,8 +92,10 @@ function lifecycle(config, event, input) {
     const transcriptPath =
       typeof input.transcript_path === "string" && input.transcript_path.trim().length > 0 ? input.transcript_path : undefined;
     if (transcriptPath) args.push("--transcript", transcriptPath);
-    if (typeof input.agent_id === "string" && input.agent_id.trim().length > 0) args.push("--agent-id", input.agent_id);
-    if (typeof input.agent_type === "string" && input.agent_type.trim().length > 0) args.push("--agent-type", input.agent_type);
+    // Claude Code's payload keys are `agent_id`/`agent_type`; Soma's flags are the
+    // qualified `--subagent-*` (bare `agent` is banned in Soma surfaces).
+    if (typeof input.agent_id === "string" && input.agent_id.trim().length > 0) args.push("--subagent-id", input.agent_id);
+    if (typeof input.agent_type === "string" && input.agent_type.trim().length > 0) args.push("--subagent-type", input.agent_type);
   }
   runSomaDetached(config, args);
 }
--- a/src/adapters/claude-code/hook-runner.mjs
+++ b/src/adapters/claude-code/hook-runner.mjs
@@ -85,6 +85,26 @@ function lifecycle(config, event, input) {
   const cwd = typeof input.cwd === "string" && input.cwd.trim().length > 0 ? input.cwd : undefined;
   if (cwd) args.push("--cwd", cwd);
   runSomaDetached(config, args);
+  // M5b: on SessionEnd, also fire the deterministic digest FALLBACK (assistant-authored
+  // digests come from the wrap-up rule, not here). Sub-agent suppression + one-per-session
+  // dedup live in the CLI/SDK; this only forwards the payload. Detached, never blocks.
+  if (event === "session-end") sessionEndDigest(config, input, id);
+}
+
+function sessionEndDigest(config, input, id) {
+  const transcriptPath =
+    typeof input.transcript_path === "string" && input.transcript_path.trim().length > 0 ? input.transcript_path : undefined;
+  if (!id || !transcriptPath) return; // nothing to summarize without a session id + transcript
+  const args = [
+    "src/cli.ts", "memory", "digest",
+    "--session", id,
+    "--transcript", transcriptPath,
+    "--soma-home", config.somaHome,
+    "--substrate", "claude-code",
+  ];
+  if (typeof input.agent_id === "string" && input.agent_id.trim().length > 0) args.push("--agent-id", input.agent_id);
+  if (typeof input.agent_type === "string" && input.agent_type.trim().length > 0) args.push("--agent-type", input.agent_type);
+  runSomaDetached(config, args);
 }
 
 function metadata(input, source) {
--- a/src/adapters/claude-code/hook-runner.mjs
+++ b/src/adapters/claude-code/hook-runner.mjs
@@ -84,26 +84,17 @@ function lifecycle(config, event, input) {
   if (id) args.push("--session-id", id);
   const cwd = typeof input.cwd === "string" && input.cwd.trim().length > 0 ? input.cwd : undefined;
   if (cwd) args.push("--cwd", cwd);
-  runSomaDetached(config, args);
-  // M5b: on SessionEnd, also fire the deterministic digest FALLBACK (assistant-authored
-  // digests come from the wrap-up rule, not here). Sub-agent suppression + one-per-session
-  // dedup live in the CLI/SDK; this only forwards the payload. Detached, never blocks.
-  if (event === "session-end") sessionEndDigest(config, input, id);
-}
-
-function sessionEndDigest(config, input, id) {
-  const transcriptPath =
-    typeof input.transcript_path === "string" && input.transcript_path.trim().length > 0 ? input.transcript_path : undefined;
-  if (!id || !transcriptPath) return; // nothing to summarize without a session id + transcript
-  const args = [
-    "src/cli.ts", "memory", "digest",
-    "--session", id,
-    "--transcript", transcriptPath,
-    "--soma-home", config.somaHome,
-    "--substrate", "claude-code",
-  ];
-  if (typeof input.agent_id === "string" && input.agent_id.trim().length > 0) args.push("--agent-id", input.agent_id);
-  if (typeof input.agent_type === "string" && input.agent_type.trim().length > 0) args.push("--agent-type", input.agent_type);
+  // M5b: on SessionEnd, forward the transcript path + sub-agent markers so the
+  // lifecycle handler can attempt the deterministic digest FALLBACK (dispatched to the
+  // Claude Code transcript adapter — `soma memory digest` itself stays neutral/body-only).
+  // Assistant-authored digests come from the wrap-up rule, not here.
+  if (event === "session-end") {
+    const transcriptPath =
+      typeof input.transcript_path === "string" && input.transcript_path.trim().length > 0 ? input.transcript_path : undefined;
+    if (transcriptPath) args.push("--transcript", transcriptPath);
+    if (typeof input.agent_id === "string" && input.agent_id.trim().length > 0) args.push("--agent-id", input.agent_id);
+    if (typeof input.agent_type === "string" && input.agent_type.trim().length > 0) args.push("--agent-type", input.agent_type);
+  }
   runSomaDetached(config, args);
 }
 
--- a/src/adapters/claude-code/session-digest.ts
+++ b/src/adapters/claude-code/session-digest.ts
@@ -1,5 +1,6 @@
 import { readFile } from "node:fs/promises";
 import { hasSessionDigest, writeSessionDigest } from "../../memory-episodic";
+import { registerSessionEndTranscriptHandler } from "../../lifecycle";
 import type { SomaMemoryDigestResult, SubstrateId } from "../../types";
 
 /**
@@ -143,11 +144,12 @@ export function extractDigestBodyFromTranscript(transcript: string): string | un
     .map(([name, n]) => `${name}×${n}`)
     .join(", ");
 
-  // Prompt lines are VERBATIM principal input. Quote + label them (JSON.stringify
-  // escapes quotes/controls) so they read as DATA — what was asked — rather than as
-  // assistant instructions. This REDUCES the prompt-injection risk (paired with the
-  // tool:<name> provenance below); it does not PROVE every downstream recall consumer
-  // treats embedded instructions as inert.
+  // Prompt lines are principal input, control-collapsed + truncated by cleanLine (a
+  // lossy pointer, NOT an exact excerpt). Quote + label them (JSON.stringify escapes
+  // quotes/controls) so they read as DATA — what was asked — rather than as assistant
+  // instructions. This REDUCES the prompt-injection risk (paired with the tool:<name>
+  // provenance below); it does not PROVE every downstream recall consumer treats
+  // embedded instructions as inert.
   const lines = [
     `- session: ${prompts.length} principal prompts, ${assistantTurns} assistant turns, ${totalTools} tool calls`,
     ...shown.map((p) => `- principal prompt: ${JSON.stringify(p)}`),
@@ -244,3 +246,8 @@ export async function writeSessionDigestFromTranscript(options: ClaudeSessionDig
     reason: digest.created ? "wrote a deterministic session-end fallback digest" : "a digest already exists for this session — no-op",
   };
 }
+
+// Dependency inversion: register this adapter's transcript fallback with core
+// lifecycle at module load. Core never imports this adapter — it looks the handler up
+// by substrate. Importing this module (CLI side-effect / tests) triggers registration.
+registerSessionEndTranscriptHandler("claude-code", (input) => writeSessionDigestFromTranscript(input));
--- a/src/adapters/claude-code/session-digest.ts
+++ b/src/adapters/claude-code/session-digest.ts
@@ -14,7 +14,9 @@ import type { SomaMemoryDigestResult, SubstrateId } from "../../types";
  */
 
 // A deterministic fallback needs enough genuine prompts to make a useful 8–15-line
-// digest — below this it skips (the assistant self-authors small sessions instead).
+// digest — below this it returns undefined and the session is simply DROPPED (no
+// fallback digest). A small session is only captured if the assistant self-authored
+// one; the fallback neither checks nor guarantees that.
 const FALLBACK_MIN_PROMPTS = 6;
 const FALLBACK_MAX_PROMPT_LINES = 13; // + 2 structural (header + tools) = 15 max
 const PROMPT_LINE_MAX_CHARS = 120;
@@ -249,7 +251,7 @@ export async function writeSessionDigestFromTranscript(options: ClaudeSessionDig
     now: options.now,
     sessionId: options.sessionId,
     body,
-    hook: "session-end",
+    lifecycleEvent: "session-end", // → the note's M0 `hook:` field
     // Distinct provenance: recall's trust banner shows this body was machine-extracted
     // from principal input, NOT assistant-authored.
     provenance: "tool:claude-session-end",
--- a/src/adapters/claude-code/session-digest.ts
+++ b/src/adapters/claude-code/session-digest.ts
@@ -1,4 +1,7 @@
+import { constants as fsConstants } from "node:fs";
 import { readFile } from "node:fs/promises";
+import { homedir } from "node:os";
+import { isAbsolute, join, relative, resolve } from "node:path";
 import { hasSessionDigest, writeSessionDigest } from "../../memory-episodic";
 import { registerSessionEndTranscriptHandler } from "../../lifecycle";
 import type { SomaMemoryDigestResult, SubstrateId } from "../../types";
@@ -183,8 +186,15 @@ export interface ClaudeSessionDigestOptions {
   now?: Date;
   /** The session this digest summarizes. */
   sessionId: string;
-  /** Path to the Claude Code session transcript (JSONL). */
+  /** Path to the Claude Code session transcript (JSONL). VALIDATED against transcriptRoot. */
   transcriptPath: string;
+  /**
+   * The directory the transcript MUST live under — a hook-controlled path is only read
+   * if it resolves inside this root and ends in `.jsonl`. Defaults to
+   * `$SOMA_CLAUDE_TRANSCRIPT_ROOT` or `~/.claude/projects` (Claude Code's transcript
+   * store). Injected by tests to point at a fixture dir.
+   */
+  transcriptRoot?: string;
   /** Claude Code sub-agent markers from the hook payload — when set, the digest is suppressed. */
   subagentId?: string;
   subagentType?: string;
@@ -196,7 +206,7 @@ export interface ClaudeSessionDigestOptions {
 
 /** Outcome of the transcript-fallback digest. Each cause is distinguished. */
 export interface ClaudeSessionDigestResult {
-  outcome: "written" | "duplicate" | "suppressed" | "skipped" | "unreadable";
+  outcome: "written" | "duplicate" | "suppressed" | "skipped" | "unreadable" | "refused";
   /** The core digest result when a write/dedup happened; absent otherwise. */
   digest?: SomaMemoryDigestResult;
   /** Human-readable reason (always set). */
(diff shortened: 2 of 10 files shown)


jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/memory-episodic.ts:416` — **forceSubagent loses to forcePrimary**
  The type promises `Force treating the invocation as a sub-agent (suppress) regardless of markers`, but `if (isSubagent && options.forcePrimary !== true)` allows `forcePrimary` to bypass suppression when both flags are true.
  Fix: Give `forceSubagent` precedence, or reject both flags when both are set.

### Architecture

- **[important]** `src/types.ts:2522` — **Claude transcript leaks into core API**
  docs/architecture.md#Core says the core is "filesystem-native and substrate-neutral", but this exported type defines `Write a session digest whose body is DETERMINISTICALLY extracted from a Claude Code transcript` and `Path to the Claude Code session transcript (JSONL).` That makes a substrate-native transcript format part of Soma's public core surface instead of keeping it behind the Claude Code adapter.
  Fix: Move Claude transcript extraction types/functions behind `src/adapters/claude-code` and expose a substrate-neutral digest body writer in core.

### Performance

- **[suggestion]** `src/memory-episodic.ts:342` — **Check duplicate before transcript parsing**
  The fallback does `const transcript = await readFile(options.transcriptPath, "utf8").catch(() => undefined);` and then `const body = extractDigestBodyFromTranscript(transcript);` before relying on `writeSessionDigest` to discover duplicates. In the intended primary path, existing assistant-authored digests make that transcript read/split/JSON parse wasted SessionEnd work.
  Fix: Call the existing session digest lookup/gate before reading `options.transcriptPath`, or expose a cheap `hasSessionDigest` helper.

### Maintainability

- **[suggestion]** `test/memory-episodic.test.ts:264` — **Extract transcript fixture setup**
  Multiple tests repeat the same temp-file fixture block: `const dir = await mkdtemp(join(tmpdir(), "soma-tx-"));`, `const tp = join(dir, "t.jsonl");`, `await writeFile(tp, SEVEN_PROMPTS, "utf8");`, and `await rm(dir, { recursive: true, force: true });`. That makes future fixture changes touch several tests and keeps cleanup logic copy-pasted.
  Fix: Add a small helper such as `withTranscriptFile(SEVEN_PROMPTS, async (tp) => ...)`.

### HonestOracle

- **[important]** `src/types.ts:2545` — **Skipped outcome hides read failures**
  The result contract says `"skipped" (too little transcript content)`, but the implementation also returns `skipped` for an unreadable transcript. API consumers relying on this comment will treat an I/O failure as a thin-session decision.

- **[important]** `src/skills/Memory/Workflows/Remember.md:57` — **Fallback guarantee is overstated**
  The workflow claims `substantial sessions are never lost`, but the PR also documents and implements silent skips for unreadable transcripts. That overstates the fallback from best-effort capture into a guarantee.


---

jcfischer (commented):

### Security

- **[important]** `src/adapters/claude-code/session-digest.ts:108` — **User prompts enter trusted memory**
  The fallback copies transcript prompt text into the digest with `...shown.map((p) => `- ${p}`)` and writes it through the normal session digest path with `body, hook: "session-end"`. Because session digests are assistant-trust memory, a user can plant durable prompt-injection text by supplying six prompts and letting the fallback run.
  Fix: Store fallback digests with untrusted/user provenance or quote/label prompt lines so trusted recall cannot treat them as assistant-authored instructions.

### Architecture

- **[important]** `src/index.ts:470` — **Claude fallback leaks through root API**
  The diff exports adapter-specific APIs from the package root: `export { extractDigestBodyFromTranscript, writeSessionDigestFromTranscript, type ClaudeSessionDigestOptions, type ClaudeSessionDigestResult, } from "./adapters/claude-code/session-digest";`. This conflicts with docs/architecture.md#Core: `The core is filesystem-native and substrate-neutral`, and makes a Claude transcript helper part of the stable-looking public surface.
  Fix: Keep these exports off `src/index.ts`; have CLI/tests import from `src/adapters/claude-code/session-digest` or expose an explicit adapter subpath.

### Performance

- **[suggestion]** `src/adapters/claude-code/session-digest.ts:178` — **Stream transcript parsing**
  The fallback reads and duplicates the full transcript with "const transcript = await readFile(options.transcriptPath, \"utf8\")" and then iterates "for (const raw of transcript.split(\"\\n\"))". Large Claude JSONL transcripts pay an avoidable memory spike from the full file string plus split array before extraction can produce a bounded 8-15-line digest.
  Fix: Stream the JSONL file line-by-line and keep only bounded prompt samples plus tool counts.

### HonestOracle

- **[suggestion]** `src/adapters/claude-code/session-digest.ts:64` — **Tool noise claim overstates filtering**
  The comment claims the extractor handles digest text with "noise/tool/command lines filtered", but the same function counts assistant `tool_use` parts into `toolCounts` and emits them as `- tools: ...`. A downstream reader could take the comment as saying tool activity is excluded rather than summarized.


---

jcfischer (commented):

3 finding(s): 2 important, 1 suggestion; 1 lens(es) failed to run: Architecture.


  - raw: undefined
  - balanced-largest: undefined
  - trailing: undefined
--- output ---
{"summary":"The transcript extractor is mostly adapter-local, but the PR still freezes Claude Code hook vocabulary into neutral memory surfaces. That matters for this lens because the public CLI/type shape is harder to unwind than adapter implementation details.","findings":[{"path":"src/types.ts","line":2518,"severity":"important","title":"Core digest API exposes hook vocabulary","rationale":"The diff adds `hook?: string;` to `SomaMemoryDigestOptions` and documents it as `Optional \`hook:\` frontmatter marker`; docs/architecture.md#Lifecycle Harness says Soma ports hooks as a substrate-neutral lifecycle harness, and CONTEXT.md#runtime-policy-inspection treats hooks as substrate-native projection mechanisms. This makes a substrate hook marker part of the core memory API instead of modeling it as neutral lifecycle provenance.","suggestion":"Replace `hook` with a neutral provenance/source field such as `lifecycleEvent: \"session_end\"` or keep the marker inside the Claude Code adapter output only."},{"path":"src/cli/memory.ts","line":15,"severity":"important","title":"Memory CLI depends on Claude adapter","rationale":"The generic memory CLI now imports `} from \"../adapters/claude-code/session-digest\";` and exposes `--transcript`, `--agent-id`, and `--agent-type` on `soma memory digest`. docs/architecture.md#Adapters assigns substrate-native translation to adapters, so this turns a neutral memory command into a Claude Code adapter dispatcher and freezes Claude-specific options into the public CLI surface.","suggestion":"Keep `soma memory digest` body-only and route transcript fallback through a Claude Code adapter/lifecycle command, or dispatch via a substrate adapter registry rather than importing Claude Code directly."}]}

### Maintainability

- **[suggestion]** `src/adapters/claude-code/session-digest.ts:65` — **Extract transcript entry classification**
  `extractDigestBodyFromTranscript` mixes line parsing, sidechain filtering, prompt filtering, and assistant tool counting inside one loop: `for (const raw of transcript.split("\n")) {` followed by `if (entry.type === "user"` and `else if (entry.type === "assistant"`. This is readable now, but future transcript-shape tweaks will have to preserve several intertwined branches at once.
  Fix: Extract a small `classifyTranscriptEntry(entry)` or separate `collectPrompt` / `collectToolUses` helpers.

### HonestOracle

- **[important]** `src/adapters/claude-code/session-digest.ts:125` — **Injection closure is overclaimed**
  The comment claims `JSON.stringify escapes quotes/controls, closing the injection vector.` Quoting prompt text does not prove downstream recall consumers will treat embedded instructions as inert, so this misleads readers into believing prompt-injection risk is eliminated rather than reduced.


---

jcfischer (commented):

7 finding(s): 4 important, 3 suggestion.


### CodeQuality

- **[important]** `src/adapters/claude-code/session-digest.ts:79` — **Guard parsed transcript entries**
  The parser does `entry = JSON.parse(line) as Record<string, unknown>;` and then immediately reads `entry.isSidechain`; a valid JSONL line of `null` makes that property access throw instead of being skipped. That contradicts the local contract that non-useful lines should `return undefined; // a non-JSON line — skip, never fail the whole extraction`.
  Fix: After parsing, return undefined unless the value is a non-null object and not an array.

### Security

- **[important]** `src/memory-episodic.ts:181` — **Validate digest provenance override**
  The diff stores caller-controlled provenance directly via `provenance: provenance ?? "conversation"`, while the new API documents only `conversation` | `import` | `tool:<name>`. A public `writeSessionDigest` caller can now persist malformed or forged provenance into trusted recall metadata.
  Fix: Reject provenance outside `conversation`, `import`, or `/^tool:[a-z0-9][a-z0-9._-]{0,63}$/i` before building the note.

### Architecture

- **[important]** `src/cli/memory.ts:15` — **Generic memory CLI imports Claude adapter**
  docs/architecture.md says the core is "substrate-neutral" and that "Adapters project Soma into substrate-native primitives," but the generic memory CLI now imports `writeSessionDigestFromTranscript` from `"../adapters/claude-code/session-digest"` and exposes `--transcript`, `--agent-id`, and `--agent-type` through `soma memory digest`. This makes the public memory command depend on the first substrate fallback instead of routing transcript harvesting through the adapter/lifecycle boundary.
  Fix: Keep `memory digest` body-only and add a Claude Code adapter/lifecycle command or adapter-owned CLI path for transcript fallback.

### Performance

- **[suggestion]** `src/adapters/claude-code/session-digest.ts:208` — **Avoid duplicate digest scans**
  The fallback checks `if (await hasSessionDigest({ homeDir: options.homeDir, somaHome: options.somaHome, sessionId: options.sessionId }))` and then writes through `const digest = await writeSessionDigest({`, whose core path already performs the session digest lookup before writing. For fallback sessions that actually write, this does the cross-date digest scan twice before one note is created.
  Fix: Let `writeSessionDigest` return the duplicate outcome, or add a core write API that accepts the already-known no-duplicate result.

### Maintainability

- **[suggestion]** `src/cli/memory.ts:98` — **Avoid nested options.options**
  The new union stores payloads as `options` inside `ParsedMemoryDigestArgs.options`, leading to call sites like `writeSessionDigestFromTranscript(parsed.options.options)`. That repeated `options` name obscures whether the outer command parse or the mode-specific payload is being passed.
  Fix: Rename the inner field to `payload` or `digestOptions`.

### HonestOracle

- **[important]** `src/adapters/claude-code/session-digest.ts:184` — **Sub-agent marker assumption unsupported**
  The comment claims "A sub-agent invocation carries a non-empty agent id or type in the hook payload," then the function advertises "Suppresses sub-agent sessions (ADR 0014)." The diff only proves suppression for marked invocations, so an unmarked sub-agent transcript silently falls through to the write path while readers are told suppression is complete.

- **[suggestion]** `test/memory-episodic.test.ts:252` — **Injection safety still overclaimed**
  The test comment says "prompt text is quoted + labeled (injection-safe)," but the assertion only checks that one benign prompt is quoted. It does not establish safety against adversarial prompt text or downstream recall consumers treating the text as inert.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/types.ts:2530` — **Document all accepted provenance values**
  The new API comment says `Optional provenance override (`conversation` | `import` | `tool:<name>`),` but the implementation accepts `conversation`, `consolidation`, `import`, and `tool:<name>`. That mismatch makes the public type comment inaccurate for valid callers.
  Fix: Include `consolidation` in the provenance comment or narrow the validator.

### Architecture

- **[important]** `src/lifecycle.ts:9` — **Core lifecycle imports Claude adapter**
  The diff adds `import { writeSessionDigestFromTranscript } from "./adapters/claude-code/session-digest";` inside core lifecycle, but docs/architecture.md says "The core is filesystem-native and substrate-neutral" and the Lifecycle Harness is "a substrate-neutral lifecycle harness". This reverses dependency direction: a core lifecycle surface now depends on one substrate adapter.
  Fix: Move SessionEnd fallback dispatch behind a substrate-neutral lifecycle extension/adapter callback, with Claude Code registering its transcript fallback.

### HonestOracle

- **[important]** `src/adapters/claude-code/session-digest.ts:146` — **Truncated prompts called verbatim**
  The comment says "Prompt lines are VERBATIM principal input," but `cleanLine` collapses control characters/whitespace and truncates text before those lines are written. That mislabels a lossy digest pointer as an exact transcript excerpt, which matters for downstream memory trust.

- **[suggestion]** `src/types.ts:2529` — **Provenance contract omits value**
  The public option docs say "Optional provenance override (`conversation` | `import` | `tool:<name>`)," but the implementation accepts `consolidation` as well. That makes the exported API contract narrower than the real accepted values.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `PR description:1` — **Sub-agent suppression is overstated**
  The description claims "Sub-agent sessions suppressed (ADR 0014; forcePrimary/forceSubagent + env overrides)", but the diff only suppresses when `agent_id` or `agent_type` is present or an env override is set. This misleads downstream readers into believing sub-agent sessions are categorically suppressed, while `hasAgentMarker` admits unmarked sub-agents fall through to the write path.


---

jcfischer (commented):

### Security

- **[important]** `src/adapters/claude-code/session-digest.ts:128` — **Tool names enter digest unescaped**
  Transcript-derived tool names are pushed with `if (typeof name === "string") tools.push(name);` and emitted as ``.map(([name, n]) => `${name}×${n}`)`` without the quoting/control collapse used for prompts. A malicious or malformed tool name can inject directive-like text into trusted recall memory via the `- tools:` line.
  Fix: Sanitize or JSON-quote tool names before adding them to the digest rollup.

### ContextDrift

- **[important]** `src/types.ts:2269` — **Bare agent leaks into lifecycle API**
  The public lifecycle options introduce `agentId?: string;` and `agentType?: string;`, and the CLI exposes matching `--agent-id` / `--agent-type` flags. CONTEXT.md:556-566 says bare `agent` is banned in Soma docs and code and names `Claude Code sub-agent` as the qualified term.
  Fix: Rename the public fields and flags to qualified Claude Code sub-agent terms, or keep these markers adapter-owned instead of core lifecycle API.

### Performance

- **[suggestion]** `src/adapters/claude-code/session-digest.ts:92` — **Avoid repeated lowercase allocation**
  Inside the transcript loop, `PROMPT_NOISE_PREFIXES.some((p) => text.toLowerCase().startsWith(p))` lowercases and allocates a new string once per prefix check for every prompt line. On large transcripts this does repeated avoidable work in the parser hot path.
  Fix: Compute `const lowerText = text.toLowerCase();` once, then use `lowerText.startsWith(p)`.

### HonestOracle

- **[important]** `test/memory-episodic.test.ts:350` — **Lifecycle test proves a primed path**
  The test claims "lifecycle session-end writes the deterministic fallback digest for claude-code" and the comment says it is "proving it was written", but this file directly imports "../src/adapters/claude-code/session-digest" first, which performs registration as a side effect. That is a surrogate endpoint: it proves lifecycle dispatch after test setup has primed the registry, not that the exported lifecycle API writes the fallback for claude-code on its own.

- **[suggestion]** `src/skills/Memory/Workflows/Remember.md:55` — **Strictly better is unsupported**
  The doc states "Self-authoring is strictly better", but the diff only establishes different mechanics and fallback limits, not universal quality superiority. Downstream agents may treat that as proven policy rather than an author preference.


---

jcfischer (commented):

### HonestOracle

- **[important]** `src/skills/Memory/Workflows/Remember.md:55` — **Sub-agent guarantee contradicts implementation**
  The workflow says the fallback "writes nothing for a sub-agent session (ADR 0014)", but the implementation only checks explicit `subagentId`/`subagentType` markers and documents that "an unmarked sub-agent is not detected". This misleads operators into believing all sub-agent sessions are suppressed when only marked ones are.

- **[suggestion]** `src/adapters/claude-code/session-digest.ts:112` — **Tool rollup injection claim overstates escaping**
  The comment claims tool names are control-collapsed "so it can't inject directive text into the `- tools:` rollup line", but `cleanLine` does not quote, delimit, or escape ordinary directive text. A tool name like `ignore previous memory` would still be written as prose in the trusted digest line.

- **[suggestion]** `src/adapters/claude-code/session-digest.ts:70` — **Non-JSON skip hides corruption**
  The comment says a non-JSON line is skipped to "never fail the whole extraction", and later documentation calls the digest deterministic, but silently dropping malformed transcript lines can produce a digest from a partial transcript with no indication that input was corrupted. That is a silent failure mode for anyone relying on the fallback as a session record.


---

jcfischer (commented):

### Architecture

- **[important]** `src/types.ts:2530` — **Hook leaks into core memory API**
  The core digest API adds `hook?: string;` and describes it as an `Optional `hook:` frontmatter marker`, making hook vocabulary part of the persisted/public memory surface. CONTEXT.md §runtime policy inspection says hooks are substrate-native projection mechanisms, and docs/architecture.md §Lifecycle Harness says Soma ports hooks as a substrate-neutral lifecycle harness instead of copying hook files.
  Fix: Rename the core field to lifecycle-neutral metadata such as `lifecycleEvent`, and keep hook-specific naming inside adapters.

### HonestOracle

- **[important]** `src/adapters/claude-code/session-digest.ts:17` — **Thin sessions are not covered**
  The comment says "below this it skips (the assistant self-authors small sessions instead)", but the fallback only returns undefined below six prompts and does not prove or enforce that an assistant-authored digest exists. That misleads downstream readers into treating skipped thin sessions as covered rather than dropped.

- **[suggestion]** `test/memory-episodic.test.ts:352` — **Source spawn overclaims exported CLI**
  The test comment says "This proves the exported CLI path writes the fallback end to end", but the command runs `bun run src/cli.ts`, not an installed/exported package binary. The measured endpoint is the source CLI entrypoint, so the claim overstates what the test proves.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/lifecycle.ts:797` — **Expose digest artifact path**
  The fallback result is collapsed to `digestNote = ` | digest: ${fallback.outcome}`;`, while the handler type is only `Promise<{ outcome: string }>`; after a fallback write, lifecycle cannot include the digest file in its returned `files` list. That makes CLI file reporting incomplete for a session-end that wrote memory.
  Fix: Return `digest?.path` from the handler and append it to lifecycle `files` when present.

### Architecture

- **[important]** `src/adapters/claude-code/session-digest.ts:245` — **Transcript writeback bypasses gate**
  The adapter writes substrate-originated transcript material directly via `const digest = await writeSessionDigest({` with `provenance: "tool:claude-session-end"`; CONTEXT.md's writeback section says an adapter writes back through the writeback gate into Soma. This creates a second writeback path outside the declared Policy boundary.
  Fix: Route the fallback through a writeback-gated core API, or define/document session digests as that governed writeback path.

### HonestOracle

- **[suggestion]** `src/types.ts:2522` — **Lifecycle marker neutrality is overstated**
  The comment says `"Neutral opaque string — no substrate hook vocabulary enters this core API"`, but the same field explicitly documents `"e.g. `session-end`"` and persists it as the note hook. That misleads readers into treating a lifecycle-facing core API as vocabulary-neutral when it already carries lifecycle hook semantics.


---

jcfischer (commented):

2 finding(s): 2 important.


### HonestOracle

- **[important]** `src/lifecycle.ts:781` — **Fallback blocking is overstated**
  The comment says "Best-effort: never blocks session end", but the code awaits `transcriptHandler(...)` before completing lifecycle output. That means transcript read/parse/write latency still blocks this lifecycle command; only thrown failures are swallowed.

- **[important]** `src/types.ts:2525` — **Hook absence does not prove authorship**
  The doc says "absent ⇒ assistant-authored", but `provenance` is an independent option and can mark a non-assistant source without setting `lifecycleEvent`. Downstream readers relying on hook absence would misclassify authorship.


---

jcfischer (commented):

### HonestOracle

- **[important]** `PR description:5` — **Verification claim lacks evidence**
  The description claims "bunx tsc --noEmit clean; 27 episodic + lifecycle + install + projection suites green (71); hook-runner node --check passes; CLI e2e confirms lifecycle writes the fallback" but provides no command output, CI link, or log excerpt. This sells local verification as established fact without reproducible evidence in the PR data.

- **[suggestion]** `PR description:3` — **Sub-agent marker names drift**
  The description says suppression happens when the hook payload carries "agentId/agentType", but the diff reads Claude Code payload keys as "agent_id"/"agent_type" in `hook-runner.mjs`. A downstream implementer relying on the description could wire the wrong marker names.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/adapters/claude-code/session-digest.ts:260` — **Recall disclosure is overstated**
  The comment claims "recall's trust banner shows this body was machine-extracted from principal input, NOT assistant-authored." The recall banner only surfaces the raw provenance token, so downstream readers are being told there is an explicit authorship disclosure that the UI does not actually provide.


---

jcfischer (commented):

1 finding(s): 1 important.


### Security

- **[important]** `src/adapters/claude-code/session-digest.ts:235` — **Hook path can drive arbitrary reads**
  The fallback reads a hook-controlled path directly: `const transcript = await readFile(options.transcriptPath, "utf8").catch(() => undefined);`. A forged or confused SessionEnd payload can point `--transcript` at any readable local file shaped as JSONL user messages, causing selected contents to be persisted into recall memory.
  Fix: Resolve and validate the transcript path against Claude Code's expected transcript/session directory before reading.


---

jcfischer (commented):

1 finding(s): 1 suggestion.


### HonestOracle

- **[suggestion]** `src/skills/Memory/Workflows/Remember.md:55` — **Silent skip claim is false**
  The workflow says the fallback "skips silently if the transcript is unreadable or too thin", but src/lifecycle.ts appends `digest: ${fallback.outcome}` to the lifecycle summary when the handler returns `unreadable` or `skipped`. This misleads readers about the observable audit trail for failed fallback attempts.


---

PR #405, opened 2026-07-04
feat(memory): soma memory backfill — bulk legacy-markdown importer (M8)

jcfischer:
## What

Ships `soma memory backfill` — a first-class, reusable command that turns a principal's existing free-form markdown memory into schema-valid, governed notes. The on-ramp missing for adopting Soma with a non-empty legacy store (a freshly-migrated store has an empty `semantic/`+`procedural/` corpus but a pile of pre-M0 markdown in `LEARNING/`, `KNOWLEDGE/`, …).

```
soma memory backfill [--from <dir>] [--type semantic|procedural] [--project <key>] [--dry-run]
```

## Design

- **Batch caller of the M1 write path** (`writeMemoryNote` + `import` trigger) — not a new governed path, so schema serialization, recall-first dedup, and event journaling are enforced in one place.
- **Markdown only** (`.md`/`.markdown`, read with `O_NOFOLLOW` + dev/ino re-check); reserved dirs, READMEs (case-insensitive), non-markdown, and symlinks (entries and the source root) are skipped/refused. Root-level files are skipped only for the default memory root; a custom `--from <dir>` imports its top-level markdown (category `""`→semantic).
- **Category→type mapping**: `LEARNING`→procedural, `KNOWLEDGE`→semantic, else semantic; `--type` forces flat (honoured on rerun — a type-switch is not a silent manifest hit).
- **`import` trigger → `quarantined` trust** for everything (MINJA-safe). Imports are recall-discoverable with a ⚠ untrusted banner but held out of the always-loaded INDEX until re-authored at higher trust — the admission filter excludes quarantined unconditionally, so `verify` (freshness only) cannot promote them; `principal-correction`/`supersede` can. (recall includes quarantined `memory-recall.ts:136,162`; INDEX excludes them `memory-index.ts:17`.)
- **Deterministic**: bodies imported verbatim (no preamble — it would inflate dedup similarity), `created` from source mtime, no LLM.
- **Idempotent** via a SHA manifest at `memory/STATE/imports/backfill/.manifest.json` (inside the Memory compartment, keyed to the source root). A no-op rerun re-emits prior entries verbatim → byte-stable even across a `touch`. Rebuilds INDEX after writes so the store stays audit-clean. A single bad source (unreadable/deleted/swapped) becomes a per-file `error` entry, never a batch abort.

## Verification

Reproduce on this branch:

```
$ bun run typecheck        # tsc --noEmit → exit 0, clean
$ bun test test/memory-backfill.test.ts   # 21 pass / 0 fail
$ bun test                 # full suite → 1744 pass / 0 fail across 112 files
```

Live acceptance on a throwaway `--soma-home` fixture: notes land in `semantic/`/`procedural/` with `trust: quarantined`, `provenance: import`, `created` = source mtime, `source_of_truth` = origin; `soma memory recall <term>` shows a `⚠ QUARANTINED (untrusted)` banner; `soma memory audit` → HEALTHY (backfill rebuilds INDEX); rerun → `0 written`, byte-identical manifest; a second identical-body file → `skipped-duplicate` (recall-first refusal), not an error.

## Out of scope (v1)

LLM distillation (deterministic-only), auto-supersede on source drift, trust elevation of imports (the principal re-authors later via `principal-correction`/`supersede`), non-markdown / recall-SQLite sources.

### DIFF
--- a/Plans/declarative-twirling-lampson.md
+++ b/Plans/declarative-twirling-lampson.md
@@ -52,10 +52,13 @@ in the source dir).
 
 **Skipped, never imported:** reserved dirs `STATE`, `episodic`, `semantic`,
 `procedural`, `archive`, `imports`; any `README.md` (case-insensitive);
-non-markdown files; symlinks anywhere (loud refusal, matching the migrator's
-stance). Files sitting directly under the root are skipped **only for the
-default `<somaHome>/memory` root** (README/`INDEX.md` territory); a custom
-`--from <dir>` imports its top-level markdown (category `""` → semantic).
+non-markdown files. **Symlink handling** (not a whole-run guarantee): the source
+root and every entry the walk returns are lstat-refused if a symlink; the leaf
+read then uses `O_NOFOLLOW` and re-checks the opened file's `dev`+`ino` against
+the scan, so a leaf- or ancestor-swap between scan and read is refused at read
+time. Files sitting directly under the root are skipped **only for the default
+`<somaHome>/memory` root** (README/`INDEX.md` territory); a custom `--from <dir>`
+imports its top-level markdown (category `""` → semantic).
 
 ### Per-file note synthesis (deterministic)
 
@@ -77,7 +80,7 @@ For each eligible source file at relative path `rel`:
 
 ### Result / reporting
 
-`runMemoryBackfill` returns `{ somaHome, from, dryRun, writtenCount, skippedManifestCount, skippedDuplicateCount, errorCount, manifestPath, entries }` (each `entries[]` item carries `relativePath`, `noteId`, `type`, `created`, `target`, and a per-file `status`); CLI prints a summary table. `--dry-run` prints the plan without writing or touching the manifest.
+`runMemoryBackfill` returns `{ somaHome, from, dryRun, writtenCount, skippedManifestCount, skippedDuplicateCount, errorCount, manifestPath, entries }` (each `entries[]` item carries `relativePath`, `noteId`, `type`, `created`, `target`, and a per-file `status`); CLI prints a summary table. `--dry-run` previews without writing or touching the manifest/INDEX — it is manifest-aware (already-imported files show as `skipped-manifest`) but does **not** run the corpus scan, so a would-import entry may still turn into a `skipped-duplicate` at a real run.
 
 ## Files to create / modify
 
--- a/Plans/declarative-twirling-lampson.md
+++ b/Plans/declarative-twirling-lampson.md
@@ -16,7 +16,7 @@ M1 trust-governance model. This is milestone **M8 — backfill**.
 
 **Key design facts established during exploration:**
 - The `import` write trigger is the sanctioned bulk path: `SOMA_MEMORY_TRIGGER_TRUST.import → "quarantined"`, needs **no authority signal** (`memory-write.ts:30`, MINJA defense). Backfilled content lands untrusted by design.
-- Lifecycle is correct, not a dead end: **recall INCLUDES quarantined notes** with a ⚠ untrusted banner (`memory-recall.ts:136,162` — only `valid_until===null` is filtered, not trust), while **INDEX EXCLUDES quarantined** (`memory-index.ts:17`). So imports are pull-discoverable immediately and earn always-on INDEX only after the principal verifies/elevates them.
+- Lifecycle is correct, not a dead end: **recall INCLUDES quarantined notes** with a ⚠ untrusted banner (`memory-recall.ts:136,162` — only `valid_until===null` is filtered, not trust), while **INDEX EXCLUDES quarantined** (`memory-index.ts:17`). So imports are pull-discoverable immediately and earn always-on INDEX only after the principal re-authors them at higher trust (the admission filter excludes quarantined unconditionally, so `verify` — freshness only — cannot promote them; `principal-correction`/`supersede` can).
 - `writeMemoryNote({mode:"create", trigger:"import", …})` (`memory-write.ts:779`) already does schema serialization, per-note dedup (recall-first refusal, Jaccard ≥0.6 / exact-body), and event journaling. Backfill is a **batch importer over this existing path**, not a new writer.
 - Idempotency pattern to mirror: `src/pai-memory-migrator.ts` (SHA manifest, skip-unchanged, byte-stable no-op rerun, symlink refusal).
 
--- a/Plans/declarative-twirling-lampson.md
+++ b/Plans/declarative-twirling-lampson.md
@@ -65,21 +65,21 @@ For each eligible source file at relative path `rel`:
 - **source_of_truth** = absolute path of the original file (so a human can verify against the archived original).
 - **project** = `--project` or `null`; **links** = `[]`; **resurface_count** = 0.
 - **hook** = short humanized recall phrase from the stem (optional; aids recall matching).
-- **body** = original file content, prefixed with a one-line blockquote preamble noting the backfill origin.
+- **body** = original file content **verbatim** — no injected preamble (a shared preamble inflates token overlap and makes the recall-first dedup over-fire on short files; origin lives in frontmatter instead). Only `.md`/`.markdown` files are imported.
 
 ### Dedup & idempotency
 
 - Files processed **sequentially** (not concurrent) so later files see earlier-written ones and the recall-first refusal dedups within the batch deterministically.
 - A per-file **recall-first refusal** (exact/near-dup already in corpus) is caught and counted as `skipped-duplicate`, not a batch abort (`memory-write.ts:618` builds the refusal — implementation inspects that surface to classify).
-- Manifest `<somaHome>/imports/backfill/.manifest.json` (schema `soma.memory-backfill.v1`, entries `{relativePath, noteId, type, sha256, mtimeMs}`). Rerun skips files whose source SHA matches AND whose target note still exists → byte-stable no-op. Source drift in v1 = re-import as a new note (documented limitation; no auto-supersede yet).
+- Manifest `<somaHome>/memory/STATE/imports/backfill/.manifest.json` (schema `soma.memory-backfill.v1`, entries `{relativePath, noteId, type, sha256, mtimeMs}`) — inside the Memory compartment, under the reserved STATE dir the source walk never re-imports. Rerun skips a file only when its source SHA matches, its resolved type matches the prior import, AND its target note still exists; the prior entry is re-emitted verbatim → byte-stable no-op (even across a `touch`). Source drift = re-import as a new note (documented limitation; no auto-supersede yet).
 
 ### Result / reporting
 
-`runMemoryBackfill` returns `{ writtenCount, skippedDuplicateCount, skippedManifestCount, errors[], notes[] }`; CLI prints a summary table. `--dry-run` prints the same plan without writing or touching the manifest.
+`runMemoryBackfill` returns `{ somaHome, from, dryRun, writtenCount, skippedManifestCount, skippedDuplicateCount, errorCount, manifestPath, entries }` (each `entries[]` item carries `relativePath`, `noteId`, `type`, `created`, `target`, and a per-file `status`); CLI prints a summary table. `--dry-run` prints the plan without writing or touching the manifest.
 
 ## Files to create / modify
 
-1. **`src/memory-backfill.ts`** (new, ~250 LOC) — `planMemoryBackfill()` + `runMemoryBackfill()`; category map, file walk (skip/symlink rules), id derivation + collision handling, mtime→date, body preamble, manifest read/render, sequential loop calling `writeMemoryNote`. Reuses the write path (serialization, dedup, events) rather than re-implementing note I/O.
+1. **`src/memory-backfill.ts`** (new, ~250 LOC) — `planMemoryBackfill()` + `runMemoryBackfill()`; category map, markdown-only file walk (skip/symlink rules), id derivation + collision handling, mtime→date, verbatim body, manifest read/render, sequential loop calling `writeMemoryNote`. Reuses the write path (serialization, dedup, events) rather than re-implementing note I/O.
 2. **`src/types.ts`** — add `SomaMemoryBackfillOptions`, `SomaMemoryBackfillPlanEntry`, `SomaMemoryBackfillResult`, manifest types, and `SOMA_MEMORY_BACKFILL_TYPE_MAP` constant (near the other memory-subsystem types).
 3. **`src/cli/memory.ts`** — add `"backfill"` to `MEMORY_ACTIONS`; `ParsedMemoryBackfillArgs` + union member; `parseMemoryBackfillArgs`; a `MEMORY_COMMAND_HELP.subcommands.backfill` usage string; dispatch `case "backfill"`; handler calling `runMemoryBackfill` + a `formatMemoryBackfillResult`.
 4. **`test/memory-backfill.test.ts`** (new) — mirror `pai-memory-migrator.test.ts` + `memory-write.test.ts` patterns: category→type mapping, id derivation + collision, mtime→created, trust=quarantined, dedup skip, `--dry-run` no-op, idempotent rerun (byte-stable manifest), symlink refusal, reserved-dir skipping.
--- a/Plans/declarative-twirling-lampson.md
+++ b/Plans/declarative-twirling-lampson.md
@@ -1,0 +1,105 @@
+# Plan: `soma memory backfill` — bulk legacy-content importer (M8)
+
+## Context
+
+The note-based Memory subsystem (M0–M7) shipped, but on a freshly-migrated store
+the **durable corpus is empty**: `memory/` has no `semantic/` or `procedural/`
+directory — only auto-generated `episodic/` digests plus **legacy pre-M0
+content** in free-form category dirs (`LEARNING/`, `KNOWLEDGE/`, `WORK/`, …).
+Legacy files can even carry a stopgap banner: *"re-promote via `soma memory
+write` once M1 ships."*
+
+For **shipping Soma to other users**, we need a first-class, reusable command
+that takes a user's existing markdown memory and turns it into schema-valid,
+governed notes — deterministically, no LLM (subsystem invariant), respecting the
+M1 trust-governance model. This is milestone **M8 — backfill**.
+
+**Key design facts established during exploration:**
+- The `import` write trigger is the sanctioned bulk path: `SOMA_MEMORY_TRIGGER_TRUST.import → "quarantined"`, needs **no authority signal** (`memory-write.ts:30`, MINJA defense). Backfilled content lands untrusted by design.
+- Lifecycle is correct, not a dead end: **recall INCLUDES quarantined notes** with a ⚠ untrusted banner (`memory-recall.ts:136,162` — only `valid_until===null` is filtered, not trust), while **INDEX EXCLUDES quarantined** (`memory-index.ts:17`). So imports are pull-discoverable immediately and earn always-on INDEX only after a human verifies/elevates them.
+- `writeMemoryNote({mode:"create", trigger:"import", …})` (`memory-write.ts:779`) already does schema serialization, per-note dedup (recall-first refusal, Jaccard ≥0.6 / exact-body), and event journaling. Backfill is a **batch adapter over this existing path**, not a new writer.
+- Idempotency pattern to mirror: `src/pai-memory-migrator.ts` (SHA manifest, skip-unchanged, byte-stable no-op rerun, symlink refusal).
+
+## Approach
+
+New subcommand:
+
+```
+soma memory backfill [--from <dir>] [--type semantic|procedural]
+                     [--project <key>] [--dry-run]
+                     [--soma-home <dir>] [--home-dir <dir>]
+```
+
+- **`--from`** (default `<somaHome>/memory`): source root walked recursively.
+- **`--type`**: force ALL notes to one type, overriding the category map.
+- **`--project`**: set note `project` field (default `null`).
+- **`--dry-run`**: print the plan table, touch nothing.
+
+### Category → type mapping (the chosen model)
+
+Map by the **top-level category dir** (first path segment under the source root):
+
+| Category dir | Note type |
+|---|---|
+| `LEARNING` (incl. `ALGORITHM/`, `PROMOTED/`, `REFLECTIONS/`) | `procedural` (how-to / behavioral lessons) |
+| `KNOWLEDGE` (incl. `PROMOTED/`, `Research/`) | `semantic` (facts) |
+| any other category | `semantic` (fallback) |
+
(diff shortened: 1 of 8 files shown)


jcfischer (commented):

8 finding(s): 5 important, 2 suggestion, 1 nit.


### CodeQuality

- **[important]** `src/memory-backfill.ts:160` — **Backfill imports non-markdown files**
  The source walk accepts any regular file with `if (!entry.isFile()) continue;` and only excludes `if (entry.name === "README.md") continue;`, so files like JSON, images, or editor artifacts under a category directory are read as UTF-8 and imported despite the command being documented as markdown-only. This can create malformed or meaningless memory notes from unrelated files.
  Fix: Require a markdown extension before pushing the source, e.g. skip files that do not end with `.md` or `.markdown`.

### Security

- **[important]** `src/memory-backfill.ts:162` — **Imports non-markdown files as memory notes**
  The diff documents a legacy-markdown importer, but the walker accepts every regular file: `if (!entry.isFile()) continue;` followed by `files.push({ ... })`, with no `.md` extension check. A category directory containing `.env`, key material, exports, or other non-markdown files would be read by `const content = await readFile(src.absPath, "utf8");` and written into recall-discoverable memory notes.
  Fix: Only import `.md`/`.markdown` files, or require an explicit opt-in flag for non-markdown sources.

### Architecture

- **[nit]** `docs/architecture.md:127` — **Use principal in formal docs**
  CONTEXT.md#principal says formal docs should use `principal`, not `user`; the added architecture text says "for a migrating user". This keeps Soma's authority/ownership vocabulary consistent in the architecture contract.
  Fix: Change to `for a migrating principal`.

### Performance

- **[important]** `src/memory-backfill.ts:399` — **Manifest hit path rereads manifest per file**
  On every unchanged source, `noteExistsForRelative` executes `const parsed = JSON.parse(await readFile(manifestPath, "utf8")) as SomaMemoryBackfillManifest;` and then `parsed.files.find((f) => f.relativePath === relativePath)`. A no-op rerun over N already-imported files therefore does N manifest reads/parses plus O(N²) linear searches, directly slowing the advertised idempotent bulk path.
  Fix: Have `readManifest` return a `Map<string, manifestEntry>` and use that cached record in the loop.

### Maintainability

- **[suggestion]** `src/memory-backfill.ts:177` — **Keep full manifest records in memory**
  `readManifest` reduces entries to SHA only with `for (const entry of parsed.files) map.set(entry.relativePath, entry.sha256);`, then `noteExistsForRelative` reparses the same file and searches `parsed.files.find((f) => f.relativePath === relativePath)`. This splits one concept across two shapes and makes manifest-field changes require edits in both places.
  Fix: Have `readManifest` return `Map<string, SomaMemoryBackfillManifestEntry>` and use that record directly for SHA, note id, and type.

### HonestOracle

- **[important]** `src/cli/memory.ts:181` — **Markdown-only claim is false**
  The help says "Bulk-import legacy free-form markdown", but `collectSources` admits every non-root file after only `if (!entry.isFile()) continue; if (entry.name === "README.md") continue;`. Downstream users relying on the documented markdown boundary can unintentionally import non-markdown files under category dirs.

- **[important]** `src/memory-backfill.ts:26` — **Byte-stable rerun is overclaimed**
  The comment claims "A pure no-op rerun writes a byte-identical manifest", but skipped manifest entries are rebuilt with `mtimeMs: src.mtimeMs`. Touching a source file without changing bytes is treated as a SHA no-op yet still changes the manifest bytes.

- **[suggestion]** `src/memory-backfill.ts:14` — **Body preamble claim contradicts code**
  The contract says "Bodies are wrapped verbatim (with a one-line provenance preamble)", while the write call later says "Body is the legacy content VERBATIM — no boilerplate preamble" and passes `body: content`. That leaves reviewers and future maintainers with two incompatible statements about the data contract.


---

jcfischer (commented):

7 finding(s): 4 important, 2 suggestion, 1 nit.


### CodeQuality

- **[important]** `src/memory-backfill.ts:314` — **Manifest skip ignores forced type**
  The manifest hit path uses prior metadata unconditionally: `if (prior?.sha256 === sha) {` followed by `type: prior.type`, so `--type` no longer "Force ALL notes to this type" for unchanged files previously imported with another type. A user rerunning `soma memory backfill --type procedural` after an earlier semantic import gets skipped entries that remain semantic.
  Fix: Only treat a manifest entry as a hit when the effective requested type matches `prior.type`, or document that manifest hits are immutable unless the manifest is removed.

### Architecture

- **[important]** `src/memory-backfill.ts:53` — **Manifest escapes Memory root**
  The diff sets `const MANIFEST_RELATIVE = "imports/backfill/.manifest.json";`, but docs/architecture.md's Memory section says Memory lives under a single lowercase `memory/` root. This creates durable Memory-subsystem state outside the compartment boundary and cements a public file layout that replication/audit will need to special-case.
  Fix: Store it under `memory/STATE/imports/backfill/.manifest.json`.

- **[nit]** `src/types.ts:2492` — **Use principal in source**
  The diff adds `Bulk-import a user's existing free-form markdown memory`; CONTEXT.md's `principal` section bans `user` in `src/` and formal docs. This is a canonical vocabulary violation in exported type comments.
  Fix: Use `principal's`.

### Performance

- **[suggestion]** `src/memory-backfill.ts:300` — **Reruns reread every imported file**
  The loop does `const content = await readFile(src.absPath, "utf8");` and `const sha = sha256Hex(content);` before checking `if (prior?.sha256 === sha)`, so an idempotent rerun still allocates and hashes the full legacy corpus just to skip manifest hits. For a bulk importer, that makes no-op runs scale with total source bytes rather than changed files.
  Fix: Store size/mtime in the manifest and skip hashing when both still match; fall back to SHA only when metadata changed.

### Maintainability

- **[suggestion]** `src/memory-backfill.ts:286` — **Extract per-source import handling**
  `runMemoryBackfill` combines manifest-hit detection, id allocation, note writing, duplicate/error classification, counters, and manifest assembly inside `for (const src of sources) {`, with the central write path starting at `try { await writeMemoryNote({`. That makes future changes to one case require keeping the whole loop in working memory.
  Fix: Extract a helper that processes one SourceFile and returns `{ entry, manifestEntry?, countKind }`.

### HonestOracle

- **[important]** `Plans/declarative-twirling-lampson.md:68` — **Plan claims nonexistent body preamble**
  The plan says "**body** = original file content, prefixed with a one-line blockquote preamble", but the implementation writes `body: content` and the tests assert no "Backfilled from" preamble. This misleads readers about the persisted note body and contradicts the advertised verbatim-import behavior.

- **[important]** `Plans/declarative-twirling-lampson.md:78` — **Plan advertises wrong result shape**
  The plan says `runMemoryBackfill` returns `{ writtenCount, skippedDuplicateCount, skippedManifestCount, errors[], notes[] }`, but the exported result type returns `errorCount` and `entries`, not `errors[]` or `notes[]`. Anyone using the plan as the command contract would depend on fields that do not exist.


---

jcfischer (commented):

### Maintainability

- **[suggestion]** `src/cli/memory.ts:832` — **Deduplicate status section formatting**
  `const written = result.entries.filter((e) => e.status === "written");`, `const dups = result.entries.filter((e) => e.status === "skipped-duplicate");`, and `const errors = result.entries.filter((e) => e.status === "error");` repeat the same filter-plus-heading-plus-loop shape. Adding more statuses will copy this block again.
  Fix: Extract a small `appendBackfillSection(lines, entries, status, heading, format)` helper.

### HonestOracle

- **[important]** `docs/architecture.md:130` — **Manifest path is misdocumented**
  The doc claims idempotency uses “a SHA manifest at `imports/backfill/.manifest.json`”, but the implementation writes `memory/STATE/imports/backfill/.manifest.json`. A downstream operator following the architecture doc will look in the wrong place for the state that controls reruns.

- **[suggestion]** `PR description:23` — **Verification claim lacks evidence**
  The description claims “Full suite **1736 pass / 0 fail**; typecheck clean” and “Live acceptance on a throwaway `--soma-home` fixture”, but the PR data provides no command transcript, CI link, or reproducible output. This sells verification as established fact without evidence in the review packet.


---

jcfischer (commented):

6 finding(s): 1 important, 4 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/memory-backfill.ts:179` — **README skip is case-sensitive**
  The filter only skips `entry.name === "README.md"`, so `README.MD` or `Readme.md` under a category dir would be imported even though the same block documents skipping `any README.md`. This can create an unintended note from a conventional metadata file on case-sensitive filesystems.
  Fix: Use a case-insensitive README check, for example `if (/^readme\.md$/i.test(entry.name)) continue;`.

### Architecture

- **[nit]** `src/types.ts:2496` — **Use principal in source comment**
  CONTEXT.md says `principal` is the only formal docs/src term and not to use `human`; this comment says `a human elevates it later via verify/supersede`. That weakens the public type surface's domain vocabulary.
  Fix: Replace `a human` with `the principal`.

### Maintainability

- **[suggestion]** `src/types.ts:2515` — **Use the WritableNoteType alias**
  The diff introduces `export type WritableNoteType = Exclude<SomaMemoryNoteType, "episodic">;` but then repeats `"semantic" | "procedural"` in `SomaMemoryBackfillOptions`, `SomaMemoryBackfillEntry`, and `SomaMemoryBackfillManifestEntry`. That leaves the alias unused in the type surface it was meant to name, so future type changes require multiple edits.
  Fix: Change those repeated fields to `WritableNoteType`.

### HonestOracle

- **[important]** `Plans/declarative-twirling-lampson.md:74` — **Source drift reimport is overclaimed**
  The plan claims "Source drift = re-import as a new note", but the same section routes "exact/near-dup already in corpus" to `skipped-duplicate`; a small edit changes SHA while still hitting recall-first refusal. That misleads operators into expecting changed sources to be captured as new notes when they can be classified as duplicates instead.

- **[suggestion]** `docs/architecture.md:130` — **Markdown coverage is overstated**
  The docs say backfill "writes each markdown file as a `quarantined` note", but the implementation explicitly skips markdown files named `README.md` via `if (entry.name === "README.md") continue;`. Readers relying on "each markdown file" will expect README markdown under category dirs to import when it is deliberately excluded.

- **[suggestion]** `src/types.ts:2501` — **Manifest path comment drifted**
  The exported type comment says the manifest is at "`imports/backfill/.manifest.json`", while the implementation uses `"memory/STATE/imports/backfill/.manifest.json"`. SDK readers looking at the public types get the wrong compartment-relative location.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-backfill.ts:162` — **Custom category roots import nothing**
  The collector skips every file at depth 0 via `if (depth === 0 && (RESERVED_CATEGORIES.has(entry.name) || entry.isFile()))`, so `soma memory backfill --from <dir>` pointed at a category directory containing `*.md` files imports zero files. This contradicts the exposed `--from <dir>` option because a valid markdown source root can have files directly under it.
  Fix: Only skip root-level files when `--from` is the default memory root, or derive category/type differently for custom roots.

### Security

- **[important]** `src/memory-backfill.ts:309` — **Symlink refusal has a TOCTOU bypass**
  The walker rejects symlinks with `if (entry.isSymbolicLink())`, but the later import reads by path with `const content = await readFile(src.absPath, "utf8");`. An attacker who can modify the source tree between collection and read can replace an accepted `.md` file with a symlink to a sensitive file, bypassing the documented symlink refusal and markdown-only gate.
  Fix: Open/read with no-follow semantics immediately before import, or re-lstat and reject symlinks after opening the exact file.

### Maintainability

- **[suggestion]** `src/cli/memory.ts:822` — **Name the backfill entry type**
  The local formatter repeats the indexed type twice: `status: SomaMemoryBackfillResult["entries"][number]["status"]` and `format: (e: SomaMemoryBackfillResult["entries"][number]) => string`. That type expression is noisy enough to obscure the helper signature and will need duplicate edits if the result shape changes.
  Fix: Add a local alias, e.g. `type BackfillEntry = SomaMemoryBackfillResult["entries"][number];`, then use `BackfillEntry["status"]` and `BackfillEntry`.

### HonestOracle

- **[suggestion]** `src/types.ts:2494` — **Comment implies body wrapping**
  The comment says "bodies are wrapped verbatim", but the implementation writes `body: content` and the PR says "bodies verbatim (no injected preamble)". This misleads API readers into expecting an added wrapper around imported content.


---

jcfischer (commented):

7 finding(s): 2 important, 4 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/memory-backfill.ts:331` — **Skipped entry reports touched mtime**
  The manifest-hit branch reuses the prior note and manifest, but the result entry sets `created,` from the current source stat after `const created = isoDate(new Date(src.mtimeMs));`. After a byte-identical `touch`, the CLI can report a different created date than the already-created target note and the preserved manifest entry.
  Fix: For manifest skips, derive `created` from `prior.mtimeMs` instead of `src.mtimeMs`.

### Security

- **[important]** `src/memory-backfill.ts:409` — **Persists absolute source paths**
  The diff records local paths with `sourceOfTruth: src.absPath`, and `--from <dir>` may point outside the Soma memory tree. Backfilling from sensitive locations can persist usernames, mount names, customer folders, or secret project paths into notes that are later exported, synced, or shared.
  Fix: Store a relative source identifier when possible, or redact/normalize absolute paths outside `<somaHome>/memory` before writing `sourceOfTruth`.

### Architecture

- **[suggestion]** `Plans/declarative-twirling-lampson.md:20` — **Do not call backfill an adapter**
  The plan says Backfill is a "batch adapter over this existing path", but CONTEXT.md's adapter section reserves adapter for one-per-substrate projection actors. Calling a Memory importer an adapter blurs the documented substrate boundary.
  Fix: Rename this to "batch importer" or "batch runner".

### ContextDrift

- **[suggestion]** `Plans/declarative-twirling-lampson.md:20` — **Avoid adapter for batch importer**
  The phrase "batch adapter over this existing path" uses `adapter` for a Memory importer wrapper, while CONTEXT.md:391-400 defines an adapter as the substrate projection actor with one adapter per substrate.
  Fix: Use "batch wrapper over this existing path" or "batch importer over this existing path".

- **[nit]** `docs/architecture.md:136` — **Use principal in formal docs**
  The phrase "until a human verifies them" uses `human` in formal docs, while CONTEXT.md:354-364 reserves `principal` for this role and lists `human` as an avoided synonym in docs/code contexts.
  Fix: Change to "until the principal verifies them".

### Maintainability

- **[suggestion]** `src/memory-backfill.ts:266` — **Deduplicate backfill planning loop**
  `planMemoryBackfill` and `runMemoryBackfill` both repeat the setup and mapping shape: `const defaultRoot = resolve(join(somaHome, "memory"));`, `const from = resolve(options.from ?? defaultRoot);`, `const sources = await collectSources(from, from === defaultRoot);`, then per-source `resolveType`, `uniqueNoteId`, `created`, and `target` construction. This makes future changes to note id/type/target derivation easy to apply in one path and miss in the other.
  Fix: Extract a shared `buildBackfillPlan(options, somaHome, from, sources)` helper and let dry-run return it while the real run consumes it.

### HonestOracle

- **[important]** `Plans/declarative-twirling-lampson.md:53` — **Root-file skip is false**
  The plan claims "Skipped, never imported:" includes "files directly under the root", but `src/memory-backfill.ts` explicitly imports root-level markdown for a custom `--from <dir>` and the test asserts "Root-level file is imported". The skip rule is conditional, not universal, so the documented safety boundary is wider than the code enforces.


---

jcfischer (commented):

1 finding(s): 1 suggestion.


### CodeQuality

- **[suggestion]** `src/memory-backfill.ts:210` — **README.markdown is imported**
  The filter only skips `"/^readme\.md$/i.test(entry.name)"`, while markdown eligibility accepts `"/\.(?:md|markdown)$/i"`; this means `README.markdown` is imported as a note even though README files are documented as skipped.
  Fix: Skip README with both markdown extensions, e.g. `/^readme\.(?:md|markdown)$/i`.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-backfill.ts:348` — **Manifest skips across source roots**
  The manifest hit is keyed only by relative path via `const prior = previous?.map.get(src.relativePath);`. Because the manifest also records `from` but that value is not checked, running backfill against a different `--from` with the same relative path/SHA/type can silently skip a requested import.
  Fix: Only reuse prior entries when `previous.from === from`, or key manifest hits by source root plus relative path.

### Security

- **[important]** `src/memory-backfill.ts:100` — **Ancestor symlink race still imports redirected files**
  The code claims the read closes TOCTOU, but `fh = await open(path, FS.O_RDONLY | FS.O_NOFOLLOW);` only rejects a symlink in the final component. After `collectSources` vets a real directory, an attacker who can write under `--from` can replace an ancestor directory with a symlink before `readSourceNoFollow(src.absPath)`, redirecting the importer to unintended `.md` content.
  Fix: Open/verify each path component without following symlinks, or re-lstat and compare ancestor device/inode immediately before reading.

### Performance

- **[suggestion]** `src/memory-backfill.ts:317` — **Preload note ids before backfill**
  Each imported source calls `const id = await uniqueNoteId(somaHome, slugifyId(src.category, src.stem), used);`, and `uniqueNoteId` probes both note stores via `noteExists` for every candidate. On large backfills this adds two serial filesystem stats per file before any write-path work starts.
  Fix: Build a Set of existing semantic/procedural note ids once, then resolve collisions against that Set plus `used`.

### HonestOracle

- **[suggestion]** `src/memory-backfill.ts:247` — **Corrupt manifest handling is overstated**
  The comment claims, "Returns null when the manifest is missing or corrupt (→ re-import)," but the implementation only rejects unreadable JSON or a non-array `files`; it accepts any object with a `files` array regardless of schema, entry shape, or `importedAt`. That misleads readers into believing malformed manifests fail closed into re-import when some semantically corrupt manifests can still drive skip behavior.


---

jcfischer (commented):

1 finding(s): 1 important.


### CodeQuality

- **[important]** `src/memory-backfill.ts:346` — **Source read errors bypass per-file accounting**
  The loop reads each source before the per-file try block: "const content = await readSourceNoFollow(src.absPath, { dev: src.dev, ino: src.ino });" while the catch only wraps "await writeMemoryNote({". If a file is deleted, becomes unreadable, or changes identity between scan and read, the whole backfill rejects instead of returning an entry with status "error" and incrementing errorCount.
  Fix: Move readSourceNoFollow, sha calculation, and created/type entry setup inside the per-file try, or add a separate catch that records status "error" before continuing.


---

jcfischer (commented):

### Security

- **[important]** `src/memory-backfill.ts:289` — **Source root symlink is followed**
  The diff claims symlinks are refused, but `const from = resolve(options.from ?? defaultRoot);` is passed straight into `collectSources(from, from === defaultRoot)` and `collectSources` starts with `readdir(dir, { withFileTypes: true })`, which follows a symlinked root before any `entry.isSymbolicLink()` check can run. A symlinked `<somaHome>/memory` or `--from` can therefore import markdown from an unintended target tree despite the stated refusal policy.
  Fix: `lstat(from)` before traversal and fail if the source root is a symlink.

### Architecture

- **[suggestion]** `docs/architecture.md:126` — **Update governed-path taxonomy**
  The diff adds `soma memory backfill` that `writes each markdown file ... through the governed write path`, but the same Memory section still says `One other governed path exists: soma memory consolidate`, making the architecture contract internally inconsistent.
  Fix: Revise the Memory section to list backfill as a governed import path, or clarify it is only a batch caller of `write|verify` rather than an additional governed path.

### Maintainability

- **[suggestion]** `src/memory-backfill.ts:211` — **Extract source eligibility decisions**
  `if (entry.isSymbolicLink())`, `if (depth === 0 && RESERVED_CATEGORIES.has(entry.name)) continue;`, `if (depth === 0 && skipRootFiles) continue;`, `if (/^readme\.(?:md|markdown)$/i.test(entry.name)) continue;`, and `if (!MARKDOWN_EXT.test(entry.name)) continue;` pack refusal, traversal, and import-eligibility policy into one branch ladder. That makes the walk harder to change because adding a rule requires reasoning through traversal side effects and import filters together.
  Fix: Extract a small `classifySourceEntry(...)` helper returning `refuse`, `descend`, `skip`, or `import`.

### HonestOracle

- **[suggestion]** `Plans/declarative-twirling-lampson.md:102` — **Audit freshness claim is unsupported**
  The plan claims '`soma memory audit` passes (schema-valid, INDEX fresh)' for live acceptance, but this diff does not show backfill rebuilding or refreshing INDEX after writes. That sells audit/INDEX health as verified behavior while the implementation only writes notes and a manifest.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/memory-backfill.ts:244` — **Scan errors bypass per-file accounting**
  The scan phase does `const stat = await lstat(abs);` before the run loop's per-file `try`, so a file deleted or made inaccessible between `readdir` and `lstat` aborts the whole backfill instead of producing an `error` entry.
  Fix: Catch `lstat` failures in `collectSources` and either skip vanished files or return an error entry for the run result.

### HonestOracle

- **[important]** `src/cli/memory.ts:846` — **Verify cannot elevate quarantined imports**
  The CLI prints "imported (quarantined — recall-discoverable, elevate via verify to reach INDEX):", but `verifyMemoryNote` only updates freshness/resurface count and the index admission path rejects `trust === "quarantined"` unconditionally. A user following this output will believe verification moves imported notes into INDEX when quarantined notes cannot reach INDEX by verify alone.


---

jcfischer (commented):

2 finding(s): 2 important.


### CodeQuality

- **[important]** `src/memory-backfill.ts:361` — **Dry-run ignores manifest state**
  `if (options.dryRun) { return planMemoryBackfill(options); }` bypasses the manifest-aware skip logic, while the CLI reports `would import ${result.entries.length} file(s):`. After a completed backfill, `--dry-run` can show already-imported files as new imports with suffixed ids even though a real run would skip them.
  Fix: Make dry-run apply the same manifest checks and return skipped-manifest entries without writing.

### HonestOracle

- **[important]** `src/types.ts:2500` — **Verify promotion claim contradicts lifecycle**
  The comment says imported content is quarantined and "the principal elevates it later via verify/supersede", but the PR description, README, and architecture text all state that verify cannot promote quarantined imports. A developer relying on this type-level contract could wire trust elevation to the wrong operation.


---

jcfischer (commented):

### HonestOracle

- **[important]** `Plans/declarative-twirling-lampson.md:59` — **Symlink refusal is overclaimed**
  The plan claims "symlinks anywhere (loud refusal)", but the implementation only checks entries returned during the initial walk and then rechecks file identity at read time; it does not prove every ancestor remains non-symlink for the whole run. This misleads readers into treating the backfill as a comprehensive symlink-refusal boundary.

- **[suggestion]** `Plans/declarative-twirling-lampson.md:87` — **Dry-run reporting overstates certainty**
  The plan says "--dry-run prints the plan without writing", while src/memory-backfill.ts explicitly says near-duplicate detection is "NOT part of the plan" and a real run may skip would-import files as duplicates. Calling it "the plan" hides that the preview is incomplete for dedup outcomes.


---

PR #414, opened 2026-07-05
Fix Codex session digest capture

### LINKED ISSUE #413: Wire automatic Codex session digest capture
## Problem

Soma's Codex projection already installs a Codex `Stop` hook that runs `bun ~/.codex/hooks/soma-lifecycle.mjs session-end`, but this does not currently produce an automatic Soma memory digest.

The current Codex lifecycle hook calls `soma lifecycle session-end --substrate codex` without a transcript path. The Soma core only writes an automatic deterministic session digest when `runSomaLifecycleSessionEnd` receives both `sessionId` and `transcriptPath` and finds a registered `SessionEndTranscriptHandler` for the substrate. Today that transcript digest handler is registered for `claude-code`, not for `codex`.

As a result, Codex sessions get metadata lifecycle capture, but not the equivalent deterministic transcript-to-`soma memory digest` fallback. The Codex projection documents this as a capture limitation and tells the agent to run `soma memory digest` manually.

## Evidence

- `src/adapters/codex/adapter.ts` installs a `Stop` hook that runs `bun ~/.codex/hooks/soma-lifecycle.mjs session-end`.
- `src/adapters/codex/hooks/codex-hook-entry.mjs` handles `session-end` by calling `runSomaLifecycle(config, event, input.session_id)`, which passes no `--transcript` and no `--cwd`.
- `src/lifecycle.ts` only invokes the transcript digest fallback when `options.transcriptPath`, `options.sessionId`, and a registered handler for `substrate(options)` are present.
- `src/adapters/claude-code/session-digest.ts` implements and registers the Claude Code transcript digest fallback; there is no equivalent Codex adapter.
- `src/adapters/codex/adapter.ts` currently states: "Codex has no automatic session-digest capture... On Codex, run `soma memory digest` manually."

## Proposed implementation

1. Add a Codex transcript digest adapter, analogous to the Claude Code adapter, that can parse Codex session JSONL transcripts deterministically and write a concise 8-15 line digest via `writeSessionDigest`.
2. Update the Codex lifecycle hook to pass enough data into `soma lifecycle session-end`: at minimum `--session-id`, `--cwd`, and `--transcript` when the hook payload exposes or can safely derive them.
3. If Codex `Stop` hook payloads do not expose a transcript path directly, add a bounded resolver for the current session transcript under `$CODEX_HOME/sessions`, keyed by session id and constrained to the Codex home session root.
4. Keep the behavior best-effort: digest failures must not block `Stop`, but should be visible in the lifecycle event summary or diagnostics.
5. Update the Codex projection docs to remove the manual-only limitation once automatic capture is real.

## Acceptance criteria

- Codex `Stop` lifecycle can create one Soma session digest automatically for a qualifying Codex transcript.
- Repeated `Stop` invocations for the same session are idempotent and do not create duplicate digest notes.
- Transcript path resolution is constrained to the Codex session root and refuses path traversal/symlink escape.
- Short sessions below the fallback threshold are skipped without error, matching the Claude Code fallback posture.
- Existing lifecycle metadata capture still runs even when digest extraction fails.
- Tests cover hook argument construction, transcript resolver safety, digest extraction, duplicate/no-op behavior, and the projected Codex limitation text update.

## Notes

This is distinct from Codex built-in memories. Codex memories may summarize prior threads in the background, but Soma needs deterministic writes into the `~/.soma` memory store.

jcfischer:
## Summary
- add a Codex session-end transcript digest fallback and register it with lifecycle
- pass session id, cwd, and a safe transcript path from the Codex Stop hook
- document automatic Codex digest capture and cover write, duplicate, skip, and path-safety behavior

Fixes #413

## Verification
- bun run typecheck
- bun test


jcfischer (commented):

### Architecture

- **[suggestion]** `src/adapters/codex/session-digest.ts:192` — **Qualify sub-agent terminology**
  CONTEXT.md's `agent (always qualified, never bare)` rule says bare `agent` is banned in Soma docs and code, but the new reason strings say `"forced sub-agent - no fallback digest written"` and `"sub-agent session - no fallback digest written"` without naming the substrate-native surface.
  Fix: Use `Codex sub-agent` or another substrate-qualified term in these reason strings.

### Performance

- **[important]** `src/adapters/codex/hooks/codex-hook-entry.mjs:58` — **Stop hook performs bounded sync tree scan**
  The Stop hook falls back to scanning the session root synchronously via `while (queue.length > 0 && scanned < 5000)` and `entries = readdirSync(current.dir, { withFileTypes: true });`, plus per-candidate `lstatSync`. On homes with many Codex sessions, this blocks the hook process for thousands of filesystem operations before lifecycle metadata can run.
  Fix: Avoid recursive discovery in the hook; pass only explicit transcript paths, or move async/indexed lookup into the lifecycle command.

- **[important]** `src/adapters/codex/session-digest.ts:216` — **Transcript parsing duplicates full file**
  The digest path reads the entire transcript with `const transcript = await readFile(options.transcriptPath, { encoding: "utf8", flag: fsConstants.O_RDONLY | fsConstants.O_NOFOLLOW })` and then iterates `for (const raw of transcript.split("\n"))`, allocating another full array of lines. Long Codex JSONL sessions pay an avoidable memory spike at session end.
  Fix: Stream the JSONL with readline and update prompt/tool counters incrementally.

### Maintainability

- **[suggestion]** `src/adapters/codex/hooks/codex-hook-entry.mjs:40` — **Deduplicate transcript path validation**
  The hook repeats the same containment shape as the adapter: `const rel = relative(root, target); if (rel === "" || rel.startsWith("..") || isAbsolute(rel)) return false;` and `const realRel = relative(rootReal, parentReal);`. The same logic also appears in `src/adapters/codex/session-digest.ts`, so future changes to the allowed transcript path rules need parallel edits.
  Fix: Extract the shared path-safety rule into one adapter helper, or document why the hook copy must remain standalone.

### HonestOracle

- **[important]** `src/adapters/codex/hooks/codex-hook-entry.mjs:77` — **Substring match can digest wrong session**
  The docs claim the hook can resolve "the current session JSONL", but the resolver accepts any filename where `entry.name.includes(sessionId)`. A session id that is a substring of another transcript filename can select the newest unrelated file and write a misleading digest for the requested session.


---

jcfischer (commented):

### Security

- **[important]** `src/adapters/codex/session-digest.ts:222` — **Environment can widen transcript root**
  The containment boundary is derived from `process.env.SOMA_CODEX_TRANSCRIPT_ROOT` in `const root = resolve(options.transcriptRoot ?? process.env.SOMA_CODEX_TRANSCRIPT_ROOT ?? join(options.homeDir ?? homedir(), ".codex", "sessions"));`, so a hook environment can authorize arbitrary readable `.jsonl` trees instead of the Codex session root. That creates a concrete ingestion path where a crafted transcript-like file outside `~/.codex/sessions` is summarized into Soma memory despite the documented invariant that paths escaping the Codex session root are refused.
  Fix: Do not honor `SOMA_CODEX_TRANSCRIPT_ROOT` in production hook paths; restrict it to explicit test options or validate it is inside the Codex home session directory.

### Performance

- **[suggestion]** `src/adapters/codex/session-digest.ts:153` — **Bound stored prompt samples**
  The streaming parser still retains every cleaned prompt via `state.prompts.push(entry.text)` even though `samplePrompts(input.prompts, FALLBACK_MAX_PROMPT_LINES)` renders at most 13 prompt lines. Very long Codex sessions therefore allocate memory proportional to all principal prompts instead of the bounded head/tail sample actually needed.
  Fix: Track prompt count plus bounded head/tail samples instead of storing every prompt string.

### Maintainability

- **[suggestion]** `test/install.test.ts:60` — **Deduplicate Codex transcript test helpers**
  `function codexTranscript(lines: object[]): string`, `function codexUser(content: unknown, extra: object = {}): object`, and `function codexAssistantTool(name: string): object` near-copy the same fixture helpers added in `test/codex-session-digest.test.ts`. Keeping two encodings of the Codex transcript shape makes parser-fixture changes easy to miss in one test file.
  Fix: Move the Codex transcript fixture helpers to a shared test helper.

### HonestOracle

- **[important]** `docs/substrate-adapters.md:48` — **Transcript resolution is overstated**
  The docs claim the hook can resolve "the current session JSONL under the Codex session root," but `findSessionTranscript` only checks `${sessionId}.jsonl` and today's `yyyy/mm/dd/${sessionId}.jsonl`. Downstream readers may assume broader Codex session-root discovery exists when nested, dated, or migrated transcript layouts are silently missed.


---

jcfischer (commented):

### Maintainability

- **[suggestion]** `src/adapters/codex/session-digest.ts:179` — **Deduplicate digest accumulation loops**
  The in-memory and file readers repeat the same accumulation shape: `collectDigestEntry(parseTranscriptLine(raw), state);` followed by `return renderDigestBody(state);`. Keeping two loops means future parsing or sampling changes must remember both entry points.
  Fix: Extract a shared `collectDigestBody(lines)` helper that accepts an iterable/async iterable of raw lines.

### HonestOracle

- **[important]** `docs/substrate-adapters.md:50` — **Unsafe explicit paths are not refused**
  The doc claims the fallback "refuses paths that escape the Codex session root", but `resolveSessionTranscript` only accepts a safe explicit path and otherwise falls through to `findSessionTranscript(root, sessionId)`. That means an escaping `transcript_path` can be silently ignored while a different exact session-id fallback is digested, misleading readers who rely on refusal semantics.


---

jcfischer (commented):

1 finding(s): 1 important.


### Security

- **[important]** `src/adapters/codex/session-digest.ts:240` — **Transcript containment check is racy**
  The code validates the parent with `realpath(resolve(target, ".."))` before later reading via `open(transcriptPath, fsConstants.O_RDONLY | fsConstants.O_NOFOLLOW)`, so a writable transcript-root subdirectory can be swapped to a symlink between check and open. That can make a path that passed containment read an outside `.jsonl`, violating the documented invariant that escaped paths are refused.
  Fix: Open relative to a trusted directory handle, or re-check the opened file's real path via `/proc/self/fd`/platform equivalent before parsing.


---

jcfischer (commented):

No findings. Sage approves.



---

PR #415, opened 2026-07-05
fix(memory): promotion mints principal-trust durable note that earns an INDEX line

jcfischer:
## Problem

`soma memory reindex` reported `admitted: 0 / excluded: 54` on the live store despite two real promotions having run. The earned-inclusion INDEX was empty.

**Root cause:** `promoteAlgorithmRunMemory` wrote a bare markdown file under `<STORE>/PROMOTED/` with no frontmatter and no semantic note. The durable note was created later by `memory backfill`, which forces `provenance: import` -> `trust: quarantined`. The admission ladder in `memory-index.ts` excludes quarantined notes unconditionally (prompt-injection defense). So promotion landed in the one tier that can never earn an INDEX line.

## Fix

`promoteAlgorithmRunMemory` now does two writes:
1. The verbatim `PROMOTED/` record (unchanged).
2. **NEW:** a durable note via `writeMemoryNote` with `trigger: principal-correction` -> `trust: principal`, `sourceOfTruth` -> the PROMOTED file.

### Authority model (honest, fail-closed)

`hasPromotionVerification` (>=1 verification entry or passed criterion) is a **PRECONDITION** that refuses to promote unverified work; it is NOT the source of principal authority. The authority is a **deliberate-escalation opt-in** that the caller must pass explicitly at every layer, mirroring `--principal-authority` on `soma memory write`:
- CLI: `soma memory promote ... --principal-authority` (required)
- SDK: `promoteAlgorithmRunMemory({ principalAuthority: true })` (required)
- algorithm `sync-from-isa`: `--principal-authority` forwarded on promote-on-complete
- **Pi.dev tool:** out-of-tree follow-up — `~/.pi/agent/extensions/soma.ts` lives under the path-guarded Pi home and is NOT part of this diff. This PR does NOT prove Pi.dev parity; the operator must apply the matching `principalAuthority` param + `--principal-authority` flag to the tool wrapper separately. Listed as a follow-up, not a verified surface.
  The in-repo CLI test `cli promotes Algorithm run memory` (test/cli.test.ts) exercises the matching `--principal-authority` flag end-to-end.
Fails closed: an omitted authority is a refusal, not a silent downgrade. Soma has no cryptographic principal authentication (the same recorded limitation the `principalAuthority` JSDoc documents for `soma memory write`); this is best-effort trust minting with that same limitation, NOT a principal-signed assertion. What IS enforced is the SURFACE: `import` stays quarantined, `--trigger consolidation` is refused on the public CLI, unverified promotions are refused, and promotions without `--principal-authority` are refused.

### Note-id collision resistance

The note id is `<store>-<title-slug>-<run-part>`, capped at 64 chars. The run part ALWAYS appends an 8-hex-char sha256 prefix of the FULL run id (32 bits) so distinct run ids are probabilistically disambiguated even when their slugs collide (e.g. `a/b` and `a:b` both normalize to `a-b`). By birthday paradox, a 50% collision probability needs ~65k promotions sharing the same title-stem — best-effort collision resistance, not a guarantee. The slug stays readable; the hash is the disambiguation tail.

### Atomicity model

The durable artifacts land in dependency order: PROMOTED file -> durable note (emits its own `memory.write.create` event) -> `memory.promotion` event -> run-provenance update. The guaranteed invariant: after each step, the durable memory artifacts that DO exist are mutually consistent. A note-write failure attempts best-effort removal of the PROMOTED file (the file may remain if the unlink fails; the caller is notified via the thrown error). The event log is append-only, so the event-append catch does NOT delete the note (that would orphan its write event); it leaves consistent durable state and surfaces the gap. A failure in the final run-provenance update leaves the promotion durable but unrecorded on the source run.

### Backfill collision guard
`memory-backfill.ts` `collectSources` skips `PROMOTED/` subtrees.

### API surface
`SomaMemoryPromotionResult` gains `notePath` and `noteId`. `SomaMemoryPromotionOptions` gains required `principalAuthority`. `SyncAlgorithmRunFromVsaOptions` gains `principalAuthority` (forwarded to promote when `promoteOnComplete`). CLI `algorithm sync-from-isa` gains `--principal-authority`. Codex adapter docs updated.

## Design note

Folds the v0 "consolidation is a later step" separation into promotion itself. Deliberate scope decision.

## Verification
- `bun run typecheck` clean
- `bun test` 1747 pass / 0 fail
- Live `soma memory reindex`: admitted 2 (was 0). The 2 existing quarantined promotion-clone notes on the live store were manually upgraded to `trust: principal` in place as a one-time out-of-band data fix (editing frontmatter directly) — this PR ships NO code migration for historical quarantined clones; operators with pre-existing quarantined promotion clones must re-run promotion or upgrade them manually
- Tests: principal-trust note asserted; INDEX admission asserted; refuse-without-authority test; long-title run-id-survival unit test; long-run-id hash-disambiguation unit test
- Sage review iterations addressed: rollback gap, false rollback claim, dense id normalizer, principal-initiated overclaim, doc contradiction, id collision (twice — 20-char cap then hash suffix), fallback cap, comment dedup, explicit-authority opt-in, stale contradicting comment

## Files
- `src/memory-promotion.ts` — mint principal-trust note; honest rollback; explicit authority gate; extracted helper; hash-suffix id disambiguation
- `src/memory-backfill.ts` — skip PROMOTED/ subtrees
- `src/types.ts` — `SomaMemoryPromotionResult.notePath`/`noteId`; `SomaMemoryPromotionOptions.principalAuthority`
- `src/algorithm-vsa-sync.ts` — forward `principalAuthority` on promote-on-complete
- `src/cli/memory.ts` — `--principal-authority` on promote
- `src/cli/algorithm.ts` — `--principal-authority` on sync-from-isa
- `src/adapters/codex/adapter.ts` — doc the flag
- `docs/architecture.md` — name promotion as a governed trust-minting path; honest authority model
- `test/memory.test.ts`, `test/cli.test.ts`, `test/algorithm-vsa-sync.test.ts` — cover the gate and id collision resistance


jcfischer (commented):

2 finding(s): 2 important.


### CodeQuality

- **[important]** `src/memory-backfill.ts:84` — **PROMOTED skip is not path-portable**
  The guard derives the store with `rel.split("/")[0]`, but `rel` is built from filesystem paths and can contain backslashes on Windows. That makes `SOMA_PROMOTION_STORE_DIR_NAMES.has(rel.split("/")[0])` miss `LEARNING\PROMOTED`, so backfill imports promotion-owned PROMOTED files instead of skipping them.
  Fix: Use `entryName === "PROMOTED" && depth === 1 && SOMA_PROMOTION_STORE_DIR_NAMES.has(dirname(rel))`, or pass the parent/store name explicitly during traversal.

### HonestOracle

- **[important]** `docs/architecture.md:180` — **Sync refusal claim hides swallowed failures**
  The docs claim "promotion without `--principal-authority` is refused (fail-closed at the promote-API layer)" and then describe the sync hook as "swallowing PRE-durable promotion refusals ... as no-ops". In code, `syncAlgorithmRunFromVsa` catches any pre-durable error and returns `noopResult()`, so readers can wrongly infer all promotion refusals are surfaced when the sync surface still silently drops broad pre-durable failures.


---

PR #420, opened 2026-07-05
refactor(memory): one symlink-safe memory-traversal seam (#408)

### LINKED ISSUE #408: [memory] 2 · One symlink-safe memory-traversal seam
## Problem

"Enumerate the `.md` notes under a memory subtree **without letting a symlink escape the memory root**" is a single security invariant, but it is re-derived four times — and the four **disagree on the policy**:

- `memory-consolidate.ts:98` `listRealMarkdownFiles` + `:115` `listEpisodicNotes` — **silent-skip** symlinks (`isRealEntry` `:91`).
- `memory-audit.ts:37` `listRealMarkdownFilesRec` — **loud-fail** on a mid-walk inode swap (`AuditTreeError` `:57`).
- `memory-backfill.ts:200` `collectSources` — **throws** on any symlink (`:213,:228`).
- `memory-write.ts:305` `collectDurableNotes` — **no symlink guard at all** (no `lstat` / `O_NOFOLLOW`).

The sharpest symptom is inside one pass: `consolidateMemory` walks episodic notes through the **hardened** `listEpisodicNotes` (`:211`) yet scans the durable corpus through the **agnostic** `collectDurableNotes` (`:500`). So `mark-stale` and `similar-pairs` operate on a corpus that was never symlink-checked, in the same pass that hardens episodic. The asymmetry is invisible because there is no seam to state the policy once.

**Deletion test.** Extract one seam and the complexity concentrates; today a single change to the compartment's symlink stance must be found and re-fixed in four already-drifted places. Highest-leverage extraction in the compartment (per the batch-path Explore agent).

## Proposal — deepen

Introduce `listMemoryNotes(dir, { recursive, onSwap: "skip" | "throw" })` in a new `src/memory-fs.ts` (or as a `SomaPaths` capability), owning the `lstat` / `O_NOFOLLOW` / TOCTOU policy. Consolidate, audit, backfill, and the write-scan all call it; skip-vs-throw becomes a flag, not a re-implementation. Aligns with the CONTEXT.md **private root** / **security trace** stance — the memory root is a boundary a walk must not cross.

## Wins (glossary)

- locality: symlink stance changes in one place
- closes the no-check durable-corpus gap (real security gap)
- the invariant finally has a test surface (today tested 4× through 4 interfaces)
- interface: one walk, four callers

## Files

`src/memory-consolidate.ts:98,115,211,500` · `src/memory-audit.ts:37,57` · `src/memory-backfill.ts:200,213,228` · `src/memory-write.ts:305` · new `src/memory-fs.ts`

## Acceptance

- One `listMemoryNotes` seam owns the walk; the four private walkers are deleted or delegate to it.
- `consolidateMemory` scans the durable corpus through the same hardened walk it uses for episodic.
- One test asserts "traversal never follows a symlink out of the memory root" at the seam (replacing the 4 scattered per-interface tests).
- Behaviour preserved: callers that want loud-fail keep loud-fail via `onSwap: "throw"`.

**Depends on #C1** for its home (once `SomaPaths` owns store layout, `listMemoryNotes` lives beside it).

jcfischer:
## What changed

"Enumerate the `.md` notes under a memory subtree without letting a symlink escape the memory root" was one security invariant, re-derived four times with disagreeing policy (per #408):

- `memory-consolidate.ts` (`listRealMarkdownFiles` + `listEpisodicNotes`) — silent-skip symlinks.
- `memory-audit.ts` (`listRealMarkdownFilesRec`) — loud-fail on a mid-walk inode swap (`AuditTreeError`).
- `memory-backfill.ts` (`collectSources`) — throws on any symlink.
- `memory-write.ts` (`collectDurableNotes`) — **no symlink guard at all** (the security gap).

New `src/memory-fs.ts` introduces `listMemoryNotes(dir, { recursive, onSymlink: "skip" | "throw" })`, owning the `lstat`/TOCTOU mechanics once. The four private walkers now delegate to it (`memory-consolidate.ts`'s `listRealMarkdownFiles` is deleted and `listEpisodicNotes` is reduced to a depth-limited per-month wrapper over the seam; `memory-audit.ts`'s `listRealMarkdownFilesRec` and `memory-backfill.ts`'s `collectSources` become thin wrappers around the seam that keep their own caller-specific business logic).

### The seam's contract (two distinct hazards, one flag)

1. **A symlinked entry** (file or directory, seen directly via `readdir`'s dirent type) — never followed. `onSymlink: "skip"` (default) omits it silently; `onSymlink: "throw"` raises a `MemoryTraversalError` naming the path.
2. **A directory replaced between the pre- and post-`readdir` `lstat`** (a TOCTOU race — the audit's original `AuditTreeError` scenario) — this is **unconditional**, regardless of `onSymlink`: a caller that only asked to tolerate plain symlinked entries never asked to trust a directory that provably changed identity mid-walk.

### A deliberate deviation from the issue's literal suggestion, and why

The issue's guardrail text suggested audit calls the seam with `onSymlink: "throw"`. Read literally, that would also make a **plain** symlinked entry abort the whole walk — which breaks audit's own existing test (`"a symlinked note is NOT followed (only real entries are audited)"`, which asserts a graceful skip, `invalidNotes: []`, no throw). Audit's actual, tested contract is: tolerate a plain symlinked note/dir (skip it), but never trust a directory that was swapped mid-read (throw). So audit calls the seam with `onSymlink: "skip"` and gets its loud-fail-on-race behavior "for free" from the now-unconditional TOCTOU check (previously audit-only code, now a baseline the whole seam enforces for every caller — a pure strengthening: consolidate's and the write-scan's paths never had this race check at all before). `onSymlink: "throw"` ends up being backfill's distinguishing stance alone (it wants ANY symlink, anywhere in the source tree, to abort the import loudly). This is documented in `memory-fs.ts`'s module doc and in `memory-audit.ts`'s comment on `listRealMarkdownFilesRec`.

### Per-caller changes

- **`memory-write.ts` — `collectDurableNotes` (the closed gap).** Was a raw `readdir(dir)` + `.endsWith(".md")` filter with **zero** symlink handling — a planted `semantic/evil.md` → outside-the-tree symlink was read like any other note. Now calls `listMemoryNotes(dir, { onSymlink: "skip" })`, matching the sibling episodic walk `consolidateMemory` already used in the same pass. This function has three callers (`findDuplicateCandidates`'s dedup gate, `memory-index.ts`'s INDEX rebuild, `memory-recall.ts`'s recall) — all three now get the guard for free.
- **`memory-consolidate.ts`.** `listRealMarkdownFiles` deleted; `listEpisodicNotes` now enumerates month dirs and calls `listMemoryNotes(monthDir, { onSymlink: "skip" })` **non-recursively** per month (preserving the original two-level `<base>/<month>/` depth — a symlink-safe re-derivation of the old per-month walk, not a recursive descent); `listArchivedMonthNotes` calls `listMemoryNotes(dir, { onSymlink: "skip" })` directly. `consolidateMemory`'s call into `collectDurableNotes` (memory-write.ts) is unchanged at the call site — it inherits the new guard automatically, closing the "sharpest bug": `mark-stale`/`similar-pairs` now scan the durable corpus through the same hardening as episodic, in the same pass.
- **`memory-audit.ts`.** `listRealMarkdownFilesRec` (and the module-private `AuditTreeError`) replaced by a thin wrapper: `listMemoryNotes(dir, { recursive: true, onSymlink: "skip" })`. Same signature, same call sites, so `auditMemory` needed no other changes. The TOCTOU-race loud-fail is preserved via the seam's unconditional check (see above); the thrown class is now the shared `MemoryTraversalError` (no test asserted the specific `AuditTreeError` class name or message text — confirmed via `grep`).
- **`memory-backfill.ts`.** `collectSources`'s recursive walk now calls `listMemoryNotes(root, { recursive: true, onSymlink: "throw", extensions: [], include })`. The `include` hook carries backfill's own business rules (reserved-top-level-category exclusion from descending, `skipRootFiles`, case-insensitive README exclusion, case-insensitive `.md`/`.markdown` matching) — these are backfill-specific filtering, not traversal-safety concerns the seam should know about, so `extensions: []` disables the seam's own (case-sensitive) default filter and `include`'s own `MARKDOWN_EXT` regex is authoritative. The source ROOT's own symlink refusal (with its specific message) stays local to `collectSources`, unchanged — `listMemoryNotes` treats an abnormal root as empty (matching every other caller), so a `--from` pointing at a symlink is still checked as an explicit precondition before the walk begins. Per-file `dev`/`ino`/`mtimeMs` capture (used by `readSourceNoFollow`'s later ancestor-race check) now happens in a follow-up `lstat` pass over the seam's vetted path list, rather than inline during the walk — functionally equivalent (the leaf-symlink defense is `O_NOFOLLOW` at read time regardless; this only affects how soon after enumeration the stat is captured, and does not weaken the invariant).

## #408 Acceptance checklist

- [x] One `listMemoryNotes` seam owns the walk; the four private walkers are deleted or delegate to it. `listRealMarkdownFiles` (consolidate) deleted and `listEpisodicNotes` reduced to a depth-limited per-month wrapper over the seam; `listRealMarkdownFilesRec` (audit) and `collectSources`'s walk (backfill) delegate; `collectDurableNotes` (write) delegates.
- [x] `consolidateMemory` scans the durable corpus through the same hardened walk it uses for episodic — both now go through `listMemoryNotes` (directly for episodic, via the now-hardened `collectDurableNotes` for durable).
- [x] One test asserts "traversal never follows a symlink out of the memory root" at the seam — new `test/memory-fs.test.ts` (**9 tests** — plus 1 in `test/memory-write.test.ts` and 1 in `test/memory-consolidate.test.ts`, 11 new total; covering: skip-mode omission, throw-mode refusal, symlinked-root-is-empty, missing-dir-is-empty, non-recursive symlink refusal, extension case-sensitivity, `include` business filtering). The four callers' own existing traversal tests are **kept** (not replaced) per this task's explicit Verify instructions — they assert caller-specific messages/stances the seam-level test doesn't cover.
- [x] Behaviour preserved: callers that want loud-fail keep loud-fail. Audit's TOCTOU-race loud-fail is preserved (via the seam's unconditional race check, called with `onSymlink: "skip"` — see the deviation note above); backfill's throw-on-any-symlink is preserved exactly (`onSymlink: "throw"`); consolidate's silent-skip is preserved (`onSymlink: "skip"`).
- [x] The `collectDurableNotes` no-check gap is closed — added a dedicated regression test (`test/memory-write.test.ts`: `"#408: a symlinked note in the durable corpus is never followed by collectDurableNotes (the closed no-check gap)"`) proving a symlinked note pointing outside the memory root is invisible to `collectDurableNotes`, and to its caller `findDuplicateCandidates`.

## Test / typecheck output

```
$ bun run typecheck
$ tsc --noEmit
(clean, exit 0)

$ bun test
 1775 pass
 0 fail
 6298 expect() calls
Ran 1775 tests across 115 files.
```

(Baseline on `origin/main` before this change: 1764 pass / 114 files — this PR adds 11 new tests: 9 in `test/memory-fs.test.ts`, 1 in `test/memory-write.test.ts`, 1 in `test/memory-consolidate.test.ts`.)

```
$ bun run lint
```
48 pre-existing baseline errors remain (confirmed identical, modulo line-number shifts, via a `git stash`/lint/`git stash pop` comparison against `origin/main` — none are in code this PR touches). This PR's new file (`src/memory-fs.ts`) and all edited call sites lint clean.

Explicitly re-ran the tests named in the task's Verify section:
- `test/memory-backfill.test.ts` — `"refuses a symlinked source root"` and `"refuses symlinks loudly"` — both pass individually (`bun test -t`).
- Full `memory-audit.test.ts` and `memory-consolidate.test.ts` suites pass (50 tests, 3 files, 0 fail).

## Test plan

- [x] `bun run typecheck` clean
- [x] `bun test` — 1775 pass, 0 fail (11 new tests, no regressions vs. `origin/main`)
- [x] `bun run lint` — no new errors (baseline-only, confirmed via stash comparison)
- [x] New seam-level tests added to `test/memory-fs.test.ts`
- [x] New regression test added to `test/memory-write.test.ts` proving the `collectDurableNotes` gap is closed

Closes #408

### DIFF
--- a/src/memory-audit.ts
+++ b/src/memory-audit.ts
@@ -1,9 +1,10 @@
-import { type Dirent, constants as fsConstants } from "node:fs";
-import { lstat, readFile, readdir } from "node:fs/promises";
-import { basename, dirname, isAbsolute, join, relative } from "node:path";
+import { constants as fsConstants } from "node:fs";
+import { lstat, readFile } from "node:fs/promises";
+import { basename, dirname, isAbsolute, relative } from "node:path";
 import { parseDigestPointerIds } from "./episodic-digest";
 import { createPaths } from "./paths";
 import { isEnoent } from "./fs-utils";
+import { listMemoryNotes } from "./memory-fs";
 import { runBoundedConcurrent } from "./internal-concurrency";
 import { memoryIndexPath } from "./memory-index";
 import { somaMemoryEventsPath } from "./memory";
@@ -28,48 +29,21 @@ import type { SomaMemoryAuditOptions, SomaMemoryAuditProbe, SomaMemoryAuditResul
  */
 const SCAN_CONCURRENCY = 16;
 
-/** Raised when a directory is replaced (swapped for a symlink or another inode)
- *  between the pre- and post-`readdir` lstat — the audit fails LOUDLY rather than
- *  trusting a possibly-redirected read. */
-class AuditTreeError extends Error {}
-
-/** All `.md` files under `dir`, recursively, REJECTING symlinked dirs/files (they
- *  could point outside the memory root — the audit only trusts real entries). */
-async function listRealMarkdownFilesRec(dir: string): Promise<string[]> {
-  // lstat the dir ITSELF before reading it — `readdir` follows a symlinked directory,
-  // so a symlinked dir could otherwise redirect the walk outside the memory root and
-  // have the audit trust foreign files.
-  const before = await lstat(dir).catch((error) => {
-    if (isEnoent(error)) return undefined;
-    throw error;
-  });
-  if (before === undefined) return []; // a missing dir is genuinely empty
-  if (!before.isDirectory()) return []; // a symlink or non-directory — never follow it
-  let entries: Dirent[];
-  try {
-    entries = await readdir(dir, { withFileTypes: true });
-  } catch (error) {
-    if (isEnoent(error)) return []; // vanished between lstat and readdir
-    throw error; // any other failure is a real blind spot — do not treat as empty
-  }
-  // Close the lstat→readdir TOCTOU: re-lstat and require the SAME real directory
-  // (inode+device unchanged, still not a symlink). A mid-read swap → fail loudly.
-  const after = await lstat(dir).catch(() => undefined);
-  if (after === undefined || after.isSymbolicLink() || !after.isDirectory() || after.ino !== before.ino || after.dev !== before.dev) {
-    throw new AuditTreeError(`directory ${dir} was replaced during the audit walk — refusing to trust the read`);
-  }
-  const files: string[] = [];
-  const subdirs: string[] = [];
-  for (const entry of entries) {
-    if (entry.isSymbolicLink()) continue; // never follow a symlink out of the tree
-    const full = join(dir, entry.name);
-    if (entry.isDirectory()) subdirs.push(full);
-    else if (entry.isFile() && entry.name.endsWith(".md")) files.push(full);
-  }
-  // Walk sibling subtrees concurrently but BOUNDED — a wide archive (many
-  // month/project dirs) neither serializes nor spikes fds on an unbounded fan-out.
-  const nested = await runBoundedConcurrent(subdirs, listRealMarkdownFilesRec, SCAN_CONCURRENCY);
-  return [...files, ...nested.flat()];
+/**
+ * All `.md` files under `dir`, recursively, via the shared `listMemoryNotes`
+ * seam (#408). `onSwap: "skip"` — the audit only trusts real entries, so a
+ * symlinked dir/file is silently omitted rather than followed (matching this
+ * probe's own tests: a symlinked note or note dir is invisible, not flagged).
+ * The seam's mid-walk directory-swap TOCTOU detection (a directory replaced —
+ * a different inode — between the pre- and post-`readdir` `lstat`) is
+ * UNCONDITIONAL regardless of `onSwap`, so the audit's original loud-fail
+ * stance on that specific race (formerly `AuditTreeError`, now the shared
+ * `MemoryTraversalError`) is preserved without asking for "throw" here (which
+ * would also make a plain symlinked entry fail the whole walk — this probe's
+ * contract is to skip those, not abort on them).
+ */
+function listRealMarkdownFilesRec(dir: string): Promise<string[]> {
+  return listMemoryNotes(dir, { recursive: true, onSwap: "skip" });
 }
 
 // Read WITHOUT following a final-component symlink — closes the TOCTOU where a listed
--- a/src/memory-audit.ts
+++ b/src/memory-audit.ts
@@ -31,19 +31,19 @@ const SCAN_CONCURRENCY = 16;
 
 /**
  * All `.md` files under `dir`, recursively, via the shared `listMemoryNotes`
- * seam (#408). `onSwap: "skip"` — the audit only trusts real entries, so a
+ * seam (#408). `onSymlink: "skip"` — the audit only trusts real entries, so a
  * symlinked dir/file is silently omitted rather than followed (matching this
  * probe's own tests: a symlinked note or note dir is invisible, not flagged).
  * The seam's mid-walk directory-swap TOCTOU detection (a directory replaced —
  * a different inode — between the pre- and post-`readdir` `lstat`) is
- * UNCONDITIONAL regardless of `onSwap`, so the audit's original loud-fail
+ * UNCONDITIONAL regardless of `onSymlink`, so the audit's original loud-fail
  * stance on that specific race (formerly `AuditTreeError`, now the shared
  * `MemoryTraversalError`) is preserved without asking for "throw" here (which
  * would also make a plain symlinked entry fail the whole walk — this probe's
  * contract is to skip those, not abort on them).
  */
 function listRealMarkdownFilesRec(dir: string): Promise<string[]> {
-  return listMemoryNotes(dir, { recursive: true, onSwap: "skip" });
+  return listMemoryNotes(dir, { recursive: true, onSymlink: "skip" });
 }
 
 // Read WITHOUT following a final-component symlink — closes the TOCTOU where a listed
--- a/src/memory-backfill.ts
+++ b/src/memory-backfill.ts
@@ -36,8 +36,9 @@
  */
 import { createHash } from "node:crypto";
 import { constants as FS } from "node:fs";
-import { lstat, mkdir, open, readFile, readdir, writeFile } from "node:fs/promises";
-import { dirname, join, relative, resolve, sep } from "node:path";
+import { lstat, mkdir, open, readFile, writeFile } from "node:fs/promises";
+import { basename, dirname, relative, resolve, sep } from "node:path";
+import { listMemoryNotes } from "./memory-fs";
 import { rebuildMemoryIndex } from "./memory-index";
 import { MemoryNoteError } from "./memory-note";
 import { memoryNotePath, writeMemoryNote } from "./memory-write";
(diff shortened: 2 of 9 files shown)


jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:110` — **Consolidation now scans nested notes**
  The new implementation says it handles a "two-level `<base>/<month>/` tree" but uses `return listMemoryNotes(base, { recursive: true, onSwap: "skip" });`, which descends into every nested subdirectory. The deleted code only listed real `.md` files directly inside each month directory, so nested files that were previously ignored can now be parsed or moved by consolidation.
  Fix: Preserve the old depth limit, e.g. list month directories first and call `listMemoryNotes(monthDir, { onSwap: "skip" })` for each.

### Security

- **[important]** `src/memory-fs.ts:145` — **Leaf paths can be swapped after enumeration**
  The seam trusts the `readdir` type and returns the path with `if (entry.isFile()) { ... files.push(full); }`, but it never `lstat`s the leaf immediately before returning it or opens it with `O_NOFOLLOW`. An attacker who can mutate the memory tree can replace a listed file with a symlink before callers such as `collectDurableNotes` read the returned path, reintroducing the symlink escape this seam is meant to prevent.
  Fix: Before returning a file path, `lstat(full)` and require a real file, or return/open file handles using `O_NOFOLLOW` at the read boundary.

### Performance

- **[suggestion]** `src/memory-backfill.ts:238` — **Avoid double sorting backfill sources**
  Backfill now gets paths via `const paths = await listMemoryNotes(root, {` while the seam unconditionally does `return result.sort();`, then backfill sorts again with `return files.sort((a, b) => ...);`. Large legacy imports now pay two O(n log n) sorts and keep the intermediate path list resident before the final `SourceFile[]` sort.
  Fix: Let callers opt out of seam sorting, or skip the final sort when `listMemoryNotes` ordering is already sufficient.

### Maintainability

- **[suggestion]** `src/memory-fs.ts:55` — **Rename the symlink policy**
  The option is declared as `onSwap?: MemorySwapPolicy;`, but the implementation uses it for plain symlinked entries: `if (entry.isSymbolicLink()) { if (onSwap === "throw")`. Since directory swaps are now unconditional, `onSwap` no longer describes the caller-visible policy it controls.
  Fix: Rename it to `onSymlink` or `symlinkPolicy`; keep directory-swap failure unconditional.

### HonestOracle

- **[important]** `src/memory-consolidate.ts:104` — **Two-level scan claim is false**
  The comment claims `"Note files under a two-level <base>/<month>/ tree"`, but the implementation immediately returns `listMemoryNotes(base, { recursive: true, onSwap: "skip" })`, which descends without a depth limit. This makes the PR's behavior-preservation claim misleading because nested files below `<base>/<month>/` are now in scope.


---

jcfischer:
## Sage R1 addressed — all 5 items (pushed `3a6e5b6`)

**1. Scan-depth regression (important)** — `src/memory-consolidate.ts`. `listEpisodicNotes` no longer walks `recursive: true`. Restored the strict two-level `<base>/<month>/` semantics: enumerate the real (non-symlink) month dirs directly under `base`, then call `listMemoryNotes(monthDir, { onSymlink: "skip" })` **non-recursive** per month, so nothing below a month dir is ever parsed/moved. The doc comment now states the depth limit explicitly. Added a regression test (`test/memory-consolidate.test.ts`: `"episodic enumeration is depth-limited: a note NESTED below a month dir is NOT archived"`) — verified it **fails** on the recursive version (`unreadable` length 1) and **passes** depth-limited.

**2. Leaf TOCTOU (important, security)** — `src/memory-fs.ts` + `src/memory-write.ts`. Two-layer close:
- Seam: each file candidate is now re-`lstat`ed right before it is returned and must still be a real regular file (`fileCandidates` → bounded-concurrent `lstat` → `verified`). A leaf swapped for a symlink between `readdir` and return is dropped — the seam never hands back a path that resolves through a symlink. Documented as "hazard 3" in the module doc.
- Reader: `collectDurableNotes` (the caller you named) now reads with `O_RDONLY | O_NOFOLLOW`, so even a leaf swapped *after* the seam returns fails the open (ELOOP → surfaced as unreadable) rather than being followed at read time.
- Tests (`test/memory-fs.test.ts`): `"leaf guard: a symlinked file leaf is never returned, and every returned path is a real regular file"` and `"leaf swapped to a symlink after enumeration is not followed at read (O_NOFOLLOW boundary)"` — the latter deterministically enumerates, then swaps the returned leaf to an outside symlink, then reads via the same `O_NOFOLLOW` flags and asserts ELOOP.

**3. Double sort (suggestion)** — `src/memory-fs.ts` + `src/memory-backfill.ts`. Added a `sort?: boolean` option (default `true`, so consolidate/audit/write keep deterministic ordering). Backfill passes `sort: false` and keeps its own authoritative `SourceFile`-by-POSIX-`relativePath` sort — no more redundant double sort.

**4. Rename option (suggestion)** — `onSwap` → `onSymlink` (type `MemorySwapPolicy` → `MemorySymlinkPolicy`) across the seam, all four callers, and the tests. The name now reflects that it governs only plain symlinked *entries*; the directory-swap and leaf-swap races are unconditional (documented).

### Verify
- `bun test` — **1775 pass, 0 fail**, 6309 expect() calls, 115 files (was 1772; +3 new tests this round).
- `bun run lint` — no new errors (baseline-only; confirmed identical error count for the two touched-but-baseline-flagged files via stash comparison; `memory-fs.ts`/`memory-backfill.ts`/`memory-write.ts` and all test files lint clean, exit 0).
- The two backfill symlink tests (`"refuses a symlinked source root"`, `"refuses symlinks loudly"`) + full audit + consolidate suites still pass.

**One-line note on preserving consolidation depth:** episodic notes live at `<base>/<month>/<note>.md` and never deeper, so `listEpisodicNotes` enumerates month dirs itself and calls the seam **non-recursively** per month — identical two-level reach to the original `listRealMarkdownFiles`-per-month, just with the symlink mechanics now owned by the shared seam.

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-fs.ts:176` — **Throw policy silently drops swapped symlinks**
  The leaf verification path uses `return st !== undefined && !st.isSymbolicLink() && st.isFile() ? full : undefined;`, so under `onSymlink: "throw"` a file that becomes a symlink after `readdir` is silently omitted instead of refused loudly. That contradicts the same function's throw-mode behavior for symlinked dirents and weakens backfill's loud-fail contract.
  Fix: If `onSymlink === "throw"` and `st?.isSymbolicLink()`, throw `MemoryTraversalError` instead of returning `undefined`.

### Maintainability

- **[suggestion]** `src/memory-fs.ts:122` — **Split the nested walk phases**
  `async function walk(current: string, depth: number): Promise<string[]> {` now owns stable-directory validation, entry policy, recursive fan-out, and leaf verification, ending with `const verified = await runBoundedConcurrent(` in the same body. Those are separable phases, so future policy changes will require editing one long branch-heavy function.
  Fix: Extract helpers like `readStableDirectory`, `partitionEntries`, and `verifyFileCandidates`.

### HonestOracle

- **[important]** `PR description:17` — **Documented option name is nonexistent**
  The description claims `listMemoryNotes(dir, { recursive, onSwap: "skip" | "throw" })`, but the diff implements `onSymlink?: MemorySymlinkPolicy` and all call sites use `onSymlink`. This misstates the seam's API contract and makes the PR description unusable as integration evidence.

- **[suggestion]** `PR description:61` — **Test-count evidence is false**
  The checklist claims `new test/memory-fs.test.ts (7 tests...)` and later says `this PR adds 8 new tests`, but the diff adds 9 `test(...)` cases in `test/memory-fs.test.ts` plus one in `test/memory-write.test.ts` and one in `test/memory-consolidate.test.ts`. The stated verification evidence does not match the submitted diff.


---

jcfischer (commented):

### Performance

- **[suggestion]** `src/memory-backfill.ts:251` — **Avoid restatting every backfill source**
  `listMemoryNotes` already re-lstats each returned file via `const st = await lstatOrUndefined(full);`, then backfill immediately does another `stat = await lstat(abs)` inside `for (const abs of paths)`. That doubles leaf `lstat` syscalls for every imported source file in large backfills.
  Fix: Have the seam optionally return verified stat metadata, or keep the backfill walker local for the metadata capture path.

### HonestOracle

- **[important]** `PR description:48` — **Checklist claims deleted walker still exists**
  The checklist says "`listRealMarkdownFiles`/`listEpisodicNotes` (consolidate) deleted", but the diff still defines `async function listEpisodicNotes(base: string): Promise<string[]>`. This misleads downstream reviewers about whether consolidate's private traversal surface was actually removed under #408.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-backfill.ts:221` — **Backfill file root now silently no-ops**
  `collectSources` now only rejects `rootStat.isSymbolicLink()` before calling the seam, while the seam treats `!before.isDirectory()` as empty via `if (before === undefined || before.isSymbolicLink() || !before.isDirectory()) return [];`. Previously a regular-file `--from` reached `readdir(dir)` and surfaced `ENOTDIR`; now a bad source path can report no sources instead of failing.
  Fix: After the symlink check, throw if `!rootStat.isDirectory()`.

### Performance

- **[suggestion]** `src/memory-backfill.ts:238` — **Backfill now double-buffers source paths**
  The diff adds `const paths = await listMemoryNotes(root, {` followed by `const files: SourceFile[] = [];` and `for (const abs of paths)`, so large backfills now hold every absolute path plus every `SourceFile` at once. The old walker pushed `SourceFile`s during traversal, avoiding this extra full-tree array.
  Fix: Let the seam optionally map verified files to caller records, or keep backfill's record construction inside the traversal callback.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### Performance

- **[suggestion]** `src/memory-write.ts:402` — **Avoid duplicate leaf verification**
  The durable scan now does `files = await listMemoryNotes(dir, { onSymlink: "skip" });`, whose seam re-lstats every returned file, then immediately opens each path with `readFile(path, { encoding: "utf8", flag: FS.O_RDONLY | FS.O_NOFOLLOW })`. For large durable corpora this adds one extra `lstat` syscall per note on recall/index/dedup paths even though the read already uses `O_NOFOLLOW`.
  Fix: Let this caller skip seam leaf re-lstat and rely on `O_NOFOLLOW` at read time.

### HonestOracle

- **[suggestion]** `src/memory-fs.ts:32` — **Unsupported historical contract claim**
  The comment claims, "No production caller has ever needed this to fail open," but the diff only shows the current callers and tests, not production history or prior runtime requirements. A downstream maintainer could treat that as established evidence for making the TOCTOU failure unconditional when it is only asserted here.


---

Issue #410, opened 2026-07-05
[memory] 4 · Name the shared corpus operations (jaccard · dates · slug · sanitize)


jcfischer:
## Problem

Compare / date / slugify / **sanitize** over the note corpus are each re-decided per call site. The team already builds these seams (`memory-terms.ts` for tokenization, `dated-slug.ts` for slugs) — Memory just declines to route through them.

| operation | copies | divergence risk |
|-----------|--------|-----------------|
| `jaccard(a,b)` | ×2 | `memory-write.ts:364` vs `memory-consolidate.ts:141`; `0.6` threshold duplicated as a bare literal (`memory-write.ts:62`) |
| `dateMs` / `ageDays` | ×3 | `memory-index.ts:78,84` · `memory-recall.ts:97,103` · `memory-consolidate.ts:80,86`, each with its own future-timestamp nuance |
| `slugify → id` | ×3 | `memory-backfill.ts:136` · `memory-episodic.ts:104` · `memory-promotion.ts:28` re-approximate the id grammar then re-validate (`:219,:64,:136`); `dated-slug.ts:4` exists but is used only by the algorithm path |
| **sanitize note text** | ×3 | `cli/memory.ts:990` `sanitizeForTerminal` (full ANSI/OSC strip, security-critical) is wired **only** into recall; `memory-index.ts:128` `oneLine` (weaker, disclaims injection defense `:168`) guards the **always-loaded INDEX**; `memory-consolidate.ts:138` `firstBodyLine` is a third strip |

The security-critical one is the tell: untrusted note bodies flow to three output surfaces, the **weakest** stripper guards the projected INDEX artifact, and the strongest is buried as a CLI-private helper unreachable by the others. A new output surface silently gets no protection — exactly the drift `memory-terms.ts` was created to prevent.

**Deletion test.** The trivial `isoDate` copies are pass-throughs (merge only opportunistically). The load-bearing ones — the sanitizer and the id-slug grammar — concentrate real behaviour/security when unified.

## Proposal — deepen

- Export `jaccard` + `noteDateMs`/`ageDays` from `memory-terms.ts` (or a new `memory-corpus.ts`); consolidate's near-dup listing calls the same scorer as `findDuplicateCandidates`.
- One `toNoteIdSlug(base)` next to the id-grammar definition — **valid-by-construction**, killing the produce-then-validate scatter; backfill/episodic/promotion call it.
[...]

Issue #409, opened 2026-07-05
[memory] 3 · Localize mutation governance (trust · provenance · audit-meta)


jcfischer:
## Problem

The M1 mutation-governance model is conceptually **one function** — `(trigger, target?, authoritySignals) → {trust, provenance, auditMeta} | refusal` — but it is implemented as six tiny predicates + a lookup table, and the gate **sequencing** is re-authored at all four mutation sites in a different subset and order:

- `src/memory-write.ts` — `TRUST_RANK:86`, `assertTierAuthority:233`, `assertMintAuthority:248`, `assertMayMutate:262`, `authorityMeta:278`, `resolveProvenance:395` (driven by `SOMA_MEMORY_TRIGGER_TRUST`, `types.ts:2386`).
- `createNote` → mint gate + `resolveProvenance` + `authorityMeta` inline (`:562,:588,:640`).
- `mergeNote` → `assertMayMutate` + `authorityMeta` (`:666,:685`).
- `supersedeNote` → mint gate + `assertMayMutate(old)` + **two** `authorityMeta` (`:712,:763-764`).
- `verifyMemoryNote` → `assertTierAuthority` + `authorityMeta` inline (`:817,:834`).

The invariant the docstring promises — "authority is **checked** at the gate and **recorded** in the event" (`:36-37`) — is exactly what is *not* localized: the check and its matching audit-meta are paired **by hand** at four sites. A fifth mutation path must re-derive the pairing correctly with nothing enforcing it.

**Deletion test.** The six predicates are genuinely distinct rules that would reappear if deleted — they earn their keep — but they belong behind **one** interface, not six, called once per path.

**Testability.** Already good: tests exercise this through `writeMemoryNote`/`verifyMemoryNote` (`memory-write.test.ts:114,140,252,276`). The friction is maintainer-side locality, not test reach.

## Proposal — deepen

One `resolveMutationGovernance(trigger, target | null, auth) → { trust, provenance, eventMeta } | MemoryNoteError` that every mutation path calls **exactly once**, so check and audit-meta can never drift apart. The six predicates + trust table become its internal implementation.

## Wins (glossary)

- locality: authority rules concentrate in one module
- check + audit-meta can't desync
- a new mutation path inherits the gate by construction
- interface shrinks; six predicates go internal

## Files

`src/memory-write.ts:86,233,248,262,278,395,562,666,712,817` · `src/types.ts:2386`

## Acceptance

[...]

Issue #407, opened 2026-07-05
[memory] 1 · Route path construction through the SomaPaths seam


jcfischer:
## Problem

The one fact every Memory module must agree on — **where `~/.soma` is, and which dir each store lives in** — is copied ~40× instead of living behind the `SomaPaths` seam that already exists (`src/paths.ts`, `src/types.ts:491`).

- **~40 hand-rolled soma-home resolvers**: 18 private `function resolveSomaHome` + 24 inline `options.somaHome ?? join(options.homeDir ?? homedir(), ".soma")`. Memory offenders: `memory-promotion.ts:24`, `memory-backfill.ts:77`; plus `lifecycle.ts:34`, `vsa.ts:69`, `observability.ts:29`, etc.
- **The copies have already forked**: `memory-backfill.ts:78` wraps `homeDir` in an extra `resolve()`; `memory-promotion.ts:25` does not. A latent behavioural fork in a path that feeds root-containment.
- **Security regression**: the hand-rolled resolvers skip `assertInsideRoot` (`paths.ts:14`) that `createPaths` provides.
- **`events.jsonl` literal reconstructed inline 12×** (8 of them in `lifecycle.ts:329,426,475,515,579,621,659,819`) vs 3 sites that use `paths.events()` (`paths.ts:49`).
- **Store-layout is a half-seam**: `SomaPaths` names `learning/work/relationship/state/…` but **not** `knowledge`, `semantic`, `procedural`, `episodic`, `PROMOTED`, `archive` — those are string literals in 4–10 files each. The store→dir map is triplicated (`PROMOTION_STORE_DIRS` `memory-promotion.ts:11`, `WRITABLE_TYPE_DIRS` `memory-write.ts:66`, `parseMemoryPromotionStore` `cli/memory.ts:570`).

**Deletion test.** Delete `defaultSomaHome`/the store accessors and almost nothing breaks — proof the knowledge is already smeared across N callers, not localised. The seam is the road not taken.

## Proposal — deepen

Make `createPaths(options)` the **only** soma-home resolver, and extend `SomaPaths` to name **every** store the note subsystem reads/writes.

- Replace all 18 private `resolveSomaHome` with `createPaths(options).root()` (memory-promotion + memory-backfill first — the compartment's own).
- Route every event-log write through `paths.events()`; add `paths.state(...segments)` for sibling `STATE/*.json`.
[...]

PR #422, opened 2026-07-05
refactor(memory): one plan/apply + atomicity shape for note writes (#412)

### LINKED ISSUE #412: [memory] 6 · One plan/apply + atomicity shape for note writes
## Problem

Two smaller shape issues in the same family — plan/apply and write-atomicity — flagged **Speculative**: real, but lower-leverage and only worth doing alongside a change already in these files.

**(a) Supersede hand-rolls a second undo.** `memory-write.ts:120` `appendMutationEvent` is the deep "one mutation = one event, else roll back" primitive. But `supersedeNote` (`:724-745`) contains a **separate** hand-written two-file undo (restore old bytes + unlink new) for the pre-event failure window, throwing an `AggregateError` — while the primitive's rollback closure (`:770-773`) does the byte-identical restore and throws a plain `Error` with `cause`. Two divergent inconsistency-error shapes for the same operation. The mutation body knows rollback mechanics the event primitive was meant to own (interface leaks across the seam).

**(b) Consolidate patches the plan after the dry-run boundary.** `memory-consolidate.ts:510` builds `result`, returns it on dry-run (`:523`), then on the real run **mutates it in place**: filters `result.markedStale` (`:533`), reassigns `result.stateGced` (`:538`) and `result.unreadable` (`:542`), recomputes `result.mutated` (`:547`). The docstring's "dry-run plan equals the real run" holds only absent TOCTOU races; `SomaMemoryConsolidateResult` conflates "planned" and "actually-applied" in one mutable object. (Defensible — the races are real and surfaced — so this is a **shape** observation, not a bug.)

`runMemoryBackfill` already proves the good shape: `plan = run({dryRun:true})`, one immutable core. **Don't touch backfill.**

## Proposal — deepen

- **(a)** A `writeNotesAtomically(files[], eventInput)` that stages all file writes, then appends the event, with a **single** rollback restoring every staged file — so create/merge/supersede/verify share one atomicity contract and one error shape.
- **(b)** Model consolidate like backfill: a pure `plan()` returning an immutable plan, plus `apply(plan)` returning an `applied` delta (skips + omissions), instead of in-place patching one result across the dry-run boundary.

## Wins (glossary)

- one atomicity contract, one error shape
- plan stays immutable across the dry-run boundary
- locality: rollback logic concentrates in the event primitive

## Files

`src/memory-write.ts:120,724-745,770-773` · `src/memory-consolidate.ts:510-547`

## Acceptance

- Supersede has no bespoke undo block; atomicity + error shape come from one primitive.
- Consolidate's dry-run and real-run share an immutable plan; applied deltas are a separate return, not in-place mutation.
- Existing tests green.

jcfischer:
## Summary

Two behaviour-preserving shape fixes flagged **Speculative** in #412 — plan/apply and write-atomicity, both in the note-write / consolidate paths.

### (a) One atomicity contract, one error shape for note writes (`src/memory-write.ts`)

`appendMutationEvent` was already the "one mutation = one event, else roll back" primitive, but `supersedeNote` hand-rolled a **separate** two-file undo for the pre-event failure window (a mid-write failure closing the old note), throwing `AggregateError` — a different shape than the primitive's own rollback (`Error` with `cause`).

Added `writeNotesAtomically(somaHome, now, substrate, kind, writes[], fields)`: it stages every file write in order, and — whether a *mid-write* failure or a *post-write event-append* failure triggers the rollback — routes through the same `rollbackAndThrow` helper, producing the identical error shape either way. `createNote`, `mergeNote`, `verifyMemoryNote`, and `supersedeNote` (2 staged writes: mint the replacement, close the old note) all now call this one function; `supersedeNote`'s bespoke `AggregateError` block is gone.

**Sanctioned behaviour change** (per the issue's own ask — this *is* the fix): on a mid-write failure whose rollback succeeds, the thrown error used to be the raw original error (unwrapped) for supersede; it is now a wrapped `Error` with `.cause`, matching the append-failure shape. Same for the rollback-also-fails case: `AggregateError` → `Error` with `.cause`. All *file contents written, events appended, and refusals* are byte-identical to before (verified by the full existing test suite, unchanged).

### (b) Consolidate: pure plan, separately-returned applied delta (`src/memory-consolidate.ts`)

`consolidateMemory` used to build one `SomaMemoryConsolidateResult` object, return it verbatim on `--dry-run`, then on a real run **mutate that same object** post-hoc (filter `markedStale`, reassign `stateGced`/`unreadable`, recompute `mutated`) to account for TOCTOU skips discovered during apply.

Split into:
- `planConsolidation(paths, options, now)` — pure reads only, returns an **immutable** `ConsolidationPlan` (the object and every array field are `Object.freeze`d — an accidental future mutation now throws a `TypeError`, since ES modules run in strict mode).
- `applyConsolidationPlan(paths, plan, now, substrate)` — performs the mutations (archive move + digest regen, stale marks, state GC, INDEX rebuild, event append) and returns a **separate** `ConsolidationApplied` delta (`markedStale`, `stateGced`, `unreadable`, `mutated` — the actual, post-TOCTOU-skip values).
- `consolidateMemory` orchestrates: `plan = await planConsolidation(...)`; dry-run returns `planToResult(plan, true)`; a real run calls `applyConsolidationPlan` and returns `planToResult(plan, false, applied)` — a fresh combination, never a patched shared object.

This mirrors `runMemoryBackfill`'s `plan()` / `apply(plan)` shape (M8), per the issue's explicit reference. `runMemoryBackfill` itself is untouched. The **public** `SomaMemoryConsolidateResult` shape and every value it produces are unchanged (verified by the full existing consolidate test suite, unchanged).

## #412 Acceptance checklist

- [x] Supersede has no bespoke undo block; atomicity + error shape come from one primitive (`writeNotesAtomically`).
- [x] Consolidate's dry-run and real-run share an immutable plan (`ConsolidationPlan`, frozen); applied deltas are a separate return (`ConsolidationApplied`), not in-place mutation.
- [x] Existing tests green — see exact numbers below.

## Tests added

- `test/memory-write.test.ts` (+3, 41 → 44 tests — incl. a first-`"w"`-overwrite rollback test added in review):
  - `writeNotesAtomically: a mid-write failure rolls back every staged write with the SAME error shape as an event-append failure (#412)` — stages a valid "wx" create followed by a "w" overwrite whose note deliberately fails `serializeMemoryNote`'s round-trip check (an embedded newline in `hook`), forcing a mid-write failure deterministically without any root-unsafe permission/directory trick. Asserts: the first write is rolled back (unlinked), the second write's target is restored to the caller-supplied prior bytes (proving the rollback ran, not merely that nothing changed), the thrown error is a plain `Error` (not `AggregateError`) with `.cause` set, and its message matches the exact same template the event-append-failure path uses.
  - `writeNotesAtomically: a failure on the FIRST write has nothing staged, so it propagates un-rolled-back and un-wrapped` — confirms the "nothing to roll back yet" branch.
  - `writeNotesAtomically` and its `AtomicNoteWrite` type are exported from `memory-write.ts` module-private-style (not public index API) for this direct test, following the file's existing convention (`resolveMutationGovernance`, `findDuplicateCandidates`, etc.).
- `test/memory-consolidate.test.ts` (+1, 13 → 14 tests):
  - `planConsolidation's plan is frozen and untouched by applyConsolidationPlan (#412)` — seeds an aged episodic note + a stale semantic note, builds a plan with real content (episodic archive + stale mark both non-empty, so the proof isn't vacuous), asserts the plan and every array field are frozen and that mutation attempts throw `TypeError`, then applies it and asserts the plan object is still deep-equal to a pre-apply snapshot while the applied delta and the on-disk state reflect the real mutation.
  - `planConsolidation`, `applyConsolidationPlan`, `ConsolidationPlan`, `ConsolidationApplied` exported module-private-style for this test.

## Verification

- `bun run typecheck` — clean (`tsc --noEmit`, no output).
- `bun test` — **1798 pass, 0 fail**, across 116 files.
- `bun test test/memory-write.test.ts test/memory-consolidate.test.ts` — **58 pass, 0 fail** (44 + 14).
- No LSP tool was available in this environment; `tsc --noEmit` is the equivalent static check and is clean.
- Reused the merged seams (#407 `createPaths`/`SomaPaths`, #408 `listMemoryNotes`, #409 `resolveMutationGovernance`, #410 `memory-corpus`'s `jaccard`/`NEAR_DUPLICATE_JACCARD_THRESHOLD`) — nothing they centralized was reintroduced. `runMemoryBackfill` (the reference plan/apply shape) is untouched.

Closes #412

### DIFF
--- a/src/memory-consolidate.ts
+++ b/src/memory-consolidate.ts
@@ -446,11 +446,13 @@ async function applyStateGc(root: string, stateGc: string[], now: Date): Promise
 
 /**
  * The pure-read output of the plan phase: every file operation consolidation
- * WOULD perform, computed once from the tree + `now`. Frozen (top level and
- * every array field) before it is returned — `applyConsolidationPlan` reads
- * this object but must never patch it, so an accidental future mutation
- * throws (ES modules run in strict mode) instead of silently letting a
- * dry-run's reported plan and a real run's applied result drift apart. This
+ * WOULD perform, computed once from the tree + `now`. SHALLOW-frozen (the
+ * top-level object and every array field, NOT the nested note/archive entries)
+ * before it is returned — enough that `applyConsolidationPlan` cannot repoint,
+ * grow, or replace the plan's arrays (an accidental structural mutation throws
+ * in strict mode); it does not deep-freeze individual note contents. That is
+ * the property the plan/apply split needs: the set of ops can't drift between a
+ * dry-run's reported plan and a real run's applied result. This
  * is the same `plan()` / `apply(plan)` split `runMemoryBackfill` uses (M8) —
  * a `SomaMemoryConsolidateResult` for a real run is built by COMBINING this
  * plan with the separately-returned `ConsolidationApplied` delta, never by
@@ -615,11 +617,13 @@ function planToResult(plan: ConsolidationPlan, dryRun: boolean, applied?: Consol
     somaHome: plan.somaHome,
     dryRun,
     archived: plan.episodic.map((e) => e.plan),
-    digestsWritten: plan.digestsWritten,
+    // Copy the plan-owned (frozen) arrays so the public result stays mutable —
+    // dry-run callers previously got fresh mutable arrays and some mutate them.
+    digestsWritten: [...plan.digestsWritten],
     markedStale: applied ? applied.markedStale : plan.staleMarks.map((s) => s.rel).sort(),
-    stateGced: applied ? applied.stateGced : plan.stateGc,
-    similarPairs: plan.similarPairs,
-    unreadable: applied ? applied.unreadable : plan.unreadable,
+    stateGced: applied ? applied.stateGced : [...plan.stateGc],
+    similarPairs: [...plan.similarPairs],
+    unreadable: applied ? applied.unreadable : [...plan.unreadable],
     mutated: applied ? applied.mutated : plan.mutated,
     indexPath: plan.indexPath,
   };
--- a/src/memory-consolidate.ts
+++ b/src/memory-consolidate.ts
@@ -22,11 +22,15 @@ import type {
 
 /**
  * Deterministic consolidation (subsystem M6). A no-LLM maintenance pass. Every op
- * is computed into a PLAN first, so a `--dry-run` reports the SAME set of file
- * operations (which notes archive, which are marked stale, which state files are
- * deleted, which pairs are lexically similar) that the real run applies — the dry-run plan
- * equals the real run's plan. (It does NOT reproduce byte-level digest/INDEX
- * content; it enumerates the operations, not their diffs.) Ops, in apply order:
+ * is computed into an immutable PLAN first (`planConsolidation`), so a `--dry-run`
+ * reports the SAME set of file operations (which notes archive, which are marked
+ * stale, which state files are deleted, which pairs are lexically similar) that a
+ * real run applies — the dry-run plan equals the real run's plan. (It does NOT
+ * reproduce byte-level digest/INDEX content; it enumerates the operations, not
+ * their diffs.) A real run separately applies that plan (`applyConsolidationPlan`)
+ * and returns what ACTUALLY happened as its own delta — the same `plan()` /
+ * `apply(plan)` split `runMemoryBackfill` uses (M8), never one mutable result
+ * object patched across the dry-run boundary. Ops, in apply order:
  *
  * 1. **Prune aged episodic** — session notes older than 90d and action notes older
  *    than 180d (by `created`) are folded into a monthly digest (`episodic/digests/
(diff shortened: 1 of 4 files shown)


jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-write.ts:175` — **First overwrite failure skips rollback**
  The branch `if (i === 0) throw writeError; // nothing staged yet — nothing to roll back` is incorrect for a first write with `flag: "w"`: the same function notes that an overwrite `"may have already truncated its target before failing"`. Single-write callers like merge and verify now route through this path, so a failed overwrite can leave a mutated note without restoration.
  Fix: For `i === 0`, only propagate directly for `"wx"`; for `"w"`, call `rollbackAndThrow(kind, "write", writeError, () => undoThrough(0))`.

### Maintainability

- **[suggestion]** `src/memory-write.ts:201` — **Group atomic write parameters**
  `export async function writeNotesAtomically(
  somaHome: string,
  now: Date,
  substrate: SomaMemoryWriteOptions["substrate"],
  kind: string,
  writes: AtomicNoteWrite[],
  fields: { summary: string; artifactPaths: string[]; metadata: Record<string, unknown> },
)` introduces six positional parameters, mixing execution context, event identity, staged writes, and event fields. That makes the four call sites harder to scan and raises future-change cost when the event envelope grows.
  Fix: Accept one object parameter, e.g. `{ somaHome, now, substrate, kind, writes, event }`.

### HonestOracle

- **[blocker]** `src/memory-write.ts:174` — **First-write rollback claim is false**
  The comment claims "Stage every file write in `writes`, then append the mutation's event, as ONE atomic unit" and that failures "rolls back every write up to and INCLUDING it", but the implementation says `if (i === 0) throw writeError; // nothing staged yet — nothing to roll back`. For first writes using `flag: "w"` in merge/verify, a write can fail after opening/truncating the target, so downstream readers are misled into relying on rollback coverage the code explicitly does not provide.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-consolidate.ts:606` — **Public result exposes frozen arrays**
  `planToResult` returns plan-owned frozen arrays directly via `digestsWritten: plan.digestsWritten`, `stateGced: applied ? applied.stateGced : plan.stateGc`, `similarPairs: plan.similarPairs`, and `unreadable: applied ? applied.unreadable : plan.unreadable`. Because `planConsolidation` now freezes those arrays, dry-run callers receive a `SomaMemoryConsolidateResult` whose arrays throw on mutation, unlike the previous fresh mutable result arrays.
  Fix: Return copies in `planToResult`, e.g. `digestsWritten: [...plan.digestsWritten]`, `similarPairs: [...plan.similarPairs]`, `stateGced: applied ? applied.stateGced : [...plan.stateGc]`, and `unreadable: applied ? applied.unreadable : [...plan.unreadable]`.

### HonestOracle

- **[important]** `src/memory-consolidate.ts:447` — **Plan immutability is overstated**
  The comment claims "The pure-read output of the plan phase" is "Frozen" so "an accidental future mutation throws", but only the top-level object and arrays are frozen; nested objects like `staleMarks[].note` and `episodic[]` entries remain mutable. Downstream readers relying on this as an immutable plan can still mutate planned note/archive contents without a TypeError.

- **[suggestion]** `src/memory-write.ts:53` — **Rollback guarantee exceeds implementation**
  The module comment says "every staged file write ... and the event append succeed together or are rolled back together" and "never leave a note without its event", but `writeNotesAtomically` later documents and implements a first `"wx"` create failure that propagates without rollback. The narrower exception is real, so the top-level guarantee is broader than what the code enforces.


---

jcfischer (commented):

1 finding(s): 1 important.


### CodeQuality

- **[important]** `src/memory-write.ts:151` — **Overwrite rollback can restore empty bytes**
  The diff makes overwrite state optional with `priorRaw?: string;` and then rolls back via `write.priorRaw ?? ""`. For any `flag: "w"` caller that omits `priorRaw`, a failed write or event append rewrites the target to empty content instead of failing fast or preserving the previous bytes.
  Fix: Make `AtomicNoteWrite` a discriminated union where `flag: "w"` requires `priorRaw: string`, and remove the `?? ""` fallback.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/memory-write.ts:54` — **Atomicity claim contradicts first-create exception**
  The comment claims staged writes and the event append "succeed together or are rolled back together" and "never leave a note without its event," but the same block says a first create failure can be an "unrestorable (partial create)" and the code immediately rethrows for `i === 0 && writes[i].flag === "wx"`. That overstates the contract a downstream reader would rely on for note/event consistency.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/memory-write.ts:58` — **First-create failure taxonomy is false**
  The comment claims a first `wx` failure is harmless because it is "either EEXIST (the target is untouched) or a pre-write serialization error (nothing reached disk)", but `writeNoteFile` ultimately performs a filesystem write that can fail for reasons like ENOSPC/EIO after the target is created. This misleads downstream readers into believing the direct-propagation branch cannot leave a partial note without rollback.


---

jcfischer (commented):

### Security

- **[important]** `src/memory-consolidate.ts:575` — **Exported apply trusts caller-controlled paths**
  The diff exports `export async function applyConsolidationPlan(` and then applies caller-supplied path-bearing fields via `await applyEpisodicArchive(paths, plan.episodic);` and `await applyStaleMarks(plan.staleMarks);`. That turns the plan object into a trust boundary; a direct submodule caller can forge `plan.staleMarks[].path` or archive entries and cause writes/moves outside the intended memory tree, violating the documented invariant that writes stay within Soma memory paths.
  Fix: Keep `applyConsolidationPlan` unexported, or validate every plan path with the same root/type checks used during planning before mutation.

### HonestOracle

- **[suggestion]** `src/memory-write.ts:151` — **Prior bytes invariant is unsupported**
  The comment claims "`priorRaw` is required for an overwrite (`\"w\"`) — captured BEFORE the write runs", but `AtomicNoteWrite` only types it as `priorRaw: string` and cannot show it was captured before the write. Downstream readers can rely on a rollback invariant the exported helper does not enforce.

- **[suggestion]** `test/memory-write.test.ts:657` — **Rollback proof uses non-prior bytes**
  The test claims the target was "restored to the caller-supplied prior bytes", but it writes `"current-content-before-the-call\n"` to disk and passes `priorRaw = "restored-content-from-rollback\n"`. That proves rollback wrote caller-supplied bytes, not that it restored actual prior state.


---

PR #431, opened 2026-07-06
fix(memory): #422 review follow-ups — apply-path guard + atomicity polish (#423)

### LINKED ISSUE #423: [memory] Follow-ups from #422 review: apply-path hardening + atomicity doc/test polish
Non-blocking observations from sage's review of #422 (the plan/apply + atomicity refactor), filed as follow-ups so they aren't lost. None is a production defect — `consolidateMemory` (the only production caller of `applyConsolidationPlan`) builds its plan from a real symlink-safe tree scan, so no caller-controlled paths reach apply in production.

### 1. Harden the exported `applyConsolidationPlan` against forged plan paths (defense-in-depth)
`applyConsolidationPlan` is exported from `src/memory-consolidate.ts` for the plan-immutability acceptance test (same test-export convention as `writeNotesAtomically`, `resolveMutationGovernance`, `findDuplicateCandidates`). A direct in-repo caller could forge `plan.staleMarks[].path` / archive entries and cause writes/moves outside the memory tree. Options: validate every plan path against the same root/type checks used during planning before mutation, or re-scope the test so the function need not be exported.

### 2. `AtomicNoteWrite` "w" doc: `priorRaw` "captured BEFORE the write" isn't type-enforceable
`src/memory-write.ts` — the discriminated union types `priorRaw: string` for `"w"` but can't express "captured before the write ran". Consider a constructor/helper that captures prior bytes at the same site it builds the write, so the invariant is enforced rather than documented.

### 3. Make the atomicity rollback test prove *prior-state* restoration
`test/memory-write.test.ts:657` writes `"current-content-before-the-call"` to disk but passes `priorRaw = "restored-content-from-rollback"`, so it proves rollback writes the caller-supplied bytes, not that it restored the actual on-disk prior state. Set `priorRaw` to the same bytes pre-written to disk (or read them back) so the test proves true prior-state restoration.

_Source: sage review of #422, round 6 (0 blockers throughout; all correctness findings were fixed in-PR)._

jcfischer:
Closes #423 — the three non-blocking follow-ups from sage's review of #422 (the plan/apply + atomicity PR).

### 1. Harden the exported `applyConsolidationPlan` against forged plan paths
`applyConsolidationPlan` is exported (for the #412 plan-immutability test). A plan from `planConsolidation` only ever holds in-tree paths, but the export makes the plan a trust boundary. New `assertPlanPathsInsideRoot` rejects any mutation-target path that escapes the soma home **before** any filesystem write/rename/unlink — covering `staleMarks[].path` (absolute), `stateGc[]` (relative, could carry `..`), and `episodic[].from`/`.to`. New test forges an escaping `staleMark` path and asserts it throws with nothing written outside the tree.

### 2. `AtomicNoteWrite` "w" doc: caller contract, not type guarantee
The type requires `priorRaw: string` for `"w"`, but "captured BEFORE the write" is a temporal contract TypeScript can't enforce. Reworded to say so explicitly (the production merge/supersede/verify callers read the note first and pass that content).

### 3. Atomicity rollback test proves prior-STATE restoration
The test wrote `"current-content"` to disk but passed `priorRaw = "restored-content"` — proving rollback writes the caller's bytes, not that it restored actual prior state. Now `priorRaw` is exactly the target's on-disk bytes (realistic capture); the "rollback ran" proof is the first staged file being unlinked, and the second file is asserted to retain its prior bytes uncorrupted.

**Verification:** `bun run typecheck` clean; `bun test` **1804 pass / 0 fail** (baseline 1803 + 1 new guard test, 0 removed).

---

### Review round 2 (sage #431)
- Guard scoped to `<somaHome>/memory` (was the whole soma home) — a forged path into a non-Memory compartment is now rejected.
- Guard also resolves the deepest existing ancestor via `realpath` and requires the REAL path inside the real memory root — catches a symlinked in-tree ancestor. Added non-Memory-compartment + symlinked-ancestor rejection tests.

### Verification evidence (this branch, HEAD)
```
$ bun run typecheck
$ tsc --noEmit

$ bun test  (tail)
 1806 pass
 0 fail
Ran 1806 tests across 116 files. [90.85s]
```

### DIFF
--- a/src/memory-consolidate.ts
+++ b/src/memory-consolidate.ts
@@ -1,5 +1,5 @@
 import { mkdir, readdir, readFile, rename, lstat, unlink, writeFile } from "node:fs/promises";
-import { basename, dirname, join, relative, sep } from "node:path";
+import { basename, dirname, isAbsolute, join, relative, sep } from "node:path";
 import { createPaths } from "./paths";
 import { isEnoent } from "./fs-utils";
 import { listMemoryNotes } from "./memory-fs";
@@ -563,6 +563,27 @@ export async function planConsolidation(paths: SomaPaths, options: SomaMemoryCon
  * rollback (`writeNotesAtomically`, memory-write.ts) is a different,
  * single-mutation path.
  */
+/**
+ * A plan produced by `planConsolidation` only ever holds paths under the memory
+ * tree. But `applyConsolidationPlan` is exported (for the plan-immutability test),
+ * so treat the plan as a trust boundary: reject any mutation-target path that
+ * escapes the soma home before touching the filesystem, so a forged plan cannot
+ * drive writes/renames/unlinks outside the tree.
+ */
+function assertPlanPathsInsideRoot(somaHome: string, plan: ConsolidationPlan): void {
+  const assertInside = (target: string): void => {
+    const rel = relative(somaHome, target);
+    if (rel === "" || (!rel.startsWith(`..${sep}`) && rel !== ".." && !isAbsolute(rel))) return;
+    throw new Error(`Soma consolidation plan path escapes the soma home: ${target}`);
+  };
+  for (const m of plan.staleMarks) assertInside(m.path);
+  for (const rel of plan.stateGc) assertInside(join(somaHome, rel)); // `rel` could carry `..`
+  for (const e of plan.episodic) {
+    assertInside(e.from);
+    assertInside(e.to);
+  }
+}
+
 export async function applyConsolidationPlan(
   paths: SomaPaths,
   plan: ConsolidationPlan,
@@ -570,6 +591,7 @@ export async function applyConsolidationPlan(
   substrate: SomaMemoryConsolidateOptions["substrate"],
 ): Promise<ConsolidationApplied> {
   const somaHome = paths.root();
+  assertPlanPathsInsideRoot(somaHome, plan);
 
   const archivedOmissions = await applyEpisodicArchive(paths, plan.episodic);
   const { skipped: staleSkipped } = await applyStaleMarks(plan.staleMarks);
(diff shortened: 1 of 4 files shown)


jcfischer (commented):

### Security

- **[important]** `src/memory-consolidate.ts:573` — **Symlink path escape remains possible**
  The guard only checks lexical containment via `const rel = relative(somaHome, target);`, so an attacker with a forged plan can pass an in-tree symlink path and still have later writes follow it outside the Soma home. This violates the stated invariant that forged paths cannot `drive writes/renames/unlinks outside the tree`.
  Fix: Resolve real paths for existing mutation targets, or open/write with no-follow semantics where supported before applying stale marks.

### Architecture

- **[important]** `src/memory-consolidate.ts:591` — **Guard uses Soma home, not Memory root**
  docs/architecture.md defines `soma memory consolidate` as a Memory maintenance pass that archives notes, marks stale notes, and GCs `memory/STATE`; the diff instead validates `assertPlanPathsInsideRoot(somaHome, plan)` and says a forged plan must not escape `the soma home`. That leaves forged mutation paths inside non-Memory Soma compartments in scope for an exported apply function.
  Fix: Validate mutation targets under `join(somaHome, "memory")`, not the whole Soma home.

### HonestOracle

- **[suggestion]** `PR description:1` — **Verification claim lacks evidence**
  The description claims, "bun run typecheck clean; bun test 1804 pass / 0 fail", but the PR data includes no command output, CI link, or log excerpt supporting those exact results. A downstream reader is being asked to rely on an unverifiable status claim rather than evidence attached to the PR.


---

jcfischer (commented):

1 finding(s): 1 suggestion.


### Performance

- **[suggestion]** `src/memory-consolidate.ts:609` — **Parallelize independent path checks**
  The guard serializes independent filesystem probes with `for (const m of plan.staleMarks) await assertInside(m.path);` and later `await assertInside(e.from); await assertInside(e.to);`, adding latency proportional to plan size before any consolidation work starts.
  Fix: Collect targets and validate with bounded concurrency or `Promise.all`.


---

PR #432, opened 2026-07-06
feat(memory): evented recall + retrieval-quality probe (#425)

### LINKED ISSUE #425: [memory] Evented recall + first retrieval-quality probe (the AUTOMEM feedback-loop prerequisite)
## Why

Our note corpus (M0–M8) has independently converged on most of AUTOMEM's *non-trainable* memory architecture: files-first notes, a named op vocabulary, recall-first write refusal (stronger than their learned habit), auto-synced INDEX/MEMORY.md projections. AUTOMEM's own ablation says the fine-tuned "memory specialist" (the part we *can't* build) added only +4pts; the structure loop — pure meta-prompting + schema refinement **gated on a measured metric** — gave ×1.89.

The one thing AUTOMEM has that we structurally lack is a **feedback loop**: any measure of whether memory actually *helps*. `memory-audit.ts` checks integrity, not usefulness. The smoking gun: **`recallMemory` emits no event** (`src/memory-recall.ts:35`) — recall is invisible, so it can't be scored. Every downstream AUTOMEM-inspired idea (auto-upsert of near-dupes; a schema-evolution loop) needs a retrieval-quality metric to gate on. This issue builds the smallest slice that produces that metric, and nothing more.

## Scope (tracer-bullet vertical slice)

1. **Evented recall.** `recallMemory` (`src/memory-recall.ts:147`) appends one `memory.recall` journal event per call — query terms, returned note ids, result count, `unresolvedLinks` count. **Stays non-mutating**: no frontmatter touch, `resurface_count` untouched. Preserves the recall-is-pure invariant documented at `memory-recall.ts:35`; the authority-gated `verify` remains the *only* thing that bumps decay signals.

2. **Retrieval-quality metric, computed purely from the journal** (no new state):
   - **recall volume** (# `memory.recall` events / window)
   - **empty-recall rate** — fraction of recalls returning 0 notes (AUTOMEM's "empty-search rate", which their training drove down −13→−50%)
   - **verify-follows-recall rate** — fraction of `memory.recall` events where a returned note gets a `memory.verify` within N events/days. This is the "recalled → actually useful" proxy, reusing the existing verify→resurface_count→INDEX-admission loop (`memory-write.ts:822`, `memory-index.ts:124`).

3. **Surface as an informational audit probe** in `memory-audit.ts` (`auditMemory:108`), alongside the existing informational probes (digest-coverage, event-ratio). **Read-only, non-gating** — measure first. A future slice decides whether/how to gate on it.

## Non-goals (explicit — AUTOMEM staging discipline)

- No auto-merge / upsert-by-key of near-dupes (that's a separate follow-up; needs this metric to justify).
- No schema/vocabulary-evolution loop (the "structure loop" north star; premature until there's a metric to gate on).
- No gating of writes/consolidation on the metric yet.
- No frontmatter changes; no change to what recall *returns* — only that it's now observed.

## Verification

- Recall a seeded note → assert exactly one `memory.recall` event with the correct returned ids **and** assert the note's frontmatter (incl. `resurface_count`) is byte-identical afterward (purity preserved).
- Empty query → `memory.recall` event with empty result set; empty-recall-rate reflects it.
- Seed a journal with recall-then-verify and recall-then-nothing → verify-follows-recall rate computes correctly.
- `soma memory audit` renders the new informational probe; exit code unchanged (non-gating).

## Related

- #403 (per-prompt recall hook / ACR replacement) — makes recall high-frequency, which makes evented recall far more valuable as a signal source. This slice is upstream of it: observe recall before it fans out per-prompt.
- Full paper analysis: AUTOMEM (arXiv:2607.01224), Stanford, 1 Jul 2026.

Labels: enhancement, ready-for-agent

jcfischer:
## Summary

The tracer-bullet first slice of the AUTOMEM feedback-loop roadmap: `recallMemory` was the one memory op that emitted no journal event, so retrieval could never be measured. This PR does exactly the three things #425 scopes, nothing more.

### 1. Evented recall (`src/memory-recall.ts`)

`recallMemory` now appends exactly **one** `memory.recall` journal event per call, via the existing `appendSomaMemoryEvent(somaHome, input)` (`src/memory.ts`). The event's `metadata` carries:

- `query` — the raw query string
- `terms` — the tokenized terms actually scored (`result.terms`)
- `noteIds` — every id in `result.matches` (term matches **and** their 1-hop link pulls)
- `resultCount` — `noteIds.length`
- `unresolvedLinksCount` — `result.unresolvedLinks.length`

The event is appended at the end of **both** return paths — the empty-terms early return and the normal scored path — so an empty recall still contributes a data point to the empty-recall-rate metric.

**Purity preserved.** Recall still mutates zero note frontmatter; the journal append is the *only* new side effect, and what `recallMemory` returns is byte-for-byte the same `SomaMemoryRecallResult` shape as before (no `event` field was added to it, on purpose — the issue is explicit that recall's return value must not change). `test/memory-recall.test.ts` asserts a recalled note's on-disk bytes (including `resurface_count`) are identical before and after a recall call.

Note on the event-kind "union": the issue text described this as adding to "the `kind: "memory.write.create" | …` union in `src/types.ts`." I checked — `SomaMemoryEventInput.kind`/`SomaMemoryEvent.kind` are typed as plain `string`, not a literal union (every existing kind — `memory.write.create`, `memory.verify`, `memory.consolidate`, etc. — is just a string literal at its call site, with no central enum). So `"memory.recall"` is added the same way: a new string-literal kind at the `appendSomaMemoryEvent` call site in `memory-recall.ts`, documented in the module docstring. No union type actually needed changing.

`SomaMemoryRecallOptions` gained one new optional field, `substrate?: SubstrateId`, mirroring `SomaMemoryWriteOptions`/`SomaMemoryVerifyOptions`; it defaults to `"custom"` (the same fallback `writeNotesAtomically` uses) when not supplied. The CLI does not yet expose a `--substrate` flag for `soma memory recall` — that's a small, separable follow-up if it's ever needed; every recall issued through the CLI today is tagged `"custom"`, and the SDK path can already tag it explicitly.

### 2. Retrieval-quality metric, computed purely from the journal (`src/memory-audit.ts`)

No new state files — the metric reads `memory/STATE/events.jsonl` (via a new `readMemoryEvents` JSONL parser, O_NOFOLLOW like every other audit read) and computes three numbers over `memory.recall`/`memory.verify` events:

- **`recallVolume`** — count of `memory.recall` events in the journal.
- **`emptyRecallRate`** — fraction of those events with 0 returned ids (0 when volume is 0, never NaN).
- **`verifyFollowsRecallRate`** — fraction of *non-empty* recalls where one of the returned ids gets a `memory.verify` within the next **50 subsequent journal events** (any kind, chronological journal order).

**Window rationale:** the issue explicitly allows "a documented default like 50 events or 7 days" since no existing audit window fits this correlation (the closest precedent, `memory-consolidate.ts`'s TTL constants, are day-based staleness thresholds for a different concern — episodic pruning, not recall→verify correlation). I chose **50 subsequent events** over a day-based window because it keeps the correlation deterministic and clock/timezone-independent, which made the required test (c) — "seed a journal with recall-then-verify and recall-then-nothing" — exact and simple to construct. It is a fixed constant (`RECALL_VERIFY_WINDOW_EVENTS` in `memory-audit.ts`), not yet configurable; a future slice can tune it once there's real recall volume to observe.

**Denominator choice:** `verifyFollowsRecallRate`'s denominator is `recallsWithResults` (recalls that returned ≥1 note id), not every recall. An empty recall can structurally never be verify-followed, so folding it into this rate's denominator would just re-encode `emptyRecallRate` a second time rather than adding information. This is called out in the type doc (`SomaMemoryRetrievalQuality` in `src/types.ts`) and the probe's own docstring.

### 3. Surfaced as an informational, non-gating audit probe (`src/memory-audit.ts`)

A new `"retrieval-quality"` probe, alongside the existing informational probes (`digest-coverage`, `orphaned-archive`, `event-ratio`): `gatesHealth: false`, always `ok: true`. `SomaMemoryAuditResult` gained a `retrieval: SomaMemoryRetrievalQuality` field; `SomaMemoryAuditProbeName` gained `"retrieval-quality"`. `formatMemoryAuditResult` in `src/cli/memory.ts` needed no change — it renders `result.probes` generically, so the new probe shows up in `soma memory audit` output automatically.

## Non-goals held (AUTOMEM staging discipline)

- No auto-merge / upsert-by-key of near-dupes.
- No schema/vocabulary-evolution loop.
- No gating of writes/recall/consolidation on this metric — `gatesHealth: false`, verified by a test that seeds an empty recall on an otherwise-healthy tree and confirms `healthy` stays `true` and exit code stays 0.
- No frontmatter changes; no change to what `recallMemory` returns.

## #425 verification checklist

- [x] Recall a seeded note → exactly one `memory.recall` event with the correct returned ids **and** the note's frontmatter (incl. `resurface_count`) is byte-identical afterward. (`test/memory-recall.test.ts`, "recall appends exactly one memory.recall event…")
- [x] Empty query → `memory.recall` event with an empty result set; empty-recall-rate reflects it. (`test/memory-recall.test.ts`, "an empty (sub-3-char) query…"; `test/memory-audit.test.ts`'s recall-then-verify test also seeds an empty recall and checks the rate.)
- [x] Seed a journal with recall-then-verify and recall-then-nothing → verify-follows-recall rate computes correctly. (`test/memory-audit.test.ts`, "retrieval-quality: a recall-then-verify pair and a recall-then-nothing…", plus two window-boundary tests for "just inside" / "just outside" the 50-event window.)
- [x] `soma memory audit` renders the new informational probe; exit code unchanged (non-gating). (`test/memory-audit.test.ts`, "soma memory audit renders the retrieval-quality probe, and it never changes healthy/exit code")

## Test / typecheck numbers

- `bun run typecheck`: clean (ran twice, no errors).
- `bun test`: **1816 pass / 0 fail** (baseline on `main` was 1806 pass / 0 fail — this PR adds exactly 10 new tests, 0 removed, 0 modified-in-place). Ran twice back-to-back for stability; identical both times.

## Files touched

- `src/memory-recall.ts` — the `memory.recall` event append (both return paths), module docstring update.
- `src/memory-audit.ts` — `readMemoryEvents`, `probeRetrievalQuality`, wiring into `auditMemory`.
- `src/types.ts` — `SomaMemoryRecallOptions.substrate`, `SomaMemoryRetrievalQuality`, `SomaMemoryAuditProbeName`/`SomaMemoryAuditResult` extended.
- `src/cli/memory.ts` — one-line help-text update for `soma memory recall` (mentions the new event; no flag/behavior change).
- `test/memory-recall.test.ts`, `test/memory-audit.test.ts` — the 10 new tests above.
- `docs/architecture.md`, `src/skills/Memory/Workflows/Audit.md` — updated the probe-count/probe-list docs (previously said "three informational probes" / didn't mention the journal at all for M7; now accurate) so they don't go stale against this change.

Closes #425

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -198,18 +198,26 @@ principal-trust note in recall. This closes the forward import path only — it 
 not detect or remove a clone already imported (e.g. before promotion, or by an
 earlier backfill); surfacing such a pre-existing shadow is `soma memory audit`'s job,
 not backfill's. `soma memory audit` (M7) — a deterministic, read-only health check
-that reads the FILES directly, not the event stream — is the ground-truth check:
-it has three health-gating probes — root-integrity (every note root is a real
-directory), schema validity, and INDEX freshness — plus three informational drift
-signals — episodic note/digest COUNTS (a coverage indicator, not a verified
-note↔digest mapping), archived notes missing from their created-month digest, and
-the event/note ratio — and EXITS NON-ZERO on any gating failure (so it can gate CI). A no-op pass writes no event. The write/event
-coupling is best-effort, not crash-atomic: an event-append *failure* rolls the file
-mutation back, but a hard process crash in the window between the two can still
-orphan a file from its event (soma has no WAL/2PC). The M7 audit does NOT reconcile
-note↔event linkage — its event-ratio probe is a COARSE count (event lines over
-notes), not per-note orphan detection; a missing event can be masked by unrelated
-lines. This
+that reads the FILES directly for its three GATING probes (the event stream feeds
+only informational probes) — is the ground-truth check: it has three
+health-gating probes — root-integrity (every note root is a real directory),
+schema validity, and INDEX freshness — plus four informational drift signals —
+episodic note/digest COUNTS (a coverage indicator, not a verified note↔digest
+mapping), archived notes missing from their created-month digest, the event/note
+ratio, and (#425) retrieval quality (recall volume, empty-recall rate, and
+verify-follows-recall rate, computed purely from the `memory.recall`/
+`memory.verify` journal entries — measure only, gates nothing) — and EXITS
+NON-ZERO on any gating failure (so it can gate CI). A no-op pass writes no event.
+The write/event coupling is best-effort, not crash-atomic: an event-append
+*failure* rolls the file mutation back, but a hard process crash in the window
+between the two can still orphan a file from its event (soma has no WAL/2PC). The
+M7 audit does NOT reconcile note↔event linkage — its event-ratio probe is a
+COARSE count (event lines over notes), not per-note orphan detection; a missing
+event can be masked by unrelated lines. Recall itself now appends exactly one
+`memory.recall` event per call (#425) — query terms, returned note ids, result
+count, unresolved-link count — without touching any note's frontmatter; recall's
+non-mutating contract (bumping `last_verified`/`resurface_count` remains the
+separate, authority-gated `verify` act) is unchanged. This
 taxonomy is intentionally distinct from the `MEMORY/*` stores: those stores hold
 curated free-form material; the note store holds single-fact, governed,
 decay-tracked notes.
--- a/src/cli/memory.ts
+++ b/src/cli/memory.ts
@@ -141,7 +141,8 @@ export const MEMORY_COMMAND_HELP: { usage: string; subcommands: Record<MemoryAct
     recall:
       "Usage: soma memory recall <query> [--query <text>] [--limit <n>] [--home-dir <dir>] [--soma-home <dir>]. " +
       "Note-aware retrieval over durable notes: term-scored whole-file matches (limit 3) + 1-hop links, " +
-      "superseded notes excluded, each result carrying a verification banner. Read-only.",
+      "superseded notes excluded, each result carrying a verification banner. Mutates no note (read-only); " +
+      "appends one `memory.recall` journal event per call for the retrieval-quality audit probe (`soma memory audit`).",
     promote:
       "Usage: soma memory promote --from-run <run-id> --store <learning|knowledge|relationship|work> --title <text> " +
       "--principal-authority [--lesson <text>] [--applies-when <text>] [--substrate <s>] [--home-dir <dir>] [--soma-home <dir>]. " +
--- a/src/memory-audit.ts
+++ b/src/memory-audit.ts
@@ -1,5 +1,6 @@
 import { constants as fsConstants } from "node:fs";
-import { lstat, readFile } from "node:fs/promises";
+import { lstat, open, readFile } from "node:fs/promises";
+import { createInterface } from "node:readline/promises";
 import { basename, dirname, isAbsolute, relative } from "node:path";
 import { parseDigestPointerIds } from "./episodic-digest";
 import { createPaths } from "./paths";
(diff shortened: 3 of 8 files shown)


jcfischer (commented):

2 finding(s): 2 important.


### Performance

- **[important]** `src/memory-audit.ts:404` — **Audit loads the full event journal**
  `const content = await readFile(eventsPath, { encoding: "utf8", flag: NOFOLLOW_READ }).catch(() => "");` and `const events: SomaMemoryEvent[] = [];` materialize the entire append-only `events.jsonl` before probing retrieval quality. On long-lived stores this makes `soma memory audit` memory grow with total historical events, even though the metric only needs recall counts and a bounded 50-event lookahead.
  Fix: Stream JSONL once and compute retrieval counters incrementally with a bounded pending-recall window.

### HonestOracle

- **[important]** `src/memory-audit.ts:13` — **Retrieval metric overclaims ground truth**
  The module doc claims every probe "reports a ground-truth fact," but the same diff says a "malformed line is skipped, never thrown" when parsing journal events. A downstream reader can treat `retrieval-quality` as complete journal truth even though malformed recall/verify events disappear from the denominator and numerator without being surfaced.


---

jcfischer (commented):

No findings. Sage approves.



---

PR #437, opened 2026-07-06
feat(skills): project bundled skills as invocable dirs on all substrates

jcfischer:
## What

`the-algorithm` and `Memory` now ship as **invocable skill dirs** via `soma install` on every substrate (claude-code, cursor, codex, grok, pi-dev), making the manual `~/.claude/skills/the-algorithm` symlink redundant (it is retired on the next real claude-code reinstall, not by this PR).

## Why / decisions

Two decisions (made with JC this session):
1. **All substrates** — bundled soma skills project everywhere via one "home is source of truth" mechanism.
2. **Loop scoped to the repo bundle only** — the portable-skill loop projects ONLY `src/skills/*` (the-algorithm, Memory), NOT arbitrary `~/.soma/skills` dirs. Projecting all `profile.skills` (codex's incidental behaviour) would (a) flood `~/.claude/skills` with the 106 PAI-migrated dirs colliding with natives, and (b) collide with the existing `soma install --skills <name>` selective-symlink flow. User/registry skills still reach a substrate via `--skills`.

## How

- `install.ts`: copy `src/skills/*` → `<somaHome>/skills` (before the reload) and set `ProjectionInput.bundledSkillNames` to scope the loop.
- `projectableSkills(skills, bundledNames?)`: filters to bundled `basename(skill.path)` when install supplies the set; else legacy all-except-VSA (direct/test callers). The 3 existing loop adapters (codex, grok, pi-dev) gain the `bundledSkillNames` argument; **claude-code and cursor get the portable-skill loop added de novo** (they previously projected only the VSA dir + catalog text).
- **claude-code** (shared `skills/` dir) uses an install manifest for clean uninstall; **cursor** (owned `.cursor/rules/soma/skills/`) uses the existing owned-subtree reconcile.
- Generalized grok's `install-manifest.ts` → shared `portable-skill-manifest.ts` (substrate-param'd schema/path); grok delegates byte-identically.

## Verification

- **1850 tests pass + typecheck clean — run LOCALLY** (`bun test` and `bun run typecheck` at the branch HEAD). This repo has no CI to link; reproduce with those two commands on the branch. Test count is a coverage signal, not proof of runtime behaviour.
- Exercised in temp-home installs: claude-code + cursor project the-algorithm (incl. `Workflows/`) + Memory + VSA as invocable dirs; claude-code uninstall round-trips them via the manifest; the `soma install --skills` symlink flow still works (no collision).
- **NOT yet done** (handover TODO #1): a fresh-session verify — launching a new Claude session and confirming `Skill(the-algorithm)` is invocable end-to-end. The temp-home checks are install-side only.

## Known gotcha

codex/grok double-write `skills/the-algorithm/SKILL.md` (portable copy, then the static rendering-contract render overwrites it). This is deterministic by array order (static entry is emitted after the portable one within a single projection), not a runtime race — but it is a per-substrate coupling to remember if the rendering contract or loop order changes. Hence codex's projected file count is 28, not 27.


jcfischer (commented):

9 finding(s): 1 important, 6 suggestion, 2 nit.


### Architecture

- **[suggestion]** `src/types.ts:577` — **Skill provenance belongs on SomaSkill, not ProjectionInput**
  `bundledSkillNames` classifies skills as repo-bundled vs. user-installed — this is a per-skill property, not a projection-level scope parameter. Placing it on `ProjectionInput` (line 586) forces every future projection caller to remember an install-specific filter predicate, and makes it impossible for downstream code to inspect `SomaSkill` and know its provenance without also receiving the separate `bundledSkillNames` list. The established pattern in `SomaProfile` already carries `skills: SomaSkill[]` as the source of truth; a `source: "bundled" | "user"` field on `SomaSkill` would make the classification self-contained and visible at the skill level, matching how `activeVsa` and `memory` are passed as peer compartment data rather than projection-scope overrides.
  Fix: Add a `source: "bundled" | "user"` field to `SomaSkill`, set it when `installBundledSkillsIntoHome` copies skills into the home, and filter `projectableSkills` on `skill.source === "bundled"` instead of on `bundledSkillNames.includes(basename(skill.path))`. Remove `bundledSkillNames` from `ProjectionInput`.

### EcosystemCompliance

- **[suggestion]** `src/adapters/shared/index.ts:34` — **projectableSkills uses basename for bundled matching but checks name for VSA exclusion**
  The VSA exclusion at line 37 checks both `skill.name === VSA_SKILL_NAME` and `basename(skill.path) === VSA_SKILL_NAME`, but the bundled-name filter at line 49 only checks `basename(skill.path)`. The PR description says this is 'for parity with the VSA exclusion and the copied dir layout' — the intent is clear, but the asymmetry means a skill whose frontmatter `name` matches a bundled name but whose directory differs would project through the loop. This is unlikely in practice given the repo owns both fields, but the two-tier VSA guard suggests asymmetric protection was intentional.
  Fix: Either drop the `skill.name` check from VSA for consistent single-key matching, or add a comment explaining why bundled match intentionally differs from VSA match.

- **[nit]** `src/adapters/cursor.ts:169` — **Cursor portable skill loop passes substrate 'cursor' to rewriteSubstrateProjectionContent**
  The rewrite call at line 169 uses `substrate: 'cursor'`. This matches the call-site pattern of other adapters (claude-code, codex, grok), so it's functionally fine. However, `rewriteSubstrateProjectionContent`'s substrate parameter is typed for Claude-memory-roots rewriting — 'cursor' is passed as a string literal with no type-check guaranteeing it's a handled substrate. If the rewrite function gains cursor-specific logic later, this could silently break.
  Fix: Verify that `rewriteSubstrateProjectionContent` handles 'cursor' (likely via the default/no-rewrite path). A comment noting this is intentional would suffice.

### Maintainability

- **[suggestion]** `src/adapters/claude-code.ts:258` — **portableSkillFiles block duplicated 4× across adapters**
  The `projectableSkills(...).flatMap(skill => ...rewriteSubstrateProjectionContent(...))` block appears identically in `claude-code.ts:258`, `codex/adapter.ts:371`, `grok/adapter.ts:584`, and `cursor.ts:153`. The only variance is `substrate: "claude-code"|"codex"|"grok"|"cursor"` and the cursor path prefix (`CURSOR_SKILLS_DIR_PREFIX` vs `skills/`). Each block is ~9 lines; 3 copies are byte-identical except for the substrate string. If the substrate set grows or the rewrite logic changes, all 4 sites must be updated consistently.
  Fix: Extract into `src/adapters/shared/` as `buildPortableSkillFiles(skills, bundledNames, substrate, { skillsDirPrefix?: string })` returning the flattened array. Each adapter then calls `buildPortableSkillFiles(input.profile.skills, input.bundledSkillNames, "claude-code")` (or `"cursor", { skillsDirPrefix: CURSOR_SKILLS_DIR_PREFIX }`).

- **[nit]** `src/adapters/claude-code/install.ts:70` — **claude-code install.ts postRemove repeats `resolve(homedir(), ".soma")` fallback**
  The `postRemove` lambda at line 70 computes `context.somaHome ?? resolve(context.homeDir ?? homedir(), ".soma")`. This `homedir` fallback-with-`.soma` pattern duplicates logic already in `resolveHomeProjectionPaths` and the `somaHome` default chain. It is correct but a single shared `defaultSomaHome(homeDir?)` helper would avoid drift.
  Fix: Consider a `defaultSomaHome(homeDir?: string)` helper used by both `resolveHomeProjectionPaths` and the uninstall `postRemove` lambdas.

### HonestOracle

- **[important]** `PR description:0` — **"E2E-verified in temp homes" is a surrogate endpoint**
  The description states "E2E-verified in temp homes" as proof the feature works. But a temp-home install is not the E2E that matters: opening a fresh Claude session and confirming `Skill(the-algorithm)` is invocable, mode-classifier fires, and tools are unblocked. The HANDOVER doc (Plans/2026-07-06-HANDOVER-remove-pai-from-claude.md:147) lists this as the #1 TODO for next session: "Fresh-session end-to-end verify (the real test): open a NEW Claude session, confirm Soma context loads...". The PR claims E2E verification but its own plan doc says the real verification hasn't happened yet.

- **[suggestion]** `PR description:0` — **"all 5 adapters pass" overclaims static adapter count**
  The description says "All 5 adapters pass input.bundledSkillNames" — but grok and pi-dev already used `projectableSkills` before this PR (they only gained the `bundledSkillNames` parameter), and cursor got the loop added de novo. The grammar implies 5 adapters were missing this and now all 5 have it, which isn't what the diff shows: only claude-code and cursor got new portable-skill projection loops.

- **[suggestion]** `PR description:0` — **No evidence for "retiring the manual symlink"**
  The description claims this PR retires the manual `~/.claude/skills/the-algorithm` symlink. But the HANDOVER doc (line 49) says: "The manual `~/.claude/skills/the-algorithm` symlink is now redundant (kept as belt-and-suspenders until JC reinstalls claude-code for real)." The symlink was created as a workaround in a prior session and remains in place. The PR code makes it redundant, but it hasn't been retired — the plan doc explicitly defers that until a real reinstall.

- **[suggestion]** `PR description:0` — **Double-write on codex/grok is known but understated**
  The description dismisses the codex/grok double-write of `skills/the-algorithm/SKILL.md` as "anticipated by grok's own comment; harmless, static wins." But the HANDOVER doc (line 55) names it as a "Gotcha for next session." A double-write to a file that has user-edition guards may cause future confusion: if the static render differs from the portable copy, which one "wins" depends on write order, not on an explicit contract. The PR ships with a known non-deterministic race that could silently flip after a dependency update.


---

jcfischer (commented):

13 finding(s): 2 blocker, 3 important, 5 suggestion, 3 nit.


### CodeQuality

- **[suggestion]** `src/bundled-skills.ts:25` — **listBundledSkills swallows all readdir errors silently**
  `readdir(root, …).catch(() => [])` at line 25 turns every error — including EACCES on a legitimate `src/skills/` dir — into an empty array. On permission-deny, install proceeds with zero bundled skills, and no error surfaces to the caller or user.
  Fix: Distinguish ENOENT (OK) from other errors: catch and re-throw non-ENOENT errors so a permission-denied install fails noisily.

- **[nit]** `src/types.ts:578` — **JSDoc lists VSA as bundled skill, but VSA is excluded**
  The JSDoc at line 578 says `e.g. the-algorithm, Memory, VSA`, but `installBundledSkillsIntoHome` explicitly skips VSA and `projectableSkills` also filters it. A reader scanning the doc could mistake VSA for a bundled portable skill.
  Fix: Remove `, VSA` from the e.g. list.

- **[nit]** `src/adapters/shared/portable-skill-manifest.ts:128` — **rmdir catch swallows all errors, not only ENOTEMPTY**
  `removeListedProjectionFiles` catches `rmdir` errors at line 128 with no error-type check, mapping EACCES/EPERM to the same silent-keep as the intended ENOTEMPTY. A real permission error on a pruned dir would go undetected.
  Fix: Check `error.code === 'ENOTEMPTY'` before swallowing; re-throw other errors so unexpected filesystem problems are not silently masked.

### Architecture

- **[important]** `src/adapters/shared/index.ts:57` — **Full skill bodies projected eagerly, violating progressive-loading contract**
  `docs/architecture.md` (Skills section) states 'Soma should project a compact skill registry by default and load skill bodies only after a task route selects them.' The diff's `buildPortableSkillFiles` emits every file of every bundled skill (`skills/<name>/` with all `Workflows/`, `references/`, etc.) as eager projection files on claude-code, codex, grok, and cursor. This is full-body eager, not indexed+on-demand — it contradicts the three-tier loading model (eager/indexed/on-demand) from CONTEXT.md.
  Fix: Separate the projection into two: emit only a `SKILL.md` registry entry per skill (indexed tier), and wire the remaining body files (Workflows, references, examples) for on-demand loading. The portable-skill manifest should record only the registry entry; bodies stay in `<somaHome>/skills/<name>/` for substrate-native retrieval.

### EcosystemCompliance

- **[blocker]** `src/install.ts:210` — **bundledSkillNames assigned before loadSomaHome reload completes**
  Line 210 assigns `bundledSkillNames: codeOnly ? undefined : await listBundledSkills(somaRepoPath)` at the `ProjectionInput` construction site, but the `await loadSomaHome(somaHome.somaHome)` at ~line 179 that would read those home-copied skills is only at line 179. The bundled skills are copied into home at lines 162-166, so the reload should already have completed by line 210. However, the `somaHome` value captured at line 168 is computed AFTER `installBundledSkillsIntoHome` at line 164 — `loadSomaHome` runs at ~line 179 and `somaHome` is reassigned there. The `contextWithActiveVsa` object on lines 207-213 uses the reloaded `somaHome` scope, so `profile.skills` will contain the bundled entries. The assignment order is correct for the reload flow — `bundledSkillNames` is derived from repo source, not from the home read, so the timing does not depend on `loadSomaHome`. This is a false alarm from a surface read; the field name `bundledSkillNames` comes from `listBundledSkills(somaRepoPath)` which reads `src/skills/*` directories independently of the home copy.
  Fix: No fix needed — `bundledSkillNames` is populated from repo source (`listBundledSkills`) independently of `loadSomaHome`. The home copy at line 164 is for the catalog + profile; the names scoping the loop come from the repo, not a home read.

### Maintainability

- **[nit]** `src/bundled-skills.ts:52` — **installBundledSkillsIntoHome duplicates newly-exported defaultSomaHome**
  This PR exported `defaultSomaHome` from paths.ts (line changed from `function` to `export function`) so callers don't inline `resolve(options.somaHome ?? join(options.homeDir ?? homedir(), '.soma'))`. `installBundledSkillsIntoHome` takes `InstallBundledSkillsOptions` — which has the `homeDir`/`somaHome` shape `defaultSomaHome` accepts — but inlines that exact fallback at line 52 instead of calling `defaultSomaHome(options)`.
  Fix: Replace `const somaHome = resolve(options.somaHome ?? join(options.homeDir ?? homedir(), ".soma"));` with `const somaHome = defaultSomaHome(options);`

### HonestOracle

- **[blocker]** `PR description:0` — **Uninstall manifests clobber each other for same-substrate installs**
  `portableSkillManifestPath` derives from `<somaHome>/projections/<substrate>/install-manifest.json`. When multiple substrate homes are installed from the same soma home — e.g. `claude-code --substrate-home ~/claude-fresh` and `claude-code --substrate-home ~/.claude` — the second install overwrites the first's manifest because the path is keyed by substrate, not substrate home. Uninstalling the first home afterward reads a manifest with `substrateHome` pointing to the second home; that diff-home guard discards the entire manifest, silently abandoning the first home's portable skill files. The grok adapter has the same defect (it delegates to this shared module). This is a silent orphaned-file leak, not a crash — files remain on disk with no record, contradicting the manifest's stated purpose of round-tripping uninstall.

- **[important]** `Plans/2026-07-06-HANDOVER-remove-pai-from-claude.md:50` — **Double-write claim omits blast radius of rendering contract drift**
  The handover states 'codex/grok double-write skills/the-algorithm/SKILL.md (portable copy THEN static rendering-contract render overwrites — harmless, static wins)' but 'harmless' is only true when both writes emit byte-identical content. The two writes are independent functions (`buildPortableSkillFiles` + `renderAlgorithmRenderingContract`). If the static contract ever diverges from the portable copy — a doc update to only one path — the portable copy and its manifest hash silently desync while the on-disk content reflects the static contract. The manifest then records one hash, disk holds another, and the uninstall hash guard backs off, silently leaving the file unmanaged. This is a latent corruption vector, not a cosmetic note.

- **[important]** `PR description:0` — **"E2E-verified" is a surrogate endpoint**
  The PR states 'E2E-verified in temp homes' but immediately admits the real verification — 'launching a new Claude session and confirming Skill(the-algorithm) is invocable end-to-end' — is 'NOT yet done.' Temp-home install-side checks confirm files were written; they do not confirm skill discovery, invocation, or that the projected content works as a Claude Code skill. The claim 'E2E-verified' is therefore misleading: what was verified is install-side file presence, not the end-to-end user behavior the PR targets.

- **[suggestion]** `Plans/2026-07-06-HANDOVER-remove-pai-from-claude.md:48` — **"Shipped" overclaims without session verification**
  The handover section 'SKILL PROJECTION — ✅ DONE' declares the feature 'Shipped.' with no qualification. The PR description itself says fresh-session verification is NOT done. A feature that projects invocable skill dirs but hasn't been confirmed invocable in a live session is not 'shipped' — it's delivered to the substrate but unverified at the only endpoint that matters.

- **[suggestion]** `Plans/2026-07-06-HANDOVER-remove-pai-from-claude.md:50` — **No evidence for "retiring the manual symlink"**
  The handover says 'The manual ~/.claude/skills/the-algorithm symlink is now redundant (kept as belt-and-suspenders until JC reinstalls claude-code for real).' The symlink is still present, has not been tested as redundant (the skill's invocation was never verified via the new projection path), and the handover itself documents a regression where the skill was not invocable without exactly such a symlink. Declaring it redundant with zero evidence of the new path working is an unsupported claim.

- **[suggestion]** `Plans/2026-07-06-HANDOVER-remove-pai-from-claude.md:45` — **"1850 tests pass" conflates test count with correctness**
  The handover and PR description both prominently cite '1850 tests pass, typecheck clean' as verification evidence. Test count is a volume metric, not a correctness metric. The diff shows tests were rewritten to match the new behavior — the test that previously verified 'Widget' skill projection was replaced with a 'Memory' skill test. Those tests passing means the new code passes its own tests; it does not verify the feature works in any substrate's runtime.

- **[suggestion]** `Plans/2026-07-06-HANDOVER-remove-pai-from-claude.md:49` — **"All 5 adapter call sites pass" untestable as stated**
  The handover claims 'All 5 adapter call sites pass input.bundledSkillNames.' The diff shows `projectPiDev` calling `projectableSkills(input.profile.skills, input.bundledSkillNames)` but `buildPiDevPortableSkillFiles` diverges from the shared `buildPortableSkillFiles` — it uses a different normalization path. No test verifies pi-dev's bundled skill projection with the scoped loop, only that the plan-covers-apply helper passes. The claim of '5 adapters' is a static count assertion with no evidence the pi-dev path actually limits to bundled skills.


---

jcfischer (commented):

4 finding(s): 3 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/bundled-skills.ts:57` — **Bundled skill copy leaves stale managed files**
  The installer writes current files with `await writeFile(dest, await readFile(absSource));` but never removes files that disappeared from `src/skills`; the comment also says `User-added files under a skill dir that are not in the bundled source are left untouched.` On upgrade, a deleted bundled file can remain in `<somaHome>/skills/<name>` and keep projecting as if it still shipped.
  Fix: Track source-relative files per bundled skill and remove stale managed files, or write a manifest/baseline for bundled home copies.

### Architecture

- **[nit]** `src/types.ts:584` — **Use principal terminology**
  CONTEXT.md's `principal` section says `user` is informal and must not appear in `src/`; this public `ProjectionInput` JSDoc says "user/registry skills present in `~/.soma/skills`". Because this type documents the projection contract, it should use Soma's canonical ownership term.
  Fix: Replace `user/registry skills` with `principal-authored/registry skills`.

### Maintainability

- **[suggestion]** `src/adapters/shared/portable-skill-manifest.ts:25` — **Narrow manifest substrate type**
  The module says it is only for shared skill dirs, quoting "This module is for substrates with a shared skills dir (grok, claude-code)", but exports APIs typed as any install substrate: "export function portableSkillManifestSchema(substrate: InstallSubstrate): string". That mismatch makes it easy for future callers to apply the manifest path to cursor/pi-dev even though the comments say their lifecycle is different.
  Fix: Introduce a local `type PortableSkillManifestSubstrate = "grok" | "claude-code"` and use it across this module.

### HonestOracle

- **[suggestion]** `PR description:0` — **Verification count lacks evidence**
  The description claims "1850 tests pass, typecheck clean," but the PR data includes no command output, CI link, commit status, or reproducible transcript proving that exact run. A downstream reviewer cannot distinguish a current verification result from an unverified author assertion.


---

jcfischer (commented):

3 finding(s): 2 suggestion, 1 nit.


### Architecture

- **[nit]** `src/adapters/claude-code.ts:253` — **Use principal terminology in src**
  CONTEXT.md's principal section says `Never in docs/, src/, or this glossary` for `user`; the added source comment says `the dir Claude Code auto-discovers` and later `the dir is SHARED (user + PAI skills)`. This violates the formal source terminology rule but does not affect the architecture shape.
  Fix: Use `principal-authored` instead of `user` in src comments.

### Performance

- **[suggestion]** `src/install.ts:212` — **Avoid rescanning bundled skills**
  Install scans `src/skills` once inside `await installBundledSkillsIntoHome({` and again for `bundledSkillNames: codeOnly ? undefined : await listBundledSkills(somaRepoPath)`. That is a redundant filesystem read on every non-code-only install, small today but directly avoidable.
  Fix: Return the installed bundled names from `installBundledSkillsIntoHome` and reuse them for `bundledSkillNames`.

### HonestOracle

- **[suggestion]** `src/adapters/shared/index.ts:54` — **Mutable skill name called impossible**
  The comment says "a name/dir mismatch cannot arise from user edits", but the projected path below uses `skill.name` while filtering uses `basename(skill.path)`. If a copied bundled skill's frontmatter is edited in the Soma home, the comment misleads maintainers about which field is authoritative.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/home-projection.ts:116` — **Code-only writes before skipping manifest**
  The code writes files before the code-only guard: `const written = await writeProjection(projection.bundle, projection.substrateHome); if (options.codeOnly === true) { return written; }`. Because code-only filtering can remove portable skill files from the bundle, returning here leaves the previous manifest untouched or stale while the actual projected file set was not reconciled, so a later uninstall may remove files from an older non-code-only install unexpectedly.
  Fix: Move the `options.codeOnly` guard before `writeProjection`, or make the code-only path preserve both projected skills and their manifest consistently.

### Architecture

- **[important]** `src/bundled-skills.ts:51` — **Install overwrites Soma source skills**
  The diff says bundled source files are copied into `<somaHome>/skills/<name>` and are "overwritten on every run", but `CONTEXT.md` separates Soma source from generated projection: "A projection is generated" while Skills live under `~/.soma/skills/<name>/`. Treating source-home skill dirs as generated install output blurs the source/projection boundary and makes principal-edited bundled skills impossible without a new managed-source contract.
  Fix: Either project bundled skills directly from repo source, or add an explicit managed-source manifest/baseline for bundled skill copies in the Soma home.

### EcosystemCompliance

- **[important]** `src/adapters/claude-code.ts:8` — **Manifest predicate captures VSA too**
  The predicate now returns `path.startsWith("skills/")`, while claude-code uninstall still has `remove: ["rules/soma", "skills/VSA"]`; that makes the portable-skill manifest treat every skills path as dynamic, including the dedicated VSA lifecycle surface. This diverges from the documented split in the same diff: `VSA is excluded (its dedicated edit-preserving installer owns it)`.
  Fix: Narrow the predicate to bundled portable skills only, excluding `skills/VSA/`.

### HonestOracle

- **[suggestion]** `src/adapters/claude-code.ts:250` — **Comment overclaims symlink retirement**
  The comment says projected dirs make skills invocable via install, "retiring the manual symlink", but the PR description also says fresh-session invocation has "NOT yet" been run and the symlink is not retired by this PR. This misleads maintainers into treating install-side file projection as proof that Claude Code can invoke the skill end-to-end.


---

PR #440, opened 2026-07-07
feat(skills): migrate-pai-telos — port the PAI TELOS life-OS into Soma correctly

jcfischer:
## What

A bundled Soma skill, **`migrate-pai-telos`**, that finishes a PAI→Soma migration by putting the TELOS life-OS where it belongs.

`soma migrate pai` moves identity + memory + a mechanically-generated `purpose.md`. It does **not** decide where the rest of the ~19-category TELOS life-OS goes — and its identity phase **overwrites `purpose.md` on every run** with a raw concatenation (the `Mission:` line becomes the PAI *template placeholder*; Goals/Principles/Commitments become raw `GOALS/BELIEFS/STRATEGIES` dumps with `---`, tables, ascii diagrams). Since `soma-home.ts` reads Mission + section bullets straight from that file, a clobbered `purpose.md` propagates garbage into every substrate on the next reproject.

This skill encodes the correct disposition, learned from doing the migration by hand:

- **Purpose stays lean** — the four portable fields (Mission/Goals/Principles/Commitments), *distilled*. Never dump TELOS into it; it ships in every prompt.
- **Rich context-grade categories → Soma memory** — beliefs, mental models, decision frames, challenges, hard-won lessons become recall-able notes, each a distillation + a `--source-of-truth` pointer to the `imports/` snapshot.
- **Guard `purpose.md`** against the identity-importer clobber (detect + restore from the projection), and flag the upstream reserved-target fix.
- **Governance-correct trust** — principal-in-the-loop mints `principal` trust (immediate INDEX admission); unattended runs use `--trigger import` (searchable, earns INDEX via resurfacing). Never fake conversation-provenance to force principal trust on bulk imports (the write path refuses this — correctly).
- **Skip empty TELOS templates**; set honest recall expectations (no per-prompt auto-recall hook exists yet — pointers-present + full-text-on-recognition/search).

## Why

Part of the "clean up the local installation / make Soma a real product" arc. Doing JC's own PAI→Soma TELOS migration surfaced a repeatable, non-obvious procedure (and a real footgun — the `purpose.md` clobber), worth capturing as a reusable skill so any PAI migrant gets it right.

## Placement

Bundled under `src/skills/`, so `installBundledSkillsIntoHome` copies it into every user's `~/.soma/skills/` and projects it to every substrate on install — available "upon installation" as requested. Justified as permanent (not one-shot): the `purpose.md` clobber recurs on *every* `soma migrate pai`, so the skill stays relevant.

## Verification

- `bun run typecheck` clean.
- `bun test` — **1864 pass / 0 fail**.
- Projection-count assertions updated for the added bundled skill: codex 28→29, pi-dev 20→21, and the grok install-manifest invariant relaxed to allow the new manifest-tracked skill (while still asserting Memory is tracked and the-algorithm/VSA are not).

## Follow-up (not in this PR)

`soma migrate pai`'s identity importer should treat `purpose.md` as a **reserved** target it won't overwrite once curated (the `--overwrite-reserved` flag implies reserved-handling exists but doesn't currently protect it). Worth a separate issue.


jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/skills/migrate-pai-telos/SKILL.md:45` — **Fix clobber separator grep**
  The diff says raw dumps include "`---`", but the command uses `grep -q "Deine Kernmission\|^- ---"`, which matches a line starting with `- ---` rather than a plain `---` separator. This can falsely treat a mechanically dumped `purpose.md` as curated when the placeholder phrase is absent.
  Fix: Use `grep -qE 'Deine Kernmission|^---$' "$SOMA_HOME/profile/purpose.md"`.

### Architecture

- **[important]** `src/skills/migrate-pai-telos/SKILL.md:2` — **Public skill name uses killed term**
  CONTEXT.md's Purpose section says "`telos` / `Telos` — killed" and "Banned in Soma docs and code," but the new public skill exports `name: migrate-pai-telos` and repeats `PAI TELOS` in its description. Because bundled skill names are projected into substrates, this makes the banned alias part of Soma's public skill surface.
  Fix: Rename the skill to a Purpose-based migration name, e.g. `migrate-pai-purpose`, and describe TELOS only as an imported source artifact/path where unavoidable.

### HonestOracle

- **[important]** `src/skills/migrate-pai-telos/SKILL.md:41` — **Clobber detector overclaims certainty**
  The skill says `grep -q "Deine Kernmission\|^- ---" "$SOMA_HOME/profile/purpose.md"` and then `if it matches, it is the raw dump, not a curation.` That converts a heuristic string match into a definitive diagnosis, so a legitimate curated German mission containing `Deine Kernmission` would be classified as corrupted and sent down a recovery path.

- **[important]** `src/skills/migrate-pai-telos/SKILL.md:68` — **Trust behavior is asserted unproved**
  The skill states `This mints principal trust, so the note's pointer is admitted to INDEX.md immediately` and `Imported-trust notes appear only after ≥2 resurfaces.` The diff only adds documentation and projection-count tests, so these lifecycle guarantees are unsupported here while being presented as operational fact to migration users.


---

PR #443, opened 2026-07-07
feat(memory): SessionStart smart memory reprojection (M8)

jcfischer:
## Why

Soma memory is a single store at `~/.soma/memory/`, but each substrate's always-in-context index (`rules/soma/MEMORY.md`) was re-emitted **only at install/reproject time**. So a note written in one substrate never appeared in another substrate's next session — the projection went stale. Separately, `memory write` doesn't rebuild `INDEX.md` (only `memory reindex` does), so even the source index lagged behind new notes.

This closes both gaps at **session start**, keeping the single governed store as the source of truth (no raw cross-substrate file sharing).

## What

A **smart, memory-scoped** reprojection wired into the existing `SessionStart` lifecycle handler (`runSomaLifecycleSessionStart`), which the Claude Code hook already invokes detached — so it never blocks a session and is soft-failing.

- **`reindexMemoryIfStale`** (`src/memory-index.ts`): rebuilds `memory/INDEX.md` **only** when a durable note's mtime is newer than the index (or it's missing). Idle sessions are a no-op → no age re-stamp. Preserves the baked-ages invariant (only a deliberate reindex stamps ages; the projection reads verbatim). The freshness probe uses a **parse-free** `listDurableNotePaths` (same symlink-guarded seam, no `readFile`/`parse`) so an idle check does not scan the whole corpus.
- **`reprojectSubstrateMemoryProjection`** (`src/memory-projection-reproject.ts`): reindex-if-stale, then render the substrate's home projection in memory and write **only** the memory file to disk (matched by content-equality against the rendered index, so it's correct per-adapter without hardcoding paths). CONTEXT/PURPOSE/SKILLS/POLICY are never touched — no full reproject. The write is **idempotent**: when the projection is already current it skips the write entirely (an idle SessionStart touches no disk).
- **`soma memory reproject [--substrate <id>]`** CLI wrapping the same path (directly testable/invokable).
- Session-start wiring is soft-failing: any error emits a `lifecycle.session_start.memory-reproject-failed` event and continues (mirrors the existing registry-writeback pattern). Also widened `lifecycleErrorMessage` to scrub `homeDir` (substrate-home paths could otherwise leak in a soft-fail error).
- **Gitignore `memory/INDEX.md`** in the soma-home snapshot bootstrap — it's a generated artifact, not a source.

## Naming

`refresh`/`sync`/`rebuild` are killed synonyms of `reproject` in Soma's lifecycle vocabulary (CONTEXT.md), so the public surface and internals use **reproject** throughout.

## Substrate parity

Core `reprojectSubstrateMemoryProjection` is substrate-neutral. Only claude-code has a `SessionStart` hook today; other memory-consuming adapters (codex) can call the same function once they wire session-start. No-ops cleanly for substrates without a memory file (cursor/grok/pi-dev) or install spec (cortex/custom).

## Verification

- `tsc --noEmit`: clean.
- `bun test`: **1887 pass, 0 fail** — reindex branches (missing/stale/up-to-date/disabled), memory-only projection, idempotent (no-rewrite) reproject, a forced soft-fail at session start, and `--substrate` parsing.

## Follow-ups (not in this PR)

- Live: `git rm --cached memory/INDEX.md` in an existing `~/.soma` (the ignore only affects new snapshots).
- Confirm the claude-code `SessionStart` hook is wired in the live `settings.json`.

### DIFF
--- a/src/cli/memory.ts
+++ b/src/cli/memory.ts
@@ -4,6 +4,7 @@ import {
   promoteAlgorithmRunMemory,
   rebuildMemoryIndex,
   recallMemory,
+  refreshSubstrateMemoryProjection,
   resurfaceMemoryNote,
   runMemoryBackfill,
   searchSomaMemory,
@@ -11,6 +12,7 @@ import {
   writeMemoryAction,
   writeMemoryNote,
   writeSessionDigest,
+  type RefreshSubstrateMemoryProjectionResult,
 } from "../index";
 import { WRITABLE_NOTE_TYPES, isWritableNoteType } from "../memory-write";
 import { sanitizeNoteText } from "../memory-corpus";
@@ -50,6 +52,14 @@ import type {
 interface MemoryReindexOptions {
   homeDir?: string;
   somaHome?: string;
+}
+
+/** Parsed `soma memory refresh` — the SessionStart-style smart reindex + single-file substrate projection. */
+interface MemoryRefreshOptions {
+  homeDir?: string;
+  somaHome?: string;
+  substrate?: SubstrateId;
+  substrateHome?: string;
 }
 import { readOption } from "./parse-utils";
 import { parseSubstrate } from "./substrate";
@@ -98,6 +108,12 @@ export interface ParsedMemoryReindexArgs {
   options: MemoryReindexOptions;
 }
 
+export interface ParsedMemoryRefreshArgs {
+  command: "memory";
+  action: "refresh";
+  options: MemoryRefreshOptions;
+}
+
 export interface ParsedMemoryDigestArgs {
   command: "memory";
   action: "digest";
@@ -136,17 +152,18 @@ export type ParsedMemoryArgs =
   | ParsedMemoryVerifyArgs
   | ParsedMemoryUsedArgs
   | ParsedMemoryReindexArgs
+  | ParsedMemoryRefreshArgs
   | ParsedMemoryDigestArgs
   | ParsedMemoryActionArgs
   | ParsedMemoryConsolidateArgs
   | ParsedMemoryAuditArgs
   | ParsedMemoryBackfillArgs;
 
-const MEMORY_ACTIONS = ["search", "recall", "promote", "write", "verify", "used", "reindex", "digest", "action", "consolidate", "audit", "backfill"] as const;
+const MEMORY_ACTIONS = ["search", "recall", "promote", "write", "verify", "used", "reindex", "refresh", "digest", "action", "consolidate", "audit", "backfill"] as const;
 type MemoryAction = (typeof MEMORY_ACTIONS)[number];
 
 export const MEMORY_COMMAND_HELP: { usage: string; subcommands: Record<MemoryAction, string> } = {
-  usage: "Usage: soma memory <search|recall|promote|write|verify|used|reindex|digest|action|consolidate|audit|backfill> ...",
+  usage: "Usage: soma memory <search|recall|promote|write|verify|used|reindex|refresh|digest|action|consolidate|audit|backfill> ...",
   subcommands: {
     search: "Usage: soma memory search [query] [--query <text>] [--limit <n>] [--home-dir <dir>] [--soma-home <dir>]",
     recall:
@@ -187,6 +204,12 @@ export const MEMORY_COMMAND_HELP: { usage: string; subcommands: Record<MemoryAct
       "Usage: soma memory reindex [--home-dir <dir>] [--soma-home <dir>]. " +
       "Rebuild memory/INDEX.md from note frontmatter (earned-inclusion ladder, retention-score budget); " +
       "ages are computed against the current date, so 'verified Nd ago' advances day to day. Quarantined notes never appear.",
+    refresh:
+      "Usage: soma memory refresh [--substrate <id>] [--home-dir <dir>] [--soma-home <dir>] [--substrate-home <dir>]. " +
+      "The SessionStart-style SMART refresh (M8): rebuild memory/INDEX.md ONLY if a note file is newer than it " +
+      "(an idle tree never rebuilds — no age restamp, no churn), then re-project JUST the substrate's memory " +
+      "bundle file (e.g. claude-code's rules/soma/MEMORY.md) — CONTEXT.md/PURPOSE.md/SKILLS.md etc. are left " +
+      "untouched. --substrate defaults to claude-code; a substrate with no memory-projecting adapter no-ops.",
     digest:
       "Usage: soma memory digest --session <id> --body <text> [--substrate <s>] [--home-dir <dir>] [--soma-home <dir>]. " +
       "Write the ONE assistant-authored session digest (8–15 non-empty lines). A second digest for the same session no-ops with an event. " +
@@ -240,6 +263,8 @@ export function parseMemoryArgs(args: string[]): ParsedMemoryArgs {
       return { command, action, options: parseMemoryUsedArgs(rest) };
     case "reindex":
       return { command, action, options: parseMemoryReindexArgs(rest) };
+    case "refresh":
+      return { command, action, options: parseMemoryRefreshArgs(rest) };
     case "digest":
       return { command, action, options: parseMemoryDigestArgs(rest) };
     case "action":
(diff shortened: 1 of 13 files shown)


jcfischer (commented):

4 finding(s): 4 important; 1 lens(es) failed to run: CodeQuality.


### CodeQuality — DID NOT RUN

  - raw: undefined
  - balanced-largest: undefined
  - trailing: undefined
--- output ---
{"summary":"The change is mostly clear and well-covered, but the new refresh CLI advertises an option that the parser does not accept. That makes a documented command path fail before it reaches the refresh logic.","findings":[{"path":"src/cli/memory.ts","line":494,"severity":"important","title":"Parse advertised substrate option","rationale":"The help advertises `\"Usage: soma memory refresh [--substrate <id>] ...\"`, but `parseMemoryRefreshArgs` only handles `case \"--substrate-home\":`; passing `--substrate` falls into `throw new Error(\`Unknown option: ${arg}\`)`. This breaks the documented way to refresh non-default substrates.","suggestion":"Add `case \"--substrate\": options.substrate = parseSubstrate(readOption(args, index, arg)); index += 1; break;`."}]}

### Architecture

- **[important]** `src/cli/memory.ts:153` — **Public command uses banned lifecycle verb**
  CONTEXT.md’s lifecycle section says `refresh` is a killed synonym for `reproject`, but the diff adds the public action `"refresh"` in `MEMORY_ACTIONS` and documents `Usage: soma memory refresh`. Shipping this CLI/API surface bakes a forbidden lifecycle term into Soma’s user-facing contract.
  Fix: Rename the public surface to a reproject-shaped term, e.g. `soma memory reproject --substrate <id>`, and align exported names accordingly.

### Performance

- **[important]** `src/memory-index.ts:444` — **Idle refresh still scans every note**
  On every non-missing index, session start runs `const { notes } = await collectDurableNotes(somaHome);` and then `notes.map(async ({ path }) => { ... return (await stat(path)).mtimeMs; })`, so an idle refresh still walks/parses the durable corpus and stats every note before deciding it is up to date. This makes SessionStart background work O(note count) even for no-op sessions, which is a concrete latency and I/O cost as memory grows.
  Fix: Use a directory-level freshness marker or update `memory write` to mark/rebuild the index, so SessionStart can avoid per-note scans when no write occurred.

### HonestOracle

- **[important]** `src/cli/memory.ts:204` — **Advertised substrate flag is not parsed**
  The help claims `"Usage: soma memory refresh [--substrate <id>]"`, and the PR says `"soma memory refresh [--substrate <id>]"`, but `parseMemoryRefreshArgs` only handles shared options and `--substrate-home`; `--substrate` falls to `Unknown option`. Anyone relying on the documented CLI path cannot invoke the substrate-neutral refresh for anything except the hardcoded default.


---

jcfischer (commented):

### Performance

- **[important]** `src/memory-projection-reproject.ts:83` — **Idle sessions still rewrite MEMORY.md**
  Even when `reindexMemoryIfStale` returns up-to-date, the projection path always reaches `await writeFile(target, `${memoryFile.content.trimEnd()}\n`, "utf8");`. That turns every SessionStart with a non-empty index into disk write churn despite the stale gate avoiding rebuild work.
  Fix: Read the existing target and skip `writeFile` when the normalized content is unchanged.

### Maintainability

- **[suggestion]** `src/memory-write.ts:464` — **Move displaced corpus-scan JSDoc**
  The new helper is inserted after the existing JSDoc ending with "never through the link to an outside target.", so that comment now sits immediately before `export async function listDurableNotePaths` instead of `collectDurableNotes`. This leaves `collectDurableNotes` with only its inline body comment and makes the symlink-guard documentation ambiguous.
  Fix: Place `listDurableNotePaths` before that existing JSDoc, or move the old JSDoc back directly above `collectDurableNotes`.

### HonestOracle

- **[important]** `PR description:11` — **Advertises nonexistent refresh entry points**
  The description claims "`refreshSubstrateMemoryProjection` (`src/memory-projection-refresh.ts`)" and "`soma memory refresh [--substrate <id>]`", but the diff adds `reprojectSubstrateMemoryProjection`, `src/memory-projection-reproject.ts`, and `soma memory reproject`. A downstream reader following the PR contract will look for APIs, files, and CLI commands that do not exist.

- **[important]** `PR description:16` — **Documents wrong failure event**
  The description claims failures emit `lifecycle.session_start.memory-refresh-failed`, but the diff emits `lifecycle.session_start.memory-reproject-failed`. Anyone wiring alerts or log queries from the PR text will monitor the wrong event name.


---

jcfischer (commented):

### Performance

- **[suggestion]** `src/memory-projection-reproject.ts:64` — **Avoid full projection for one file**
  The diff does `const profileInput = await loadSomaHome(somaHome);` and then `const projection = buildSubstrateHomeProjection(options.substrate, input, ...)` even though the function writes only the memory file. On every SessionStart reproject this loads the full Soma home and allocates the full substrate bundle to find one file by `file.content === indexContent`.
  Fix: Add an adapter-level memory projection helper or metadata path so this can render only the memory file.

### HonestOracle

- **[important]** `src/memory-write.ts:457` — **Skipped directories can hide stale memory**
  The comment claims "a genuine readdir failure is skipped — it can't contribute a staleness signal," but a directory that cannot be read may contain notes newer than memory/INDEX.md. That turns an access failure into a false up-to-date result, misleading callers that SessionStart kept memory current.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/memory-index.ts:471` — **Deleted notes do not trigger reindex**
  The freshness gate only rebuilds when `newestNoteMtimeMs > indexMtimeMs`, then otherwise returns `"up-to-date"`. If a durable note already present in `memory/INDEX.md` is deleted after the index was built, no remaining note necessarily has a newer mtime, so the projected index can keep advertising a note that no longer exists.
  Fix: Track a cheap corpus signature/count in the index metadata, or rebuild when the indexed note set cannot be proven to match the current durable path set.

### Architecture

- **[important]** `src/lifecycle.ts:367` — **Session start performs reproject**
  CONTEXT.md defines `reproject` as triggered by Soma on source change or Principal CLI, while `load` is the substrate action at session start; the diff wires `session start — rebuild memory/INDEX.md only if stale, then re-project` inside `runSomaLifecycleSessionStart`. This makes a load-time lifecycle event mutate projection state, violating the documented lifecycle boundary.
  Fix: Trigger this from a Soma-owned source-change/CLI reproject path, or define a separate Soma-owned lifecycle event for memory reproject before session load.

### HonestOracle

- **[suggestion]** `src/memory-projection-reproject.ts:31` — **Content match overclaims adapter correctness**
  The comment claims matching by content “stays correct without this module needing to know every adapter's layout,” but the implementation selects `projection.bundle.files.find((file) => file.content === indexContent)`. That proves only that a projected file has identical bytes, not that it is the substrate memory file, so the comment sells a heuristic as correctness.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/cli/memory.ts:494` — **Reproject parser ignores substrate option**
  The new parser only handles shared options and `case "--substrate-home":`, so `--substrate <id>` from the help text is still rejected as `Unknown option`. This makes the advertised command path unusable for non-default substrates.
  Fix: Add a `--substrate` case that sets `options.substrate = parseSubstrate(readOption(args, index, arg))` and advances the index.

### Maintainability

- **[suggestion]** `test/memory-projection-reproject.test.ts:19` — **Deduplicate memory test fixtures**
  This file introduces local fixtures with `function note(overrides: Partial<SomaMemoryNote> & { id: string; body: string }): SomaMemoryNote` and `async function seed(somaHome: string, n: SomaMemoryNote): Promise<void>`, while the adjacent memory index tests already exercise the same note seeding shape. Keeping corpus fixture builders per file makes future schema changes touch multiple near-identical helpers.
  Fix: Move note/seed helpers into a shared test fixture module.

### HonestOracle

- **[important]** `src/lifecycle.ts:735` — **Home-dir scrub is overclaimed**
  The comment claims `"scrub homeDir too"`, but the implementation uses `const homeDir = dirname(somaHome)` rather than the lifecycle `options.homeDir` or the actual substrate home. This misleads downstream readers into believing soft-fail errors cannot leak substrate-home paths when a custom `somaHome` is not under the real home directory.


---

jcfischer (commented):

### Performance

- **[suggestion]** `src/memory-index.ts:455` — **Bound concurrent note stats**
  The diff uses `const noteMtimes = await Promise.all( notePaths.map(async (path) => { ... return (await stat(path)).mtimeMs; ... }) );`, so an idle SessionStart with thousands of notes issues thousands of concurrent `stat` calls. That is a concrete filesystem I/O spike and can hit descriptor or scheduler limits despite avoiding `readFile`/parse.
  Fix: Process note stats with a small concurrency limit or sequential async iterator.

### HonestOracle

- **[important]** `src/memory-index.ts:427` — **Deletion detection is overstated**
  The comment claims the gate catches "a note deleted (a durable-dir mtime newer than the index — a removed note bumps its dir but leaves no file to stat)." That only holds when the durable directory mtime can be read; `listDurableNotePaths` later suppresses `stat(dir)` failures with `catch(() => undefined)`, so the claim presents deletion coverage as guaranteed when the implementation can silently lose that signal.


---

jcfischer (commented):

1 finding(s): 1 suggestion.


### HonestOracle

- **[suggestion]** `PR description:43` — **Verification claim lacks evidence**
  The description claims "`tsc --noEmit`: clean" and "`bun test`: **1887 pass, 0 fail**", but the PR data includes no command output, CI link, or reproducible evidence for those results. A downstream reviewer relying on this cannot distinguish an actually verified run from an asserted one.


---

jcfischer:
Verification on `c666100` (local):
```
$ bun run typecheck   # tsc --noEmit
exit 0

$ bun test
1889 pass, 0 fail — Ran 1889 tests across 120 files
```
Deferred as follow-ups (sage suggestions, not blocking): a dedicated adapter-level memory-file accessor (vs content-equality selection), and sharing note/seed test fixtures across the memory tests.

PR #445, opened 2026-07-08
feat(claude-code): project a Soma status line (script + settings.json)

jcfischer:
## What

Soma's claude-code adapter now **projects a status line** — a compact one-line replacement for the old 58 KB PAI/Cortex statusline (greeting, quote, weather, sparklines, full-width rules → gone).

Rendered line:
```
⚙E3 slug·phase · opus4.8 · soma ⎇main● · ctx 38% · 5h 12%⟳2h14 · 7d 41%⟳3d1h
```
- **`⚙E{effort} slug·phase`** — active VSA/run for this session (`○ task` when native/complete; empty when none), read from existing `current-work`/`active-algorithm-run` state.
- **model · dir ⎇branch●↑↓** · **ctx %** · **5h / 7d usage windows with reset countdowns** — all from Claude Code's stdin (`rate_limits.*.resets_at` is a Unix epoch; the script handles that), no network.

## How

Modeled exactly on the existing Soma hook projection (mode-classifier / policy-guard / precompact):
- Bundled asset `src/adapters/claude-code/statusline.sh` — the status-line script (field-tested live this session, then hardened through review: no `eval` in the parse, `-uno` git, ASCII-unit-separator field reads, numeric guards), with a `__SOMA_HOME__` placeholder that is shell-escaped and substituted at projection time (so custom soma-homes work; `SOMA_HOME` env still overrides).
- On install/reproject (default-on; opt out with `statusLine: false`): write `hooks/soma/soma-statusline.sh` (0755) and set `settings.json` `statusLine.command` to its absolute path.
- Uninstall removes the script and removes the `statusLine` entry **only when its `command` still points at our script** (a user's unrelated statusLine is never clobbered).
- Registered in `optionalHomeFiles` + the uninstall removal loop; `hooks/soma` is already an owned reconcile subtree.

## Tests

`test/claude-code-statusline.test.ts` (10) — default-on/opt-out plan, executable script with placeholder substituted, settings command set, disable, idempotent second install (byte-stable), custom `--soma-home`, uninstall removes script + settings, uninstall does NOT touch an unrelated statusLine, uninstall idempotency, unrelated settings keys preserved. Plus the AC-2 exact-file-list assertion updated.

- `tsc --noEmit`: clean.
- `bun test`: **1901 pass, 0 fail**.

## Substrate scope

claude-code only. Codex CLI has no status-line surface; pi-dev already renders its own Soma-aware phase footer via widgets. Both out of scope by design.

### DIFF
--- a/src/adapters/claude-code/hooks.ts
+++ b/src/adapters/claude-code/hooks.ts
@@ -14,7 +14,7 @@ import { isClaudeCodeInstallOptions } from "./install-options";
  */
 export function claudeCodeHookEnabled(
   options: unknown,
-  hook: "modeClassifier" | "policyGuard" | "preCompact",
+  hook: "modeClassifier" | "policyGuard" | "preCompact" | "statusLine",
 ): boolean {
   if (!isClaudeCodeInstallOptions(options)) return true;
   return options[hook] !== false;
@@ -28,6 +28,9 @@ export const SOMA_CLAUDE_POLICY_GUARD_RELATIVE_PATH = "hooks/soma/soma-policy-gu
 export const SOMA_CLAUDE_POLICY_GUARD_CONFIG_RELATIVE_PATH = "hooks/soma/soma-policy-guard.config.json";
 export const SOMA_CLAUDE_PRECOMPACT_RELATIVE_PATH = "hooks/soma/soma-precompact.mjs";
 export const SOMA_CLAUDE_PRECOMPACT_CONFIG_RELATIVE_PATH = "hooks/soma/soma-precompact.config.json";
+// soma statusline: self-contained bundled script (SOMA_HOME baked in at
+// projection time) — no bunPath, no argv dispatch, no companion config.json.
+export const SOMA_CLAUDE_STATUSLINE_RELATIVE_PATH = "hooks/soma/soma-statusline.sh";
 // PreToolUse matcher for the fail-closed enforcement guard: every tool whose
 // input can carry a dangerous command, an outbound exfiltration, or a
 // credential-path read/write that the runtime policy must inspect.
@@ -591,6 +594,30 @@ export async function unpatchClaudeCodePreCompactSettings(substrateHome: string,
   return writeJsonIfChanged(settingsPath, settings, before);
 }
 
+// The status line is a top-level `statusLine` key, not a hooks[] entry — it
+// has no matcher, no argv action, and (unlike every other soma-owned hook) no
+// bunPath, since Claude Code execs the bundled script directly via its
+// shebang. Soma always writes its own entry on install (last writer wins, as
+// with every other soma-owned settings key); uninstall only removes it when
+// the recorded `command` still points at OUR projected script path, so a
+// user's unrelated statusLine is never clobbered.
+export async function patchClaudeCodeStatusLineSettings(substrateHome: string, scriptPath: string): Promise<string[]> {
+  const settingsPath = resolve(substrateHome, SOMA_CLAUDE_SETTINGS_RELATIVE_PATH);
+  const { before, settings } = await readSettingsWithRaw(settingsPath);
+  settings.statusLine = { type: "command", command: scriptPath };
+  return writeJsonIfChanged(settingsPath, settings, before);
+}
+
+export async function unpatchClaudeCodeStatusLineSettings(substrateHome: string, scriptPath: string): Promise<string[]> {
+  const settingsPath = resolve(substrateHome, SOMA_CLAUDE_SETTINGS_RELATIVE_PATH);
+  const { before, settings } = await readSettingsWithRaw(settingsPath);
+  if (before.trim().length === 0) return [];
+  const statusLine = settings.statusLine;
+  if (!isObject(statusLine) || statusLine.command !== scriptPath) return [];
+  delete settings.statusLine;
+  return writeJsonIfChanged(settingsPath, settings, before);
+}
+
 export async function unpatchClaudeCodeSomaHookSettings(substrateHome: string, bunPath = resolveBunExecutable()): Promise<string[]> {
   const settingsPath = resolve(substrateHome, SOMA_CLAUDE_SETTINGS_RELATIVE_PATH);
   const { before, settings } = await readSettingsWithRaw(settingsPath);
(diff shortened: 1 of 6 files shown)


jcfischer (commented):

5 finding(s): 3 important, 1 suggestion, 1 nit.


### CodeQuality

- **[important]** `src/adapters/claude-code/hooks.ts:866` — **Shell path is not escaped**
  The diff embeds the raw path with `const rendered = source.replace("__SOMA_HOME__", () => somaHome);`, which makes the generated `SOMA_HOME="${SOMA_HOME:-...}"` wrong for paths containing `$`, backticks, quotes, or backslashes. That breaks the documented custom `--soma-home` behavior for valid local paths.
  Fix: Shell-escape the substituted default, or escape characters significant inside double quotes before replacement.

### Performance

- **[important]** `src/adapters/claude-code/statusline.sh:84` — **Avoid full git status per refresh**
  The status line runs on the interactive refresh path, and `sb=$(git -C "$cwd" --no-optional-locks status -sb --porcelain 2>/dev/null)` scans the worktree every time to derive branch, dirty, ahead, and behind. In large repos or network filesystems this can add visible prompt/status latency.
  Fix: Use cheaper commands for branch/ahead/behind and make dirty detection optional, cached, or time-bounded.

### Maintainability

- **[nit]** `src/adapters/claude-code/statusline.sh:15` — **Remove always-true branch**
  The condition `if [ -t 1 ] || true; then` is always true, so the surrounding branch is dead structure rather than a real terminal check. Keeping it makes later color behavior harder to reason about because the apparent guard does not guard anything.
  Fix: Remove the `if` wrapper or replace it with the intended terminal/color-mode condition.

- **[suggestion]** `src/adapters/claude-code/statusline.sh:104` — **Deduplicate rate window rendering**
  The 5h and 7d blocks repeat the same shape: `rr=$(until_str "$r5r")` plus `out+="$(sep)${DIM}5h ...` and then the same again for `r7`. A small helper would keep label, percent, and reset formatting in one place when the status-line format changes.
  Fix: Extract `append_window "5h" "$r5" "$r5r"` and reuse it for `7d`.

### HonestOracle

- **[important]** `src/adapters/claude-code/statusline.sh:61` — **Session-scoped status is overstated**
  The script claims "active VSA/run for THIS session", but the effort value is read from the global "$STATE_DIR/active-algorithm-run.json" rather than session-scoped current-work state. Downstream users can be shown another session's effort while the status line presents it as belonging to the current Claude Code session.


---

jcfischer (commented):

4 finding(s): 1 blocker, 1 important, 2 suggestion.


### CodeQuality

- **[important]** `src/adapters/claude-code/statusline.sh:81` — **Task fields break on spaces**
  `read -r phase task status < <(jq -r '[.phase // "", .task // .slug // "", .status // ""] | @tsv' "$cw" 2>/dev/null | tr '\t' ' ')` converts TSV separators into spaces, then `read` splits a task like `write adapter` into multiple fields. That truncates the displayed task and can prevent `status` from receiving `complete`.
  Fix: Read tab-separated fields with `IFS=$'\t' read -r phase task status` and remove the `tr`.

- **[suggestion]** `src/adapters/claude-code/statusline.sh:97` — **Branch names can be truncated**
  `branch=$(printf '%s' "$head" | sed -E 's/^## ([^ .]+).*/\1/')` stops at dots and spaces, so a valid branch like `release/v1.2` displays as `release/v1`. This makes the status line report the wrong branch name.
  Fix: Parse the porcelain header up to `...`, ` [`, or line end instead of excluding dots.

### Security

- **[blocker]** `src/adapters/claude-code/statusline.sh:21` — **Shell eval permits stdin injection**
  The diff runs `eval "$(printf '%s' "$input" | jq -r '` over Claude Code stdin, but rate-limit percentages are emitted unquoted via `"r5=" + (.rate_limits.five_hour.used_percentage // "" | tostring)` and `"r7=" + (.rate_limits.seven_day.used_percentage // "" | tostring)`. A crafted JSON string such as `$(touch /tmp/pwn)` in either field becomes shell code during eval.
  Fix: Shell-quote r5/r7 with `@sh`, or avoid `eval` and read parsed fields through a safer delimiter format.

### HonestOracle

- **[suggestion]** `src/adapters/claude-code/statusline.sh:4` — **Optional segment claim is false**
  The comment says, "Every segment is optional: a missing field simply drops its segment; the line never errors." But missing `cwd` is coerced to `.` and missing `context_window.used_percentage` is coerced to `0`, so the script still emits dir and `ctx 0%` rather than dropping those segments.


---

jcfischer (commented):

### Performance

- **[suggestion]** `src/adapters/claude-code/statusline.sh:116` — **Collapse extra git parsing forks**
  After `sb=$(git -C "$cwd" --no-optional-locks status -sb -uno --porcelain 2>/dev/null)`, the script still forks `head`, `tail`, and three `grep` pipelines on every statusline render via `head=$(printf '%s\n' "$sb" | head -1)`, `dirty=$(printf '%s\n' "$sb" | tail -n +2 | grep -c .)`, `ahead=$(...)`, and `behind=$(...)`. This is a repeated prompt-refresh path, so those extra processes add avoidable latency once the git status output is already in memory.
  Fix: Parse `sb` with bash parameter expansion and `[[ $head =~ ... ]]` regex matches.

### HonestOracle

- **[important]** `PR description:15` — **Field-tested parity is unsupported**
  The description claims `byte-identical to the field-tested script except a __SOMA_HOME__ placeholder substituted at projection time`, but the diff provides no source, checksum, fixture, or citation for the alleged field-tested script. Downstream reviewers cannot verify that this asset is actually the same artifact being invoked as prior evidence.

- **[suggestion]** `src/adapters/claude-code/statusline.sh:6` — **Never-errors claim is too broad**
  The script comments claim `The line never errors`, but the implementation depends on external commands such as `jq`, `git`, `date`, `sed`, `grep`, and `basename` and can still emit command errors or malformed arithmetic diagnostics for unexpected inputs. The diff only shows best-effort fallbacks for some missing data, not a no-error guarantee.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### Maintainability

- **[suggestion]** `src/adapters/claude-code/hooks.ts:672` — **Drop unused context field**
  `context: { somaHome: string; somaRepoPath: string; substrateHome: string }` includes `somaRepoPath`, but `installClaudeCodeStatusLine` only reads `context.somaHome` and `context.substrateHome`. Keeping unused fields in narrow helper contracts makes future readers check irrelevant state.
  Fix: Type the helper context as `{ somaHome: string; substrateHome: string }`.

### HonestOracle

- **[suggestion]** `PR description:28` — **Out-of-scope claim lacks support**
  The description states, "Codex CLI has no status-line surface; pi-dev already renders its own Soma-aware phase footer via widgets," but the PR data includes no Codex or pi-dev evidence. This turns a scope decision into an asserted substrate fact downstream readers cannot verify from the diff.


---

PR #446, opened 2026-07-09
feat(claude-code): session-scoped mode+effort feeder for the statusline

jcfischer:
## What

Follow-up to #445: bring back an **accurate, session-scoped** mode+effort indicator (`⚙E3`) in the claude-code statusline.

The Soma segment now renders `⚙E3 task` (algorithm), `○ task` (native), or nothing (minimal/absent) — driven by the session's actual mode classification.

## Why not the obvious sources

- The per-session `current-work` pointer only carries `native`/`complete` phase — no mode, no effort.
- The global `active-algorithm-run.json` has no session id and leaks across concurrent sessions (that's why #445 dropped the effort digit).
- Parsing the transcript for the `Soma MODE:` header is unreliable — that string also appears in ordinary conversation and tool output, so it matches false positives (verified failing against a real transcript).

## How — a mode feeder (entirely in the claude-code adapter)

The mode-classifier hook already asks `soma algorithm classify` for `{mode, effort}` on every prompt. This persists that authoritative answer per session for the statusline to read. The portable `algorithm classify` command stays **pure** — the whole feed lives in the adapter:

- **`src/adapters/claude-code/mode-classifier-hook.mjs`**: the sole writer — after classifying each prompt it writes `<somaHome>/memory/STATE/statusline-mode-<sid>.json` = `{mode, effort, updatedAt}` inline, best-effort (try/catch never breaks the hook). It sanitizes the session-id filename segment (`[^A-Za-z0-9._-]→-`) so a crafted id can't traverse out of `memory/STATE/`. (No separate TS helper — a standalone `.mjs` can't import TS, and the read side is bash; the filename convention is shared cross-language with `statusline.sh` and kept in lockstep.)
- **`src/adapters/claude-code/statusline.sh`**: the Soma segment reads mode/effort from the state file and the task from `current-work`; renders `⚙<effort> <task>` / `○ <task>` / nothing. Session-scoped only — no global reads. It guards `$sid` (`^[A-Za-z0-9._-]+$`) before any path use, so an unsafe id drops the segment on the read side too.

(Separator note: field reads route through `jq --arg sep "$US"` rather than embedding a raw unit-separator byte.)

## Tests

- `test/claude-code-install.test.ts`: hook-execution tests — the mode-classifier hook writes `statusline-mode-<sid>.json` after classifying; a missing `session_id` writes no file; a traversal `session_id` sanitizes to a single basename inside `STATE`.
- `test/claude-code-statusline.test.ts`: render tests — algorithm (`⚙E3` + full multi-word task), native (`○ task`), minimal/missing (segment dropped, rest of line intact), and an unsafe `$sid` (`../…`) dropping the Soma segment.
- `classify` is unchanged (byte-identical to its pre-feature base) — this PR adds no CLI surface.

- `tsc --noEmit`: clean.
- `bun test`: **1917 pass, 0 fail**.
- Verified end-to-end: a multi-file-refactor prompt classifies as `algorithm E3` → state file → statusline renders `⚙E3 <task>`. Traversal covered: a `../evil` session id sanitizes to a single basename on write and drops the Soma segment on read. The main-parse numeric-field injection and empty-field-shift guards from #445 are unchanged and reconfirmed.


jcfischer (commented):

### Architecture

- **[important]** `src/cli/algorithm.ts:80` — **Claude statusline leaks into Algorithm CLI**
  docs/architecture.md says the Algorithm `classify` command is part of the portable command surface, while adapters own substrate-native primitives; the diff adds `"(--session-id also best-effort writes the statusline mode+effort state file for that session)"` and calls `writeStatuslineModeState(...)` from `algorithm classify`. That makes a substrate UI feed a side effect of the core Algorithm classifier instead of a Claude Code adapter concern.
  Fix: Keep `classify` pure and have the Claude Code hook write the session statusline state after classification, or move this behind a Claude Code adapter command.

- **[important]** `src/index.ts:509` — **Statusline state becomes public API**
  The root export adds `statuslineModeStatePath` and `writeStatuslineModeState`, while the new module describes itself as `"Session-scoped mode+effort feeder for the claude-code statusline"`. Exporting a Claude Code statusline helper from `src/index.ts` makes a substrate-specific implementation detail part of the public Soma surface.
  Fix: Do not export this from the root API; keep it under the Claude Code adapter or expose a substrate-neutral session classification state abstraction.

### Performance

- **[suggestion]** `src/cli/algorithm.ts:704` — **Hook now waits on state write**
  The Claude Code hook runs classification via `spawnSync(...)`, and the classify path now does `await writeStatuslineModeState({` before returning output. That adds a filesystem mkdir/write to the prompt hook's synchronous latency on every classified prompt.
  Fix: Return classification first or make the feeder write fire-and-forget from the hook/CLI path.

### HonestOracle

- **[important]** `PR description:36` — **Unsupported injection-proof claim**
  The description claims "Injection-proof and empty-field-safe reconfirmed on the changed parse," but the diff explicitly builds state paths from a "RAW session id" and "`$sid`, unsanitized" in `src/statusline-mode-state.ts`. The shown tests cover shell numeric-field injection, not malicious or malformed `session_id` values crossing the writer and `statusline.sh` read path, so downstream readers are being told a broader safety property than the PR demonstrates.


---

jcfischer (commented):

### Maintainability

- **[suggestion]** `src/adapters/claude-code/mode-classifier-hook.mjs:56` — **De-duplicate statusline state writer**
  The hook repeats the same writer logic as the TS helper: `const token = sessionId.replace(/[^A-Za-z0-9._-]+/g, "-");` mirrors `return sessionId.replace(/[^A-Za-z0-9._-]+/g, "-");`, and both build `statusline-mode-${...}.json` payloads. Future edits to sanitization or payload fields now have to be made in two production paths.
  Fix: Move the shared writer/path logic to a small importable `.mjs` helper and have the TS tests exercise that implementation.

### HonestOracle

- **[important]** `PR description:31` — **CLI test claim is unsupported**
  The description claims "classify --session-id writes the file (both `--json` and text)", but the diff only adds hook coverage via `runClaudeModeClassifierHook` and direct helper tests. That misleads downstream reviewers into believing the portable CLI path is tested in this PR when no `src/cli/algorithm.ts` change or CLI test is present.


---

jcfischer (commented):

### Performance

- **[suggestion]** `src/adapters/claude-code/statusline.sh:93` — **Avoid needless jq on missing state**
  The statusline now runs `jq -j --arg sep "$US" '[(.mode // ""), (.effort // "")] | join($sep)' "$STATE_DIR/statusline-mode-$sid.json"` on every safe session id, even before confirming that the per-session state file exists. On sessions before the first classifier write or after cleanup, that is a guaranteed failed subprocess on the statusline hot path.
  Fix: Guard with `[ -r "$STATE_DIR/statusline-mode-$sid.json" ]` before invoking jq.

### HonestOracle

- **[important]** `PR description:16` — **Describes a nonexistent state helper**
  The description claims "src/adapters/claude-code/statusline-mode-state.ts (new, NOT root-exported)" exists, but the changed files contain no such file and the helper is implemented inline in mode-classifier-hook.mjs. A downstream reviewer relying on this will look for a contract boundary that the PR does not actually create.

- **[suggestion]** `src/adapters/claude-code/statusline.sh:88` — **Comment invents classify session flag**
  The comment says the feed is "written by `soma algorithm classify --session-id` on every prompt", but the diff only calls classification from the hook and then writes state in adapter code. This misattributes the source of the state file to a CLI behavior not shown in the PR.


---

jcfischer (commented):

1 finding(s): 1 suggestion.


### Performance

- **[suggestion]** `src/adapters/claude-code/statusline.sh:102` — **Extra jq per statusline render**
  The new `IFS="$US" read -r smode seff < <(jq -j --arg sep "$US" '[(.mode // ""), (.effort // "")] | join($sep)' "$modefile" 2>/dev/null)` adds another process spawn in the statusline hot path whenever the mode file exists. The script already runs `jq` again for current-work, so active sessions now pay two JSON parser launches before rendering the Soma segment.
  Fix: After resolving `cw`, read mode state and task in one jq invocation when both files exist.


---

Issue #28, opened 2026-05-16
Automatic migration of existing PAI installation to Soma home


jcfischer:
## Problem

`src/pai-importer.ts` exists but covers only a thin slice — identity (principal + assistant) plus a hand-rolled telos rendering. A real PAI installation at `~/.claude/` contains far more state that a user expects to keep when moving to Soma:

```
~/.claude/
  CLAUDE.md                       # global instructions
  PAI/                            # USER + DOCUMENTATION + ACTIONS + FLOWS + ...
    USER/                         # PRINCIPAL_IDENTITY, DA_IDENTITY, TELOS/, PROJECTS/, ...
    Algorithm/v6.3.0.md           # already importable via `soma import algorithm`
  skills/                         # 70+ user skills (ISA, Browser, Tana, Calendar, ...)
  agents/                         # subagents
  commands/                       # slash commands
  hooks/                          # PAI hooks
  bin/                            # user scripts
  daemon/                         # IvyHeartbeat etc.
  embeddings/                     # cached embeddings
  memory/                         # auto-memory store (already populated for many projects)
  events/                         # event log
  history.jsonl                   # session history
```

Current `soma import pai` only writes:
- `profile/principal.md`
- `profile/assistant.md`
- `profile/telos.md`
- `profile/imports/claude/<source files>`

Everything else (skills, agents, commands, hooks, memory, telos sub-files like PROJECTS.md, BELIEFS, STRATEGIES) is left behind. A user who runs `soma import pai && rm -rf ~/.claude/PAI` loses most of their assistant.

## Vision

`soma migrate pai` is a **single, idempotent command** that produces a Soma home (`~/.soma/`) which is functionally equivalent to the user's existing PAI installation, with every PAI concept mapped to its Soma equivalent and every loose end either migrated or explicitly flagged.

Re-running it on an already-migrated home is a no-op (or a delta sync) — never destructive.

## Migration Map

| PAI source | Soma target | Notes |
| --- | --- | --- |
| `~/.claude/CLAUDE.md` | `~/.soma/profile/claude-md.md` (snapshot) + parse for `@`-imports | Parse imports, follow them, ingest each. Keep raw snapshot for audit. |
| `~/.claude/PAI/USER/PRINCIPAL_IDENTITY.md` | `~/.soma/profile/principal.md` (already handled) | Existing path. |
| `~/.claude/PAI/USER/DA_IDENTITY.md` | `~/.soma/profile/assistant.md` (already handled) | Existing path. |
[...]

PR #447, opened 2026-07-09
fix(migrate): reserve profile/purpose.md so migrate/import don't clobber curation (#441)

### LINKED ISSUE #441: migrate pai: identity importer clobbers curated purpose.md on every run (should be a reserved target)
## Problem

`soma migrate pai`'s **identity phase regenerates and overwrites the profile `purpose.md` on every run**, clobbering any hand-curation with a mechanical concatenation of the raw PAI TELOS source files.

After a curated `purpose.md` exists, re-running `soma migrate pai --apply` replaces it with a dump where:
- the `Mission:` line becomes the PAI **template placeholder string** (the unfilled boilerplate header from the source `MISSION.md`), not the real distilled mission;
- `## Goals / ## Principles / ## Commitments` become raw concatenations of the source `GOALS.md` / `BELIEFS.md` / `STRATEGIES.md`, including `---` separators, tables, and ascii diagrams.

This is not cosmetic: `src/soma-home.ts` (~L231) reads `Mission:` + the section bullets **straight from `purpose.md`** to build the projected Purpose, so a clobbered file propagates the placeholder mission + dump into **every substrate** on the next `reproject`.

## Repro

1. Curate the profile `purpose.md` down to the four portable fields (a lean Mission line + short Goals/Principles/Commitments bullets).
2. `soma migrate pai --apply` (e.g. to pull newer TELOS/memory content).
3. Observe `purpose.md` is now the raw dump; any startup-context read (`soma precompact capture`) shows the placeholder mission instead of the curated one.

Observed live: a curated ~1.4 KB `purpose.md` was silently overwritten with an ~11 KB dump.

## Expected

`purpose.md` should be a **reserved** target: the identity importer creates it if absent, but does **not** overwrite an existing curated file unless explicitly forced. The `--overwrite-reserved` flag implies reserved-target handling already exists, but it does not currently protect `purpose.md`. (`assistant.md` / `principal.md` are safe because they are deterministic distillations with nothing to lose; `purpose.md` is the one that diverges once curated.)

## Proposed fix

- Add the profile `purpose.md` to the identity importer's reserved set so re-migration skips it by default (log a `skipped reserved` line), and only rewrites it under `--overwrite-reserved`.
- Alternatively, detect divergence (on-disk file differs from the freshly-generated one) and skip-with-warning rather than silently overwrite.

## Workaround (today)

If clobbered, reconstruct `purpose.md` from the still-clean generated `PURPOSE.md` projection that any substrate install produces, in the parser's format: `Mission: <one line>` then `## Goals` / `## Principles` / `## Commitments` with `- ` bullets. This is Step 1 of the new `migrate-pai-telos` skill (#440).

## Related

- #440 (migrate-pai-telos skill — its Step 1 is the manual guard/restore for this bug)

jcfischer:
Fixes #441.

## Bug

`soma migrate pai` (and `soma import pai`) regenerated and overwrote `profile/purpose.md` on **every** run, clobbering hand-curation with a mechanical dump of the raw PAI TELOS sources. Since `src/soma-home.ts` reads `purpose.md` to build every substrate's projected Purpose, a clobber propagated a placeholder mission into every substrate on the next reproject. (`principal.md`/`assistant.md` are safe — deterministic distillations.)

## Fix

Make `profile/purpose.md` a **reserved** identity target:
- `src/pai-importer.ts`: `RESERVED_IDENTITY_TARGETS = { profile/purpose.md }`. The write loop skips a reserved target when it already `existsSync` on disk and `overwriteReserved !== true`, recording it in a new `skippedReserved` result field instead of overwriting. Content is still computed; only the write is gated. First run (absent) still creates it; `--overwrite-reserved` still replaces it.
- `src/types.ts`: `PaiImportOptions.overwriteReserved?` + `PaiImportResult.skippedReserved?`.
- Plumbing: `migrate pai` threads its existing `--overwrite-reserved` flag into the identity phase; `soma import pai` gains a matching `--overwrite-reserved` flag. Both now protect `purpose.md`.
- Reporting: both CLIs emit `skipped reserved (exists): profile/purpose.md — re-run with --overwrite-reserved to replace` when applicable (ephemeral CLI output only). The skip protects any existing file (it doesn't attempt to *detect* hand-curation), and a non-regular-file at the target throws rather than being silently treated as reserved.

## Manifest idempotency (subtle)

The persisted `MIGRATION.md` manifest has a byte-stability invariant across idempotent reruns. A naive skip would move `purpose.md` from `files` to `skippedReserved` between run 1 and run 2, changing the manifest's identity count/fingerprint. Resolved the same way the codebase already handles the memory phase: the manifest counts/fingerprints the **union** of written + skipped (a stable end-state fact); the written/skipped split lives only in ephemeral CLI output.

## Tests

8 new (TDD, RED→GREEN): rerun skips an existing `purpose.md` (byte-unchanged, in `skippedReserved` not `files`; principal/assistant still rewritten); `overwriteReserved: true` replaces it; first run creates it; a directory at the target throws; CLI reporting; migrate-level full-manifest byte-stability across a genuinely idempotent rerun.

- `tsc --noEmit`: clean. `bun test`: **1918 pass, 0 fail**.

### DIFF
--- a/src/cli/import.ts
+++ b/src/cli/import.ts
@@ -1,4 +1,3 @@
-import { relative } from "node:path";
 import {
   importAlgorithm,
   importPaiDocs,
@@ -9,6 +8,7 @@ import {
   planPaiImport,
   planPaiPackImport,
 } from "../index";
+import { formatReservedSkipLine } from "../pai-importer";
 import type {
   AlgorithmImportOptions,
   AlgorithmImportPlan,
@@ -232,15 +232,11 @@ function formatSimpleImportResult(title: string, dirEntries: string[], files: st
 // soma#441 — per-run reserved-skip detail lives only in this ephemeral
 // CLI summary, not in any on-disk/byte-compared surface (mirrors the
 // memory phase's "written N / unchanged M" precedent in
-// `formatPaiMigrationResult`).
+// `formatPaiMigrationResult`). Line shape is single-sourced via
+// `formatReservedSkipLine` so this and the migrate summary can't drift.
 function formatSkippedReservedLines(somaHome: string, skippedReserved: string[] | undefined): string[] {
   if (!skippedReserved || skippedReserved.length === 0) return [];
-  return [
-    "",
-    ...skippedReserved.map(
-      (path) => `skipped reserved (curated): ${relative(somaHome, path)} — re-run with --overwrite-reserved to replace`,
-    ),
-  ];
+  return ["", ...skippedReserved.map((path) => formatReservedSkipLine(somaHome, path))];
 }
 
 function formatPaiImportResult(result: PaiImportResult): string {
--- a/src/cli/import.ts
+++ b/src/cli/import.ts
@@ -1,3 +1,4 @@
+import { relative } from "node:path";
 import {
   importAlgorithm,
   importPaiDocs,
@@ -57,7 +58,7 @@ export type ParsedImportArgs =
 export const IMPORT_COMMAND_HELP: { usage: string; subcommands: Record<ImportSource, string> } = {
   usage: "Usage: soma import <pai|algorithm|pai-pack|pai-docs> ...",
   subcommands: {
-    pai: "Usage: soma import pai [--dry-run] [--apply] [--home-dir <dir>] [--claude-home <dir>] [--soma-home <dir>]",
+    pai: "Usage: soma import pai [--dry-run] [--apply] [--home-dir <dir>] [--claude-home <dir>] [--soma-home <dir>] [--overwrite-reserved]",
     algorithm: "Usage: soma import algorithm [--dry-run] [--apply] [--home-dir <dir>] [--pai-algorithm-dir <dir>] [--soma-home <dir>]",
     "pai-pack": "Usage: soma import pai-pack [--dry-run] [--apply] [--home-dir <dir>] --pai-pack-dir <dir> [--soma-home <dir>] [--skill-name <name>] [--overwrite] [--include-unrecognized]",
     "pai-docs": "Usage: soma import pai-docs [--dry-run] [--apply] [--home-dir <dir>] --pai-source-dir <dir> [--soma-home <dir>]",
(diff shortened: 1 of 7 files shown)


jcfischer (commented):

### Maintainability

- **[suggestion]** `src/cli/import.ts:240` — **Deduplicate reserved-skip CLI text**
  The same user-facing message is built in two places: `skipped reserved (curated): ${relative(somaHome, path)} — re-run with --overwrite-reserved to replace` and `identity skipped reserved (curated): ${relative(result.somaHome, path)} — re-run with --overwrite-reserved to replace`. Keeping the policy text duplicated makes future wording or flag-name changes easy to miss in one command.
  Fix: Extract a small shared formatter for reserved-skip lines, with the command-specific prefix as an argument.

### HonestOracle

- **[important]** `test/pai-migration-issue-441.test.ts:47` — **Byte-stability claim is not tested**
  The test comment claims the manifest body must not change "byte-for-byte", but the assertion only compares `identityLine(secondManifest)` to `identityLine(firstManifest)`. This lets timestamp, fingerprint, or other manifest drift ship while the PR description claims "migrate-level manifest byte-stability across reruns."

- **[suggestion]** `test/pai-migration-issue-441.test.ts:13` — **Unlocatable Sage citation**
  The comment cites "established invariant — see `renderManifest`'s Sage r1 #28 comment", but the supplied diff does not contain a `Sage r1 #28` comment in `renderManifest`. That reference makes the invariant look externally grounded when the cited source is not locatable here.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/pai-importer.ts:345` — **Validate reserved target is a file**
  The new guard `if (RESERVED_IDENTITY_TARGETS.has(relativePath) && existsSync(target) && options.overwriteReserved !== true) {` treats a directory or other non-file at `profile/purpose.md` as curated, then `skippedReserved.push(target); continue;` reports success. That lets `soma import pai` silently leave an unusable purpose target instead of surfacing the bad filesystem state.
  Fix: Use `stat`/`lstat` and skip only regular files; throw for directories or unsupported node types.

### HonestOracle

- **[important]** `src/pai-importer.ts:13` — **CLI asserts unproven curation**
  The emitted text says "skipped reserved (curated)" even though the skip condition only checks that the target exists; the new idempotency test explicitly exercises a second run where the file was generated, not hand-curated. This misleads operators into believing Soma detected human curation when it did not.

- **[suggestion]** `PR description:26` — **Test count is inaccurate**
  The description claims "7 new (TDD, RED→GREEN)", but the diff adds four tests in `test/pai-importer.test.ts` and four in `test/pai-migration-issue-441.test.ts`. That overstates precision in the evidence summary.


---

jcfischer (commented):

1 finding(s): 1 suggestion.


### HonestOracle

- **[suggestion]** `PR description:7` — **Unshown substrate propagation claim**
  The description claims, "src/soma-home.ts reads purpose.md to build every substrate's projected Purpose," but this PR data does not include that file or any citation proving every substrate consumes it that way. A downstream reader is being asked to accept the blast-radius claim without evidence in the diff.


---

PR #448, opened 2026-07-09
fix(install): key portable-skill manifest per substrate home (#438)

### LINKED ISSUE #438: Portable-skill manifest clobbers across same-substrate multi-home installs
## Problem

The portable-skill install manifest is keyed by `(somaHome, substrate)`, not by substrate home:

```
<somaHome>/projections/<substrate>/install-manifest.json
```

So when two substrate homes of the **same** substrate are installed from **one** soma home — e.g. `soma install claude-code --substrate-home ~/.claude` and `... --substrate-home ~/claude-fresh` — the second install **overwrites** the first's manifest. Uninstalling the first home then reads a manifest whose `substrateHome` points to the second home; the diff-home guard discards the whole manifest, silently **orphaning the first home's projected skill dirs** (the-algorithm/, Memory/) on disk with no record.

This is a **pre-existing** defect (grok's manifest had the same keying before the shared generalization in #437). PR #437 extends it to claude-code. Surfaced by Sage's pi.dev review of #437.

## Impact

Low in practice today (the multi-home scenario is mainly disposable `--substrate-home` test homes), but it is a real silent orphaned-file leak that contradicts the manifest's stated round-trip-on-uninstall purpose.

## Fix

Key the manifest per substrate home. In `src/adapters/shared/portable-skill-manifest.ts`:

- Change `portableSkillManifestPath(somaHome, substrate)` → `(somaHome, substrate, substrateHome)`, incorporating a stable hash (or sanitized segment) of the resolved substrate home into the filename/subdir, e.g. `projections/<substrate>/<hash(substrateHome)>/install-manifest.json`.
- Update `write/read/reconcile/remove` (they already receive `substrateHome`).
- Update grok's `grokInstallManifestPath(somaHome)` wrapper → `(somaHome, substrateHome)` and the grok tests that call it.
- The existing in-manifest `substrateHome` diff-guard stays as defense-in-depth.

Acceptance: two same-substrate homes installed from one soma home each get an independent manifest; uninstalling one round-trips only its own skills and leaves the other's intact.

jcfischer:
Fixes #438.

## Bug

The portable-skill install manifest was keyed `(somaHome, substrate)` — `<somaHome>/projections/<substrate>/install-manifest.json` — ignoring `substrateHome`. Two homes of the SAME substrate installed from ONE soma home (`soma install claude-code --substrate-home ~/.claude` and `... --substrate-home ~/claude-fresh`) clobbered one manifest (last write wins). Uninstalling the first then read a manifest whose in-file `substrateHome` guard rejected it as foreign → the first home's projected skill dirs were silently orphaned.

## Fix

Key the manifest per resolved substrate home:
`<somaHome>/projections/<substrate>/<sha256(resolve(substrateHome)).slice(0,12)>/install-manifest.json`

- `src/adapters/shared/portable-skill-manifest.ts`: new `substrateHomeSegment(substrateHome)`; `portableSkillManifestPath` and `readPortableSkillManifest` take `substrateHome`; internal write/reconcile/remove thread it. The in-file `substrateHome` diff-guard stays as defense-in-depth (now genuinely a *second* line, not the only one).
- `src/adapters/grok/install-manifest.ts`: `grokInstallManifestPath`/`readGrokInstallManifest` take `substrateHome`. `home-projection.ts` / `claude-code/install.ts` / `grok/install.ts` already passed `substrateHome` in their options — no changes needed there.

## Tests

- NEW regression (#438): install claude-code to two `--substrate-home` dirs from one soma home → distinct manifest paths; uninstall home A leaves home B's manifest AND projected skill dirs intact.
- Rewrote 3 existing tests that (under the new scheme) would have passed via "manifest not found" instead of exercising the in-file mismatch / path-escape guards — they now write the manifest at the correct per-home path so the guard is genuinely tested.

## Upgrade note (not fully harmless)

All reads now use the per-home path, so the old flat manifest is no longer consumed. Concretely: for an install made **before** this change, an `uninstall` right after upgrading will **not** round-trip those skills — they stay projected-but-untracked — until a `soma install`/reproject regenerates the manifest at the new per-home path (which re-tracks them). So after upgrading, **reinstall/reproject before relying on uninstall** for a pre-fix install. Install/uninstall are verified not to crash on a leftover old-format manifest. No automatic migration is built (per #438's scope — the multi-home scenario is mostly disposable `--substrate-home` test homes).

- `tsc --noEmit`: clean. `bun test`: **1921 pass, 0 fail**.

### DIFF
--- a/src/adapters/grok/install-manifest.ts
+++ b/src/adapters/grok/install-manifest.ts
@@ -11,8 +11,10 @@ import {
 /**
  * Grok's portable-skill install manifest is the shared
  * `portable-skill-manifest` mechanism bound to the `grok` substrate: schema
- * `soma-grok-install-manifest-v1`, stored at `<somaHome>/projections/grok/`.
- * These thin wrappers preserve the grok-named API its adapter/doctor/tests
+ * `soma-grok-install-manifest-v1`, stored at `<somaHome>/projections/grok/
+ * <substrateHomeHash>/` (soma#438: keyed per substrate home so two `.grok`
+ * homes installed from one soma home don't share a manifest). These thin
+ * wrappers preserve the grok-named API its adapter/doctor/tests
  * already import; the byte output is identical to the shared helper with
  * `substrate: "grok"`. See `../shared/portable-skill-manifest.ts` for the
  * full contract (SHARED skills dir → manifest-tracked round-trip).
@@ -21,8 +23,8 @@ export const GROK_INSTALL_MANIFEST_SCHEMA = portableSkillManifestSchema("grok");
 
 export type GrokInstallManifest = PortableSkillManifest;
 
-export function grokInstallManifestPath(somaHome: string): string {
-  return portableSkillManifestPath(somaHome, "grok");
+export function grokInstallManifestPath(somaHome: string, substrateHome: string): string {
+  return portableSkillManifestPath(somaHome, "grok", substrateHome);
 }
 
 export function writeGrokInstallManifest(options: {
@@ -33,8 +35,8 @@ export function writeGrokInstallManifest(options: {
   return writePortableSkillManifest({ ...options, substrate: "grok" });
 }
 
-export function readGrokInstallManifest(somaHome: string): Promise<GrokInstallManifest | null> {
-  return readPortableSkillManifest(somaHome, "grok");
+export function readGrokInstallManifest(somaHome: string, substrateHome: string): Promise<GrokInstallManifest | null> {
+  return readPortableSkillManifest(somaHome, "grok", substrateHome);
 }
 
 export function reconcileGrokPortableSkillProjection(options: {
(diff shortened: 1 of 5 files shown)


jcfischer (commented):

### CodeQuality

- **[nit]** `test/portable-skill-manifest.test.ts:63` — **Comment no longer matches exercised path**
  The comment says "Same manifest bytes, read under a DIFFERENT substrate → schema mismatch → null", but the changed call `readPortableSkillManifest(somaHome, "grok", substrateHome)` now reads from a different per-substrate path, so it returns null via missing file rather than schema mismatch.
  Fix: Update the comment/test to describe missing per-substrate manifest, or write bytes at the grok path if schema mismatch is intended.

### HonestOracle

- **[important]** `PR description:31` — **Old manifest impact is understated**
  The description says "a pre-fix manifest at the old flat path becomes an inert stale orphan" and "No migration built (documented, harmless)", but the diff changes all reads to `portableSkillManifestPath(..., substrateHome)`, so uninstall after upgrade no longer consumes the old flat manifest. This misleads operators because existing projected skills can remain untracked until a reinstall regenerates the per-home manifest.


---

jcfischer (commented):

1 finding(s): 1 suggestion.


### HonestOracle

- **[suggestion]** `test/portable-skill-manifest.test.ts:195` — **Regression comment names wrong victim**
  The comment says home A uninstall reads home B's manifest and causes "orphaning or deleting home B's projected files," but the implementation checks `resolve(manifest.substrateHome) !== substrateHome` and returns without touching home B. This misstates the actual failure mode: home A's projected files are left behind because its manifest was clobbered.


---

PR #449, opened 2026-07-09
test(install): greenfield install acceptance across all adapters (#373)

### LINKED ISSUE #373: Greenfield install acceptance test in CI (all adapters)
## Problem

`soma install <substrate>` is only ever exercised against long-lived, pre-populated home directories. Fresh-machine installs regress silently (the 2026-06 Vincent onboarding fixes were this class of bug).

## Proposal

Greenfield install acceptance test in CI, per adapter (claude-code, codex, cursor, grok, pi-dev): install into an empty temp directory and assert the projected files exist, parse, and reference no missing paths.

## Acceptance

- CI job runs `soma install <s> --home-dir <tmp> --soma-home <tmp-soma>` for all five adapters on every PR.
- Assertions: expected file set exists; markdown/JSON parse; no dangling path references (grep for absolute paths that do not exist in the tmp tree).
- A deliberately broken adapter fixture fails the job.

jcfischer:
Fixes #373.

## Why

`soma install <substrate>` was only ever exercised against long-lived, pre-populated home dirs, so fresh-machine installs regressed silently (the 2026-06 Vincent onboarding fixes were this class of bug). CI runs `bun test` (`.github/workflows/portability.yml`), so a suite test guards every PR — no new CI job needed.

## What

`test/greenfield-install.test.ts` — for each of the 5 projection-lifecycle substrates (codex, pi-dev, claude-code, cursor, grok): install into TWO fresh empty temp roots (`homeDir`, `somaHome`) via the same `installSomaFor*` functions the CLI's `runInstall` dispatches to (no pre-seeded state), then `auditProjectedFiles` asserts:
1. every reported projected file exists and is non-empty;
2. every `*.md` is non-empty UTF-8, every `*.json` parses;
3. no projected file references an absolute path INTO the temp tree that doesn't resolve on disk.

`auditProjectedFiles`/`extractPathReferences`/`findDanglingPathReferences` are exported standalone and unit-tested. The dangling-path exemption is an **injected** predicate (default: never exempt), so it only narrows check (3) — checks (1)/(2) are unconditional.

## Findings (all adapters install clean)

All 5 install greenfield with no throw. The dangling-ref scan surfaced real, non-vacuous hits (codex 14, pi-dev 7, grok 3) that trace to **intentional** patterns, exempted with source citations in `isKnownLazyOrCrossSubstrateReference`:
- cross-substrate defensive privacy roots (e.g. codex hard-codes `.claude/memory*` regardless of whether claude-code is installed);
- lazily-created dirs populated by a later command (`profile/imports/**` via `migrate pai`; `memory/RAW/untrusted` + `memory/SECURITY/inbound-content`; `memory/LEARNING/ALGORITHM` on first Algorithm run).
The exemption skips cursor's degenerate `defaultHome = "."` case (which would otherwise neuter the scan).

## Regression proof

`auditProjectedFiles flags a deliberately broken projection` feeds a fixture with a dangling backtick path, malformed JSON, and an empty `.md`, and asserts each is caught with the right reason; a sibling clean-fixture test guards against false positives.

- `tsc --noEmit`: clean. `bun test`: **1928 pass, 0 fail**.

### DIFF
--- a/test/greenfield-install.test.ts
+++ b/test/greenfield-install.test.ts
@@ -1,19 +1,28 @@
 // soma#373 — greenfield install acceptance test.
 //
-// Every other install test exercises `installSomaFor*` against a home that
-// (within the same test) has already been bootstrapped, so a fresh-machine
+// Every other install test exercises the `installSomaFor*` functions against
+// a home that (within the same test) has already been bootstrapped, so a fresh-machine
 // regression — a file the plan promises but apply never writes, a projected
 // doc that references a path nothing ever creates, a hand-edited JSON
-// fixture that stops parsing — can slip through unnoticed. This file installs
-// into a completely empty `homeDir` + `somaHome` pair (nothing pre-created,
-// mirroring `soma install <substrate> --home-dir <tmp> --soma-home <tmp>
-// --apply` end to end: `installSomaFor*` IS what the CLI's `runInstall`
-// calls — see `src/cli/substrate-lifecycle.ts`'s `installers` map) for every
-// substrate on `PROJECTION_LIFECYCLE_SUBSTRATES`, then audits the result.
+// fixture that stops parsing — can slip through unnoticed. This file drives
+// the REAL CLI install path — `parseInstallArgs(["install", <substrate>,
+// "--home-dir", <tmp>, "--soma-home", <tmp>, "--apply"])` fed through
+// `runSubstrateLifecycleCli` (src/cli/substrate-lifecycle.ts) — so it
+// exercises argv parsing, dispatch, and the `--apply` gate, not just the
+// installer function. It runs for every substrate on
+// `PROJECTION_LIFECYCLE_SUBSTRATES` against empty existing temp roots, then
+// audits what actually landed on disk.
 //
-// Three acceptance checks per substrate, run by `auditProjectedFiles`:
-//   1. every file the install reports as projected (`result.somaHome.files`
-//      + `result.substrateHome.files`) exists on disk and is non-empty.
+// `runSubstrateLifecycleCli` returns a formatted string, not a structured
+// file list, so the audit's file SET is derived two ways (soma#373 sanctions
+// either): the static install PLAN's declared files (asserted present on
+// disk — the "declared promise kept" check), plus a full walk of the temp
+// roots (every projected file, including dynamic bundled-skill/lifecycle
+// output the static plan omits) which feeds the content checks.
+//
+// Three acceptance checks, run by `auditProjectedFiles`:
+//   1. every projected file that landed exists on disk and is non-empty; and
+//      every file the static plan declares is among them (declared→landed).
 //   2. every projected text file (`.md`, `.json`, and other non-binary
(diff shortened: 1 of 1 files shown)


jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `test/greenfield-install.test.ts:14` — **UTF-8 validation is overclaimed**
  The diff claims "every projected `*.md` file is non-empty UTF-8 text", but the audit only does `readFile(file, "utf8")`, which decodes invalid byte sequences rather than proving the file is valid UTF-8. The regression fixture also lacks malformed UTF-8, so downstream readers are misled into believing this suite enforces an encoding guarantee it does not enforce.


---

jcfischer (commented):

### Architecture

- **[suggestion]** `test/greenfield-install.test.ts:194` — **Avoid banned lazy terminology**
  CONTEXT.md bans `lazy` in favor of precise loading/lifecycle terms; the diff adds `function isKnownLazyOrCrossSubstrateReference` and `const lazyPrefixes = [` for paths populated by later commands. This creates a new code-level alias where the glossary explicitly says `lazy` is banned.
  Fix: Rename to `isKnownOnDemandOrCrossSubstrateReference` and `onDemandPrefixes`, or use a more specific `laterCreatedPrefixes`.

### HonestOracle

- **[important]** `test/greenfield-install.test.ts:8` — **Installer call sold as CLI end-to-end**
  The comment claims the test is "mirroring `soma install <substrate> --home-dir <tmp> --soma-home <tmp> --apply` end to end", but the diff calls `install({ homeDir, somaHome })` directly rather than invoking the CLI path. That misleads downstream readers into believing CLI parsing, dispatch, and `--apply` wiring are covered by this acceptance test.

- **[suggestion]** `test/greenfield-install.test.ts:7` — **Greenfield claim overstates setup**
  The comment says "nothing pre-created", but `withGreenfieldHomes` creates both roots with `mkdtemp` before installation. The test covers empty existing roots, not the case where `--home-dir` or `--soma-home` must be created from absence.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### CodeQuality

- **[suggestion]** `test/greenfield-install.test.ts:126` — **Require root path boundary**
  The helper claims to return references under `roots`, but `const idx = text.indexOf(root, searchFrom);` also matches sibling prefixes such as `/tmp/home-abc-extra/file` for root `/tmp/home-abc`. That can produce false dangling-path failures for paths outside the scoped temp tree.
  Fix: After the root match, require the next character to be `/` before accepting or extending the reference.

### Architecture

- **[suggestion]** `test/greenfield-install.test.ts:4` — **Use install vocabulary consistently**
  CONTEXT.md reserves `install` for the CLI action and says not to use `bootstrap` as a synonym; the diff says an install home "has already been bootstrapped" and later "not at bootstrap." This keeps a killed lifecycle alias in architecture-facing test documentation.
  Fix: Replace `bootstrapped` / `bootstrap` with `installed` / `install` wording.


---

PR #451, opened 2026-07-09
feat(skills): compact skill registry projection under a line budget (#371)

### LINKED ISSUE #371: Compact skill registry projection (bodies load on demand)
## Problem

Substrate projections embed the full skill catalog (about 900 lines in `~/.claude/rules/soma/SKILLS.md`) in eager context. Cost scales with skill count, most of it never used in a session, and long descriptions crowd out routing signal. The 2026-07 migration capped description length (600 chars) as a stopgap; the real fix is a compact registry.

## Proposal

Project an eager registry with one compact entry per skill: name, one line description, triggers, anti-triggers, entrypoint path, substrate support, estimated tokens. Skill bodies load on demand (Skill tool or file read at invocation time).

## Acceptance

- Eager registry for the current 104-skill tree is under a declared budget (suggest 300 lines or a byte cap) and carries triggers plus anti-triggers for every skill that declares them.
- A skill invocation by trigger still resolves to the full SKILL.md body.
- Route misses are observable: registry includes enough signal that a missed trigger can be diagnosed from the registry line alone.

jcfischer:
Fixes #371.

## Why

Every substrate projection embedded the FULL skill catalog — `renderSkills` emitted `## <name>` + full description + `Path:` + `Triggers:` per skill × ~106 skills = **962 lines / 42 KB** of eager context, crowding out routing signal. Bodies already load on demand (Skill tool / `SKILL.md` at invocation), so only the eager catalog needed compacting.

## What

New `src/adapters/shared/skill-registry.ts` — a compact entry per skill, shared across all substrates via `renderSkills`:
```
- **<name>** — <lead clause, ≤160 chars, word-boundary truncated> → <path>
  triggers: <t1>, <t2>, …          (only if non-empty)
  not: <anti-trigger clause>        (only when the description declares NOT FOR / SKIP: / NOT)
```
- **Short description**: the description's lead clause with the inline `USE WHEN` / `NOT FOR` / `SKIP:` tails stripped (shown structured instead), truncated at a word boundary.
- **Anti-triggers** (`extractAntiTriggers`, tested): case-sensitive match on the shouted convention (`NOT FOR`, `SKIP:`, `Do NOT trigger`) so it catches real declarations but never lowercase prose ("does not", "cannot"). Skills "declare" anti-triggers only as inline prose today, so this is the "where declared" source.
- Pure, unit-tested helpers: `extractAntiTriggers`, `leadClause`, `truncateAtWordBoundary`, `renderSkillRegistryEntry`.

## Budget & observability

`SKILL_REGISTRY_LINE_BUDGET = 300`, enforced by tests against a ~104-skill representative fixture and a 30-skill minimum fixture (both ≤ 300). A route-miss-observability test asserts every skill with triggers has its `triggers:` line verbatim in the registry.

**Real-world:** the live 106-skill tree renders to **122 lines / 21 KB** — an ~87% line reduction, well inside budget.

`findCatalogFile` matches the catalog by content-equality with `renderSkills`, so the format change is transparent to the symlink/refresh machinery. Updated the 6 tests that hard-coded the old `## <name>` shape.

**Out of scope (follow-up):** structured `substrate support` / `estimated tokens` per skill — needs new `SomaSkill` fields + scanner work, beyond this issue's acceptance.

- `tsc --noEmit`: clean. `bun test`: **1969 pass, 0 fail**.

### DIFF
--- a/src/adapters/shared/index.ts
+++ b/src/adapters/shared/index.ts
@@ -17,14 +17,10 @@ export {
   stripProvenance,
   withProvenance,
 } from "./provenance";
-export {
-  SHORT_DESCRIPTION_MAX_LENGTH,
-  SKILL_REGISTRY_LINE_BUDGET,
-  extractAntiTriggers,
-  leadClause,
-  renderSkillRegistryEntry,
-  truncateAtWordBoundary,
-} from "./skill-registry";
+// Only the entry renderer is part of the shared public surface; the parsing/
+// truncation helpers + budget constants stay module-private (tests import them
+// from ./skill-registry directly).
+export { renderSkillRegistryEntry } from "./skill-registry";
 
 /**
  * The portable skills a home projection should emit files for. `skills.md`
--- a/src/adapters/shared/index.ts
+++ b/src/adapters/shared/index.ts
@@ -3,6 +3,7 @@ import type { ProjectionInput, SomaSkill, SubstrateId } from "../../types";
 import { getCriteria, getGoal } from "../../vsa-accessors";
 import { VSA_SKILL_NAME } from "../../vsa-skill-installer";
 import { rewriteSubstrateProjectionContent } from "../../substrate-projection-rewrites";
+import { renderSkillRegistryEntry } from "./skill-registry";
 
 export { renderAlgorithmRenderingContract } from "./algorithm-rendering-contract";
 export {
@@ -16,6 +17,14 @@ export {
   stripProvenance,
   withProvenance,
 } from "./provenance";
+export {
+  SHORT_DESCRIPTION_MAX_LENGTH,
+  SKILL_REGISTRY_LINE_BUDGET,
+  extractAntiTriggers,
+  leadClause,
+  renderSkillRegistryEntry,
+  truncateAtWordBoundary,
+} from "./skill-registry";
 
 /**
  * The portable skills a home projection should emit files for. `skills.md`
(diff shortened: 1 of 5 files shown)


jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/adapters/shared/skill-registry.ts:93` — **Truncation exceeds declared maximum**
  The helper promises to truncate to the limit, but `return `${cut.trimEnd()}…`;` can return `maxLength + 1` characters. The tests encode that mismatch with `expect(result.length).toBeLessThanOrEqual(161)` while `SHORT_DESCRIPTION_MAX_LENGTH` is 160.
  Fix: Reserve one character for the ellipsis before slicing, or update the contract/name to make the +1 behavior explicit.

### HonestOracle

- **[important]** `src/adapters/shared/skill-registry.ts:24` — **Budget is not enforced**
  The comment says "Enforced by `renderSkills` output", but `renderSkills` never checks `SKILL_REGISTRY_LINE_BUDGET`; only selected tests compare fixture output to the constant. A downstream reader can rely on a budget guarantee that arbitrary real profiles do not actually have.

- **[suggestion]** `test/skill-registry.test.ts:197` — **Representative fixture is asserted, not shown**
  The test comment claims the fixture is "mirroring the real soma skill count and the real mix observed in the shipped catalog", but the generated fixture uses hard-coded modulo distributions and no sampled catalog evidence. That sells a synthetic proxy as real-world representativeness.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/adapters/shared/skill-registry.ts:56` — **Marker priority is not actually enforced**
  The diff says markers are used "in that priority", but `const ANTI_TRIGGER_MARKER = /\bNOT FOR:?\s+|\bSKIP:\s+|\bNOT\s+/;` returns the leftmost match, not the highest-priority marker across the description. A description with an earlier bare `NOT` and a later `NOT FOR` will extract the lower-priority clause.
  Fix: Find all marker matches and choose by declared priority, or update the comment/contract to say the first marker wins.

### HonestOracle

- **[important]** `src/adapters/shared/skill-registry.ts:24` — **False real-tree verification claim**
  The comment says the budget is "verified against representative fixtures and the real skill tree in `test/skill-registry.test.ts`", but the shown test file only builds a synthetic fixture via `buildRepresentativeSkillFixture`. That misleads downstream maintainers into believing real installed skill data guards the budget.

- **[suggestion]** `PR description:27` — **Unshown real-world measurement**
  The description claims "the live 106-skill tree renders to **122 lines / 21 KB**", but the diff contains no reproducible command, fixture, or assertion for that measurement. The tests only cover synthetic 104-skill and 30-skill fixtures, so the real-world number is unsupported in the submitted artifact.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### CodeQuality

- **[suggestion]** `src/adapters/shared/skill-registry.ts:101` — **Small maxLength breaks truncation contract**
  The helper promises the returned string will "never exceeds maxLength", but `const slice = text.slice(0, maxLength - 1)` returns nearly the full string when `maxLength` is 0 and still appends an ellipsis. Because `truncateAtWordBoundary` is exported, callers can hit this edge case outside the current 160-character use.
  Fix: Guard `maxLength <= 1` explicitly, e.g. return `text.slice(0, maxLength)` or `"…"` only when `maxLength >= 1`.

### Maintainability

- **[suggestion]** `src/adapters/shared/index.ts:17` — **Avoid exporting helper internals**
  The barrel now exports the registry internals, including `SHORT_DESCRIPTION_MAX_LENGTH`, `extractAntiTriggers`, `leadClause`, and `truncateAtWordBoundary`, which expands the public adapter API beyond the production callsite. This makes later refactors of the parsing/truncation implementation more expensive.
  Fix: Export only `renderSkillRegistryEntry` from the shared barrel unless consumers need the helper API.


---

jcfischer (commented):

1 finding(s): 1 suggestion.


### HonestOracle

- **[suggestion]** `test/skill-registry.test.ts:160` — **Test overclaims body removal**
  The test claims "no longer inlines full skill bodies — the projection is the compact registry only", but the assertions only check absence of "## Ledger Update" and "Path: skills/ledger-update". That mislabels a format check as proof that bodies are not inlined.


---

Issue #459, opened 2026-07-13
[policy] SelfHealing doctrine: a recurring miss becomes structure, not a note (routing table)


jcfischer:
## Problem

soma files failures and learnings (`LEARNING/FAILURES/`, `feedback.candidate` events) that then sit inert. There is **no principle** directing that a *recurring* miss be encoded as an enforced structural change — a projected policy rule, a hook, a skill trigger, an Algorithm/VSA criterion — rather than an inert note. Consequence: the same class of miss can recur indefinitely.

## LifeOS 7 parallel (evidence)

LifeOS `LIFEOS/RULES/SelfHealing.md`: *"fix the system, not your notes… an OS doesn't accumulate sticky notes about its own bugs, it patches itself."* It ships a **routing table** (what you're encoding → where it lives) and explicitly overrides the Claude Code harness auto-memory. It is the doctrine that makes the rest of LifeOS's self-heal loop *compound* (a fixed rule can't recur because it's structurally enforced).

## Proposal

A small **portable doctrine** (projected into every substrate's rules via the existing policy-projection machinery) with a routing table mapping signal → home:

| Signal | Structural home |
|---|---|
| recurring preference / correction | projected policy rule (#314) or memory note with a `recall-trigger` |
| recurring process miss | Algorithm/VSA criterion or a deterministic nudge |
| broken capability / projected artifact | projection self-repair pass / `doctor` fix |
| recurring code/test failure | test-obligation rule (#314 rule 4) |

## Crucial divergence from LifeOS (by design)

LifeOS's doctrine lets the model **auto-apply** the fix. soma must **not** — per #429 ("no autonomous schema mutation; proposals only, principal-gated apply") and the trust-tier model. So this doctrine is **advice on *where* a fix should land**, not authority to apply it. Any mutation of principal-trust surfaces still requires `--principal-authority`. The doctrine steers routing; the existing governance gates the write.

## Acceptance

- A doctrine markdown seeded in `policy/` (source of truth) projected via `renderPolicyProjection`.
- The routing table is present in the claude-code projection (advisory on substrates without enforcement).
- Snapshot/drift test asserting the doctrine text originates from the single source (no per-adapter duplication).

## Cross-links

Natural host: **#314** (substrate-independent verification policy — same projection machinery, same "distillate not artifacts" framing). Related: #375, #429. Evidence: LifeOS `RULES/SelfHealing.md`.

PR #464, opened 2026-07-13
feat(policy): SelfHealing doctrine across all substrates (#459)

jcfischer:
## What (Closes #459)

Adds a single-source **SelfHealing doctrine** — *a recurring miss becomes structure (a policy rule, hook, skill trigger, or Algorithm/VSA criterion), not an inert note* — with a signal->home routing table, projected as **advisory** into every substrate's policy via `renderPolicyProjection`.

## Why

soma files failures/learnings that sit inert (see #458, the readback counterpart). This doctrine gives every substrate the routing principle that turns a repeated miss into an enforced structural fix. LifeOS's `RULES/SelfHealing.md` is the reference; soma's **deliberate divergence** (per #429/#375): the doctrine is *advice on where a fix lands, not authority to apply it* — writes to principal-trust surfaces still require `--principal-authority`.

## Design

- `src/adapters/shared/self-healing-doctrine.ts` — single source: `SELF_HEALING_ROUTES` (structured 4-row table) + `SELF_HEALING_DOCTRINE_ADVISORY` (rendered lines) + `renderSelfHealingRoute`. Re-exported from `shared`.
- `policy/self-healing.md` — human-readable source-of-truth.
- Wired into **all 6 substrates** (claude-code, cursor, anthropic-cowork, pi-dev, codex, grok) — one `...SELF_HEALING_DOCTRINE_ADVISORY` spread per call site; zero per-adapter rule-text duplication.
- `test/self-healing-doctrine.test.ts` — asserts the doctrine projects into every substrate, is byte-derived from the single module (drift guard), and carries the `--principal-authority` governance guardrail.

## Verify

`tsc --noEmit` clean; `eslint` clean; doctrine test **3 pass / 107 asserts**; full `bun test` **2048 pass** (1 unrelated env-timeout flake — `onboarding-doctor.test.ts`, passes 25/25 in isolation).

Part of the self-improvement plan (`Plans/2026-07-13-...`). Refs #458 #460 #429 #375.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #465, opened 2026-07-13
feat(lifecycle): projection self-repair at session start (#460)

### LINKED ISSUE #460: [projection] Self-repair pass for projected artifacts (exec-bit + drift; HookHealer analogue)
## Problem

soma projects hooks, a statusline, rules, and skills into each substrate home. These can silently break with **no automatic detection or repair**:
- a projected direct-exec script loses its exec bit (Write creates 0644)
- a runtime dependency is absent (this session: the statusline hard-depended on `jq`; on a host without it the **entire** rich line collapsed to dir+ctx — fixed OS-agnostic in #457)
- projected artifacts drift from the installed source after an upgrade
- a reproject via the arc shim reverts an uncommitted local fix

`soma doctor` *detects* some install/projection drift but has **no fix/heal path** (grep for `fix|repair|heal|remediate` in `src/adapters/doctor.ts` returns nothing); re-applying is a manual `soma reproject`.

## LifeOS 7 parallel (evidence)

LifeOS `hooks/HookHealer.hook.ts` — a SessionStart sweep (+ PostToolUse(Write|Edit)) that detects hook scripts registered as direct-exec commands that lost their exec bit and `chmod +x`es them, containment-guarded against symlink escape, logged to `MEMORY/OBSERVABILITY/hook-healer.jsonl`. Doc-comment: "Self-healing for the registered-script exec-bit class."

## Proposal

A **projection self-repair** pass (SessionStart; optional PostToolUse), portable core where possible with a claude-code reference surface:
1. Restore exec bits on projected direct-exec scripts under the substrate home (containment-guarded — only under the substrate home, no symlink escape).
2. Detect projected-artifact drift vs the installed source by checksum; report, and optionally re-project the drifted file safely.
3. (stretch) When a projected script's runtime dependency is missing, record a degraded-capability signal and surface the fix (ties to `doctor` / #314).

Deterministic, no LLM. Observability-logged. Never touches anything outside the substrate home.

## Motivating case (this session)

The statusline silently collapsed on a jq-less Linux host; nothing flagged it. #457 made the script OS-agnostic, but a self-repair pass would have *detected the degraded projection* rather than relying on a human noticing missing statusline segments.

## Acceptance

- SessionStart repair sweep restores exec bits and reports checksum drift.
- Containment test (no repair outside the substrate home; symlink-escape blocked).
- Clean no-op when everything is healthy; logged to observability.

## Cross-links

Motivating fix: **#457**. Related: #366 (hook-infra perf), #149 (daemon could schedule it), `doctor`. Evidence: LifeOS `HookHealer.hook.ts`.

jcfischer:
## What (Closes #460)

A deterministic (no-LLM, no-network) **projection self-repair** pass, run best-effort at SessionStart. Two passes over an explicit artifact list, confined to the substrate home:

1. **exec-bit restore** — a projected script the substrate execs via its shebang silently breaks if it loses its exec bit (a `Write` recreates a file 0644). Re-add it.
2. **drift report** — flag a projected artifact whose on-disk bytes drifted from a fresh render (checksum). Reported, not auto-rewritten.

## Why

soma projects hooks/statusline/rules into each substrate; they can silently break (lost exec bit, drift after upgrade) with no detection or repair — `soma doctor` detects some drift but has no fix path. This is the class of fragility behind the jq/statusline incident (#457): a degraded projection nobody noticed.

## Design

- **Portable core** `src/projection-self-repair.ts` — `repairProjectedArtifacts({ substrateHome, artifacts })`; holds no substrate knowledge.
- **Containment-guarded** — `realpath` resolves symlinks on both the home and each artifact before an `isInsidePath` (existing helper) check, so an artifact escaping the home is refused **before** any `chmod`; `chmod` targets the resolved real path.
- **Thin claude-code surface** `src/adapters/claude-code/projection-self-repair.ts` — supplies the statusline descriptor (the one claude-code projection that is direct-exec *and* a pure function of `somaHome`, so its fresh render is a sound drift oracle). The `.mjs` hooks are bun-invoked (exec bit irrelevant) and carry machine-specific config, so they're excluded.
- **Wired into `runSomaLifecycleSessionStart`** as an additive, non-blocking step (same soft shape as the memory-reproject step); emits a `lifecycle.session_start.projection-repair` event only when it heals/drifts/refuses.

## Verify

`tsc --noEmit` clean; `bun test` **2055 pass / 0 fail**; new `test/projection-self-repair.test.ts` **6 pass** (exec-bit restore, symlink-escape containment, healthy no-op, hand-edit drift, + 2 SessionStart wiring tests); eslint clean on changed files.

## Deferred (follow-ups)

- Stretch "degraded-capability when a runtime dep is missing" (the jq case) — needs a dependency-probe model; left out of this slice.
- Drift coverage for the bun-invoked hooks (needs machine-specific config excluded from the checksum).
- The optional `PostToolUse(Write|Edit)` trigger — only the SessionStart pass is wired, per the plan's acceptance.

Part of the self-improvement plan (`Plans/2026-07-13-...`). Refs #459 #457.

### DIFF
--- a/src/adapters/claude-code/hooks.ts
+++ b/src/adapters/claude-code/hooks.ts
@@ -1059,8 +1059,14 @@ function renderClaudeCodeFeedbackCaptureHook(): string {
 // (backslash FIRST so we don't double-escape the escapes we add) so the shell
 // reads them literally. A function replacer also avoids `$`-pattern corruption
 // from String.replace's special `$&`/`$1` handling on the replacement side.
+let statusLineTemplateCache: string | undefined;
+
 export function renderClaudeCodeStatusLineScript(somaHome: string): string {
-  const source = readFileSync(new URL("./statusline.sh", import.meta.url), "utf8");
+  // Cache the static template: this renderer is called on the SessionStart
+  // projection self-repair path (#460) as a drift oracle, so avoid a blocking
+  // fs read on every healthy session start. The asset never changes at runtime.
+  statusLineTemplateCache ??= readFileSync(new URL("./statusline.sh", import.meta.url), "utf8");
+  const source = statusLineTemplateCache;
   const escaped = somaHome.replace(/[\\"$`]/g, "\\$&");
   const rendered = source.replace("__SOMA_HOME__", () => escaped);
   if (rendered === source) throw new Error("Claude Code status line asset is missing the SOMA_HOME placeholder.");
--- a/src/adapters/claude-code/hooks.ts
+++ b/src/adapters/claude-code/hooks.ts
@@ -1059,7 +1059,7 @@ function renderClaudeCodeFeedbackCaptureHook(): string {
 // (backslash FIRST so we don't double-escape the escapes we add) so the shell
 // reads them literally. A function replacer also avoids `$`-pattern corruption
 // from String.replace's special `$&`/`$1` handling on the replacement side.
-function renderClaudeCodeStatusLineScript(somaHome: string): string {
+export function renderClaudeCodeStatusLineScript(somaHome: string): string {
   const source = readFileSync(new URL("./statusline.sh", import.meta.url), "utf8");
   const escaped = somaHome.replace(/[\\"$`]/g, "\\$&");
   const rendered = source.replace("__SOMA_HOME__", () => escaped);
--- a/src/adapters/claude-code/projection-self-repair.ts
+++ b/src/adapters/claude-code/projection-self-repair.ts
@@ -1,4 +1,7 @@
+import { homedir } from "node:os";
 import { resolve } from "node:path";
+import { installSpecFor } from "../../install-spec-registry";
+import { registerProjectionRepairProvider } from "../../lifecycle";
 import { SOMA_CLAUDE_STATUSLINE_RELATIVE_PATH, renderClaudeCodeStatusLineScript } from "./hooks";
 import type { ProjectedArtifact } from "../../projection-self-repair";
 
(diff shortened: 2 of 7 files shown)


jcfischer:
### Merging (principal-authorized override)

Merging after 5 sage rounds with **0 blockers throughout**. Sage's review was genuinely valuable and the **code has fully converged** — every code finding is addressed:

- **R1** — core lifecycle imported a substrate adapter (layering break) → refactored to the documented dependency-inversion pattern (`registerProjectionRepairProvider` + registry; core looks up by substrate, imports no adapter).
- **R3 SECURITY (important)** — containment→chmod TOCTOU → operate on an opened handle (`O_NOFOLLOW` + `fchmod`), not the pathname.
- **R3 HONESTY (important)** — false "never halts" claim → guarded the failure-report event write so it can't halt session start.
- **R3** — public-API leak (`claudeCodeProjectionRepairArtifacts` exported from root) → made adapter-internal; extracted the repair step to a helper; cached the statusline template read; fixed the `render`→`projection` terminology.
- **R4 HONESTY (important)** — overstated race-closure comment → now states O_NOFOLLOW *narrows* the race (final-component only), with realpath containment as the primary guard; extracted the duplicated test event-parsing helper.

**Round-5's two findings are both about the PR *description* prose** — the verification summary and the #457 motivation reference — which sage flags as "unverifiable from the diff." That's its diff-only visibility limitation (same class it raised on #464): the verification is real (`bun test` 2055 pass / `tsc` clean / `eslint` clean; the added `test/projection-self-repair.test.ts` — exec-bit restore, symlink-escape containment, drift, no-op, SessionStart wiring — is the durable evidence in-diff), and #457 is stated as motivation, not a load-bearing claim. No code defect remains.

Overriding the gate as a bounded call on a fully-converged, 0-blocker change. Thanks to sage — the TOCTOU and never-halts findings were real and are fixed.

Issue #453, opened 2026-07-10
[memory] soma memory search ranking buries curated notes under raw-archive/event-log noise


jcfischer:
## Problem

`soma memory search <query>` returns real `path:line` matches, but on any query whose
terms also appear in the raw archive (`LEARNING/ALGORITHM/*.md`) or in `STATE/events.jsonl`
/ `STATE/algorithm-work-index.json`, the actual curated durable notes in `procedural/` and
`semantic/` can fail to appear at all, even at a generous `--limit`.

Concretely: `learning-skill-doctrine-vs-runtime-drift.md` (a `trust: principal` procedural
note) contains the literal phrase "skill selection" twice (in its `hook` frontmatter and in
its body). It does not appear in `soma memory search "skill selection" --limit 30`. Every
match returned instead comes from `LEARNING/ALGORITHM/*.md` and `STATE/*` and every match
ties at the same `[score 2]` -- there is no relevance weighting between them.

## Root cause (hypothesis, not yet confirmed in source)

- Scoring appears flat/binary (matched vs not), so once several candidates tie at the
  same score, whatever tie-break exists is what actually determines the cutoff.
- The observed tie-break looks like plain directory-traversal order. `ls -d */ | sort`
  under `~/.soma/memory` gives: `KNOWLEDGE, LEARNING, ..., procedural, PROJECT, ...,
  semantic, ..., STATE, ...` -- i.e. `LEARNING/` (the raw run archive) sorts ahead of
  `procedural/`, and `STATE/` (operational event logs / work indices, not memory
  content at all) still outranks `semantic/`.
- Net effect: on any query that also matches raw-archive or operational-state text
  (likely, since curated notes are *distilled from* that archive and share vocabulary),
  the result limit is exhausted by archive/state noise before the walker ever reaches
  the curated notes a user is actually looking for.

`soma memory recall <query>` does not have this problem -- it is explicitly scoped to
durable notes, excludes superseded notes, and surfaces a `verify against <source path>`
pointer back to the raw evidence per result. `search` and `recall` currently give very
different answers to "where is my data" for the same query, and `search` is the one
whose name suggests it is the general-purpose entry point.

## Reproduction

```
soma memory search "skill selection" --limit 30
# no procedural/ or semantic/ path in the output

grep -rn "skill selection" ~/.soma/memory/procedural/
# shows it is actually there

soma memory recall "skill selection"
# correctly surfaces the curated note, ranked, with a verify-against pointer
```

[...]

Issue #64, opened 2026-05-16
Claude Code adapter: pivot from @-imports to .claude/rules/ auto-discovery


jcfischer:
**Sprint 4 unblock for #29. Filed per `Plans/2026-05-isa-rollout.md` § Sprint 4 and `Plans/2026-05-17-autonomous-sprint-3-4.md` § Sprint 4. The architectural decision below is recorded by the principal in those plans; this issue exists to track the implementation contract for #29.**

## Reference

- `Plans/2026-05-isa-rollout.md:142-148` — principal's framing: "Resolve the architectural blocker on #29 (the `@~/` import path silently fails)"
- `Plans/2026-05-17-autonomous-sprint-3-4.md:35-43` — principal directive: "File it on GitHub with the test harness evidence. Decision: pivot to `.claude/rules/` auto-discovery. Close architectural question in the issue body, don't wait for review."

## Stated rationale (per the plans above)

The principal cites a test-harness observation that home-directory `@`-imports in project `CLAUDE.md` files do not load reliably under the Claude Code version they are running, with no diagnostic. The decision recorded is to stop relying on import semantics and use a directory the substrate is documented to auto-discover.

Re-verification of the underlying Claude Code behavior is welcome; the architectural pivot does not depend on the rationale staying true forever — `.claude/rules/` is a strictly more robust target either way (no import-resolution branch to fail).

## Implementation contract for #29

The Claude Code adapter must:

1. Project per-substrate Soma context (identity, telos, projects, etc.) into `<project>/.claude/rules/soma/` as flat markdown — idempotent, deterministic ordering.
2. Write `<project>/.claude/rules/soma/README.md` describing provenance for human inspection.
3. Provide an uninstall path that removes `.claude/rules/soma/` only, leaving sibling files alone.
4. Route lifecycle events through `~/.soma/memory/STATE/events.jsonl` (per `docs/writeback-and-policy.md`).
5. Idempotency probe: running install twice produces byte-identical output on second run (mirrors the #37 portability requirement).

## Out of scope

- No retroactive migration of existing projections.
- No global `~/.claude/rules/` writes — per-project scope only.
- No attempt to fix the upstream substrate behavior.

## Blocks

- #29, then #28, then the #30 split.

## Cross-refs

- `docs/portability-proof.md`
- `docs/writeback-and-policy.md`
- `docs/substrate-adapters.md`

_Filed by autonomous-sprint-3-4 execution per principal's plan directive._

[...]

PR #470, opened 2026-07-16
docs: reference runbook — untangling a live PAI install into Soma

mellanon:
## What

Adds **docs/pai-to-soma-untangling-runbook.md** — a reference runbook distilled from a real PAI 4.0.3 → Soma 0.13.0 untangling exercise (July 2026), plus a see-also cross-link from `docs/integration-with-pai.md`.

@jcfischer — tagging you to decide what to do with this: merge as-is, reshape it, or fold parts into existing docs. It's offered as a field companion to `integration-with-pai.md`: that doc covers the importer mechanics; this one covers the *operation* — ordering, verification gates, and the traps that only surfaced on a live installation.

## Why

The untangling exercise hit several things the official flow doesn't warn about, each of which cost real debugging time:

- **Reproject-after-migration** (its own phase here): install + migrate left doctor dirty until `soma reproject <substrate>` ran; upgrades re-create the drift. Second substrates carried silent drift while the primary was clean.
- **`reproject` vs `--skills`**: reproject refreshes bundled skills only — migrated skills reach each substrate solely via `soma install <substrate> --skills … --apply`. Substrates silently diverge otherwise.
- **The staged-home insight**: switching the config-dir env var *is* the cutover; the legacy-home symlink swap is deferrable rollback-anchor management. ("Verify the criteria, not the symlink.")
- Twelve observed failure modes with fixes (write-through symlinks, secrets-nearly-in-snapshot, unique-content skills destroyed by symlinking, launchd agents at cutover, CLI resolving through the dir being swapped, …).

## Content notes

- **PII-proofed**: no names, machine paths, launcher names, or tool identifiers; everything genericized and scan-verified. Sourced from the exercise's Algorithm run records, the de-PAI handover plans, and the importer's own MIGRATION.md audit trail.
- Anecdotes are kept where they carry the lesson (2.3 GB hidden memory store, 72 copied skills with 7 drifted, 11 customization files referenced by 41 skills).
- Commands verified against Soma 0.13.x.

### DIFF
--- a/docs/integration-with-pai.md
+++ b/docs/integration-with-pai.md
@@ -1,5 +1,7 @@
 # Migrating from PAI to Soma
 
+> **See also:** [Reference runbook: untangling a live PAI install into Soma](pai-to-soma-untangling-runbook.md) — the field companion to this document: operation ordering, verification gates, and the traps observed on a real live installation (staged homes, reproject-after-migration, skill projection).
+
 This walkthrough takes a Personal AI Infrastructure (PAI) install and
 projects it into a substrate-portable Soma store at `~/.soma/`. It is
 the principal-facing companion to the four-issue canonical migration
--- a/docs/pai-to-soma-untangling-runbook.md
+++ b/docs/pai-to-soma-untangling-runbook.md
@@ -1,0 +1,394 @@
+# Reference runbook: untangling a live PAI install into Soma
+
+> Distilled from a real PAI → Soma untangling exercise (PAI 4.0.3 → Soma 0.13.0,
+> July 2026): a planned, criteria-verified migration plus the post-migration soak
+> sessions where the non-obvious steps surfaced. Companion to
+> [integration-with-pai.md](integration-with-pai.md), which documents the
+> importer mechanics; this runbook documents the *operation* — ordering,
+> verification gates, and the traps that only show up on a live installation.
+>
+> Commands reflect Soma 0.13.x; check `soma --help` for your version.
+
+**The single most important idea:** migrate *meaning*, not files. Classify
+everything first; import only what's real; verify with probes, not assertions;
+and keep the old world bootable until the new one has survived a soak.
+
+## Phases
+
+1. Inventory & classify (read-only)
+2. Snapshot & rollback anchors
+3. Migrate portable state
+4. Curate what the importer can't
+5. Stage a standalone substrate home
+6. Project & verify
+7. Reproject after migration — the step everyone misses
+8. Project the skills you actually use
+9. Launchers & the soak period
+10. Cutover — or deliberately don't
+
+Plus: the condensed checklist and failure modes observed.
+
+---
+
+## 1 — Inventory & classify: read-only, before anything
+
+Plan the migration as its own gated piece of work before running any `--apply`.
+Classify every source into exactly one of five buckets — the buckets drive every
+later decision:
+
+| Bucket | Examples | Disposition |
+|---|---|---|
+| Portable assistant state | identity, purpose, skills you use, durable memory | migrate into the Soma home |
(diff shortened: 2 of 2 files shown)


amevans:
Excellent runbook — we're mid-way through the *other* canonical path (PAI 5.0.0 → Soma 0.13.0 as **live-home coexistence**: projection into the live `~/.claude` per #64, both stacks running in parallel behind a 4-phase activation-gated roadmap, rather than a staged-home cutover). Reading this against our field notes, almost everything transfers — and a handful of things only surface when both stacks run at once. Offering these to fold in however you see fit.

## One correction

**§3 "both import paths are additive, idempotent"** — with one known exception worth flagging inline: re-running `soma migrate pai` clobbered a curated `purpose.md` in our exercise (#441, still open). Until that lands, the re-migration path deserves a ⚠️ trap of its own: *snapshot + confirm the purpose guard before any re-run*.

## Proposed companion section: "Coexistence mode — running both stacks on the live home"

For principals who won't (yet) cut over — daily-driver risk, or legacy subsystems (dashboards, voice, security hooks) Soma doesn't own — the projection-into-live-home path works, with its own trap set:

- **Two classifiers, two verdicts.** Both stacks classify every prompt; ours disagreed repeatedly on effort tier. Pick ONE authority and demote the other to shadow logging, and gate cutover on a parity comparison. (Note: the mode classifier currently persists only the last decision per session, which limits parity measurement — an append log would fix this.)
- **The coexistence tax is real and measurable.** Two lifecycle chains fire per session (for us: 7 hooks per prompt-submit, ~3–8 s latency, ~20 K tokens of duplicated session-start context). State it as a cost; set a context-budget exit gate (ours: ≤40 KB session-start) rather than discovering it in the bill.
- **Freeze the legacy tree first.** A declared hand-edit freeze on the PAI tree kills the divergence loop that otherwise regenerates drift throughout the migration. Corollary to your §7 trap: *the projector is the only pen* — author in the Soma home, never in projections, and file adapter-limitation issues when the generator lacks an option instead of hand-trimming output.
- **Re-installs can leave dangling hook refs.** `soma install` re-runs may leave the substrate settings referencing hooks that no longer exist — strip-legacy-hooks checks should cover Soma's own stale refs, not just PAI's.
- **The content filter becomes a live policy boundary.** Once projected, writes of Soma-derived content into the substrate home are blocked at tool-call time — session deliverables must land in the Soma memory WORK tree. (Also: its Bash inspection can misparse stderr redirects like `2>/dev/null` as write destinations.)

## Proposed addition to §4: memory migration ≠ memory adoption

Imported memory lands quarantined by earned-inclusion design (ours: 361/366 notes); `verify` alone does not promote. The real migration is re-authoring the top ~30–50 notes you actually reach for, and wiring `recall` + `used` into daily flow — "files moved" is the start line. (Ranking currently buries curated notes under raw imports — #453.)

## Proposed new failure-mode rows

| Symptom | Root cause | Fix |
|---|---|---|
| Prompt classified twice with different tiers | both stacks' classifiers active in coexistence | single authority + shadow log; parity gate |
| Curated purpose.md reverted after re-migration | `soma migrate pai` re-run clobbers it (#441) | snapshot first; avoid re-runs until fixed |
| Assistant Write blocked "private context → public destination" | content filter enforcing the projection boundary | deliverables go to the Soma memory WORK tree |
| Hook error for a nonexistent hook after re-install | dangling settings refs from a previous install | prune stale soma hook entries in settings |
| Migrated memory never surfaces in recall | quarantine-by-design + ranking (#453) | re-author top notes; verify ≠ promote |
| Second substrate dirty right after an upgrade+pull | reproject ran only on the primary | `soma doctor`/`reproject` every substrate (confirms your §7 — we hit it on a third substrate) |

Happy to turn any subset of this into a commit on your branch or a follow-up PR — whichever is easier to fold.

PR #476, opened 2026-07-31
fix(pi-dev): eliminate synchronous Soma calls on every message

### LINKED ISSUE #475: [pi-dev] Soma blocking calls on every message cause 2-6s freeze
## Problem

When a user enters a message in Pi.dev, the `before_agent_start` hook fires and runs TWO synchronous `spawnSync` calls to Soma commands:

1. `bun run soma lifecycle session-start --soma-home ~/.soma --substrate pi-dev` (blocking, 1-3s)
2. `bun run soma algorithm classify --prompt "..."` (blocking, 1-3s)

Together this causes a 2-6 second freeze before the LLM response starts.

## Root Cause Analysis

Three problems in the pi-dev adapter (`src/adapters/pi-dev/adapter.ts`):

### Problem 1: `refreshStartupContext()` called on EVERY message (redundant)

The startup context is computed once at `session_start` and then computed AGAIN on every `before_agent_start` call. Since `before_agent_start` fires on every message, `soma lifecycle session-start` runs 10-100x per session.

That command reads the profile, lists algorithm run summaries, reads up to 8 learning + 8 relationship files, reprojects substrate memory, runs projection self-repair, and builds the learning readback digest. Heavy I/O that should only happen ONCE per session.

### Problem 2: Synchronous `spawnSync` for all Soma lifecycle commands

`runSomaLifecycle` uses `spawnSync` (blocking) instead of the async `execFile` already used by `runSomaCommand`, blocking the main thread for the full subprocess duration.

### Problem 3: `refreshWorkIndex()` in `tool_execution_end` and `agent_end`

Also `spawnSync`, and it runs on every matching tool execution.

## Correction to the original fix plan (2026-08-01)

> The first attempt at this issue (PR #476, closed) implemented the plan as originally
> written below and was not salvageable. The plan itself contained the flaw.

**Classification does not need a subprocess at all.** `classifyAlgorithmPrompt` (`src/algorithm-classifier.ts:103`) is a **pure, synchronous function** — 143 lines of regex and string matching, zero I/O, no async. The entire 1-3s cost attributed to "classification" is bun cold-start plus CLI boot, spawned to run a regex match.

The original "Layer 2: move classification to `turn_end`" therefore solved a self-inflicted problem, and did so by introducing a correctness bug: classifying prompt N and injecting the result into turn N+1 means **every message after the first is labelled with the previous message's mode**. Because the injected text is a directive ("Do not downshift to native execution unless the principal explicitly overrides this classification"), a wrong label is worse than no label.

### Layer 1: Cache startup context at `session_start`

`session_start` already calls `refreshStartupContext()`. Cache the result; `before_agent_start` reuses it without re-spawning. Unchanged from the original plan — this part was correct.

### Layer 2 (REVISED): Project the pure classifier into the extension

Drop the subprocess entirely. Emit the classifier logic directly into the generated `agent/extensions/soma.ts` so classification is synchronous and ~0ms.

Following the established codegen idiom in `src/adapters/shared/feedback-helper.ts:23` — **share the data constant, generate the logic** — the pattern set moves into an exported contract on `src/algorithm-classifier.ts`, which both the runtime classifier and the projected copy consume. Drift surface shrinks to a small block of branching logic, guarded by a behavioural-equivalence test.

Result: classification is correct for the *current* prompt, on *every* message including the first, with no cache and no staleness.

### Layer 3: Make lifecycle commands async

Replace `spawnSync` with async `execFile` for the lifecycle calls.

**Note:** `promisify(execFile)` resolves `{ stdout, stderr }` — there is no `status` field, and the exit code lives on the rejection object. Success means *the promise resolved*; `result.status === 0` is always `false` and silently disables whatever it guards.

### Layer 4: Defer work index refreshes

`setImmediate` for `refreshWorkIndex` in `tool_execution_end`, `agent_end`, and `session_before_compact` — plus an **in-flight guard**. Without one, a 30-edit turn spawns 30 concurrent `bun run soma lifecycle algorithm-updated` processes all writing `algorithm-work-index.json`; `spawnSync` previously serialised them. `somaFeedbackCaptureInFlight` in the same file is the pattern to copy.

### Layer 5 (NEW): `session-end` must be awaited

`captureSessionEnd` runs on `session_shutdown`. Fire-and-forget means the process exits and kills the child before it records the session-end lifecycle event — silent data loss at exactly the moment memory is captured. The handler is already `async`; it must `await` with a bounded timeout.

## Impact

| Scenario | Before | After |
|----------|--------|-------|
| First message of session | 2-6s freeze | no blocking Soma call |
| Subsequent messages | 2-6s freeze | no blocking Soma call |
| Classification availability | Always present (blocking) | Always present, non-blocking, current-prompt |
| Startup context availability | Always present (blocking) | Always present (cached at session_start) |

## Verification

- The generated extension must transpile — `test/substrate-adapters.test.ts:50` is the existing guard (regression #402). PR #476 shipped code that failed it.
- Projected classifier must be behaviourally identical to `classifyAlgorithmPrompt` across a prompt corpus.
- `session-end` must still run to completion on shutdown.

## Files to Change

1. `src/algorithm-classifier.ts` — expose the pattern contract (behaviour unchanged)
2. `src/adapters/shared/algorithm-classifier-source.ts` — new shared renderer
3. `src/adapters/pi-dev/adapter.ts` — consume it; async lifecycle; caching; guards

jcfischer:
## Problem

The Pi adapter calls `soma lifecycle session-start` and `soma algorithm classify` synchronously via `spawnSync` on every `before_agent_start` hook, causing 2-6 second freezes when the user enters a message.

Fixes #475

## Solution

**4-layer async optimization:**

1. **Eliminate redundant startup context calls** — `session_start` now caches `cachedStartupContext`; `before_agent_start` reuses it.

2. **Shift classification to `turn_end` (post-response)** — classify the prompt asynchronously after the response. Result cached for the NEXT turn. First message has no classification (fast); every subsequent message gets precomputed classification (no blocking).

3. **All Soma commands async** — `spawnSync` → `execFileAsync` everywhere: lifecycle, classification, startup context.

4. **Defer work index refreshes** — `setImmediate` on `tool_execution_end`, `agent_end`, `session_before_compact`.

## Expected Impact

| Scenario | Before | After |
|----------|--------|-------|
| First message | 2-6s freeze | ~100ms |
| Subsequent messages | 2-6s freeze | ~100ms |

**Trade-off:** First message of each session has no classification context. Every subsequent message gets the previous turn's classification without blocking.

## Changes

- `src/adapters/pi-dev/adapter.ts` (+74/-30 lines)
- Template array generates the optimized `soma.ts` extension code


jcfischer:
Closing: this branch was produced by a small model as a test run and is not salvageable.

The generated pi-dev extension does not parse (verified by emitting `agent/extensions/soma.ts` and transpiling it — `Unexpected export` at line 261), which is the exact failure class the #402 regression guard in `test/substrate-adapters.test.ts:50` exists to catch. Beyond the syntax errors, `result.status === 0` on a promisified `execFile` result is always false, so classification and startup-context refresh would both be dead even after fixing the parse.

Reopening #475 for a fresh implementation. Commit preserved at 947f9a9 if needed.

PR #489, opened 2026-08-01
fix(pi-dev): remove blocking Soma subprocesses from the message path

### LINKED ISSUE #475: [pi-dev] Soma blocking calls on every message cause 2-6s freeze

jcfischer:
## Problem

`before_agent_start` fired two blocking `spawnSync` calls on every message — a full session-start lifecycle plus `soma algorithm classify` — freezing Pi.dev before the response started.

Fixes #475. Supersedes #476 (closed).

## The key finding

**Classification never needed a subprocess.** `classifyAlgorithmPrompt` (`src/algorithm-classifier.ts`) is a pure, synchronous function — 143 lines of regex and string matching, zero I/O. The entire cost attributed to "classification" was bun cold-start, spawned to run a regex match.

That reframes the fix. #475 originally proposed moving classification to `turn_end` and injecting the cached result into the *next* turn — which is what #476 implemented, and it introduced a correctness bug: every message after the first carried the **previous message's** mode. Since the injected text is a directive (*"Do not downshift to native execution unless the principal explicitly overrides this classification"*), a wrong label is worse than no label.

Projecting the classifier removes the latency **and** the trade-off. Classification is correct for the current prompt, on every message including the first.

## Changes

**Classifier projection.** New `src/adapters/shared/algorithm-classifier-source.ts` renders the classifier into the generated extension. Following the `renderFeedbackHookHelper` idiom (`feedback-helper.ts:23` — *share the data, generate the logic*), the pattern set is serialised from a new `ALGORITHM_CLASSIFIER_CONTRACT` rather than retyped, so the projected copy cannot drift from the runtime one. `classifyAlgorithmPrompt`'s behaviour is unchanged — verified identical across the corpus before and after the contract refactor.

**Startup context** computed once at `session_start`, reused by `before_agent_start`.

**Lifecycle calls async** — with success detected correctly. `promisify(execFile)` resolves `{ stdout, stderr }`; there is no `status` field and the exit code lives on the rejection. `#476` used `result.status === 0`, which is *always false*, silently disabling both classification and startup-context refresh even after its syntax errors were fixed.

**Work-index refreshes deferred and coalesced.** `spawnSync` previously serialised these; an unguarded async fan-out would let a 30-edit turn race dozens of concurrent writers on `algorithm-work-index.json`. In-flight requests now collapse into one trailing run.

**`session_shutdown` awaits session-end capture** (bounded). Fire-and-forget loses the record entirely — shutdown kills the child mid-write.

## Measurements

Measured on this machine, warm, against a bootstrapped empty Soma home:

| Call | Cost |
|---|---|
| `soma algorithm classify` | ~210ms |
| `soma lifecycle session-start` | ~210ms |

Both are dominated by bun cold-start. `before_agent_start` paid both, synchronously, per message.

**Honest caveat:** I could not reproduce the 2–6s figure from #475 on an empty home — the floor is ~420ms. `session-start` reads all run files plus up to 8 learning and 8 relationship files and runs reprojection and self-repair, so it scales with home size; that is almost certainly where the multi-second reports come from. Worth confirming against your real `~/.soma`. After this change the message path spawns **nothing**, so the magnitude of the original freeze doesn't change the outcome.

## Verification

- `bun test` — **2079 pass, 0 fail**
- `bun run typecheck` — clean
- `bun run lint` — 51 errors, identical to `main`'s baseline (these files add none)
- Generated extension transpiles — the guard #476 failed

New tests:
- `test/pi-dev-classifier-projection.test.ts` — projected classifier vs runtime across a branch-covering corpus. **Mutation-tested**: injecting drift into the renderer makes it fail; reverting makes it pass.
- `test/pi-dev-blocking-calls.test.ts` — pins the message path as subprocess-free, plus the transpile check, the `result.status` trap, the coalescing guard, and awaited session-end. Also asserts `cachedClassification` / `turn_end` never come back.

Updated tests that pinned the removed subprocess (`current-work-pointer`, `home-projection`, `install`) — each now asserts the same intent under the new design, with a comment on why the per-prompt refresh is deliberately gone.

## Note

The `/soma` command and the `soma_context` tool both called `refreshStartupContext` without awaiting it once it became async — in #476 that rendered `[object Promise]` into the user-visible command output. Both are awaited here.


jcfischer:
## Confirmed on the real environment

Jens-Christian projected this to his live Pi home (`soma install pi-dev --apply`) and confirms the slowness is fixed.

This closes the open caveat in the PR description. I measured a ~420ms floor for the two blocking subprocesses against a bootstrapped *empty* Soma home and could not reproduce the 2-6s figure from #475 — the gap was `soma lifecycle session-start` scaling with home size (all run files, up to 8 learning + 8 relationship files, reprojection, self-repair). Against a populated real home the freeze was real, and removing both subprocesses from the message path resolves it.

Verified on disk after projection:
- `~/.pi/agent/extensions/soma.ts` — 21,582 bytes / 481 lines, transpiles clean
- sibling extensions (`soma-path-guard.ts`, `soma-algorithm.ts`) transpile clean
- no surviving `spawnSync` / `result.status` code paths (only explanatory comments)
- `~/.soma/profile/purpose.md` checksum unchanged — install did not clobber it

Issue #526, opened 2026-08-04
Implement the probe registry gate (DD-16 Amendment A)


jcfischer:
## Task

Implement the probe registry decided in **DD-16 Amendment A** (#524) and specified
in `docs/work-graph.md` §2.2. Until this lands, the runner shipped by #498
executes `command` probes ungated — which is the behaviour the amendment
rejects — so this ticket **inherits #524's blocking edge on #499**: the
ships-with-consumer merge stays gated until the gate exists.

### What to build

1. **Registry document** under the existing `soma policy` surface (§4 — no
   parallel policy registry), held in **soma-home**, scoped by repo identity:
   declared `command` entries as `run` + `cwd` pairs, and declared `url` hosts.
   Not repo-local — §1 clause 5 keeps enforcement off the tree it guards, and a
   committed registry is writable by any agent holding Write (#510).
2. **Enforcement in the probe runner** (`src/work-graph-probes.ts`):
   - `command` — refuse unless the exact `run` **and** `cwd` pair is declared.
     `cwd` is part of the match, not incidental.
   - `url` — refuse unless the target host is in the declared host set.
   - `git-ref-exists` / `git-merged-into` / `artifact-exists` — unchanged,
     ungated.
   A refused probe is a **failed** probe (`outcome: "fail"`, reason in
   `observed`), not an exception and not a skip — same fail-closed rule the
   runner already applies to unrunnable probes, so `assertClosable` refuses the
   close through the existing path.
3. **Uniformity** — identical for every autonomy class and for the phase-2
   headless tick. No registry on a machine ⇒ those closes refuse.
4. **Reading is not executing** — `soma graph node` and `soma graph frontier`
   must keep reading any node unchanged. Only the close path gates. A test
   should pin this, since it is the easy thing to break.
5. **Declaration surface** — how an adopter adds an entry (`soma policy …`
   verb or edited document). Follow the asymmetric-mutation rule §4 states:
   tightening additive, loosening identity-bound and fail-closed.
6. **Migration** — existing nodes with undeclared `command` probes become
   unclosable by machine until declared. The refusal message must name the
   exact undeclared `run`/`cwd` so the fix is copy-pasteable.

### Also update

`docs/work-graph.md` §2.2 currently marks the registry **specified, not yet
enforced**, pointing here. Clear that marker as part of closing this.

### Out of scope

Sandboxing / confinement of probe execution. DD-16 Amendment A rejected it *as
[...]

Issue #524, opened 2026-08-04
Probe execution trust boundary: command probes run tracker content as shell


jcfischer:
## Question

Surfaced by #498. `soma graph close` executes the node's declared `command`
probes as shell strings on the machine running the close. The probe declaration
lives in the issue body's node block — **tracker content** — so the question is
whose code the gate is willing to run.

### Why this is a §1 clause 1 tension, not a hardening nit

Clause 1: *a gate is deterministic iff it reads facts the agent cannot author.*
A `command` probe is agent-authored content that the gate then **executes**. The
P1 safeguard is that probes are *specified up front* and only later *probed* —
that defends against a session inventing a passing result after the fact, but it
does not defend against a session specifying `run: "curl … | sh"` at creation
and having the close path run it later under a different identity.

Concretely, once `ivy-agent` can create nodes (it can today, Triage):

1. the agent creates a node with an arbitrary `command` probe;
2. the principal runs `soma graph close` on it;
3. the probe executes with the principal's credentials and keychain reach.

That is agent-authored content escalating to execution in the principal's
session — the exact direction #511 is closing off from the other end. It is not
merely "the agent can already run commands in its own session": the *closer* may
be a different identity, and under phase-2 (§5) a scheduled tick closes nodes
headlessly, which turns any node body on the tracker into a scheduled execution.

Note also that #496 probe 7 established the agent **cannot** edit a
principal-authored issue body under Triage — but it can author its own nodes,
and #510 raises the role to Write.

### What this ticket decides

- Whether `command` probes stay unrestricted, get an allowlist, get a
  declared-at-install probe registry (named probes rather than free shell), or
  require ratification before first execution.
- Whether the answer differs by autonomy class, by who authored the node block,
  or by whether the closer is the node's author.
- Whether the phase-2 headless tick may run `command` probes at all.
- Whether execution should be confined (subprocess sandbox, no network, dropped
  env) rather than gated.

Read alongside: the non-`command` probe types (`url`, `git-ref-exists`,
`git-merged-into`, `artifact-exists`) already execute as argv with no shell and
carry no such exposure — an allowlist that leaves only those is a real option.

### Scope

This is likely a **spec change**, so per map #495's Notes it goes back through
`design/design-decisions.md` (DD-16) rather than being settled inside this map.
Wired as blocking [Orienteer re-base + ships-with-consumer merge to main](https://github.com/the-metafactory/soma/issues/499)
[...]

Issue #523, opened 2026-08-04
scratch failing probe


jcfischer:
Scratch node with a probe that fails.

Issue #514, opened 2026-08-04
scratch: soma graph verb exercise (#498) — safe to close


jcfischer:
Throwaway root for exercising the `soma graph` verbs as #498 close evidence. Not part of map #495. Closed automatically at the end of the run.

PR #528, opened 2026-08-04
feat(work-graph): the work graph primitive, its verbs, its trust gate, and orienteer

jcfischer:
The **ships-with-consumer merge** of DD-16's work graph: the primitive, its
verbs, its trust gate, and its first consumer arrive on `main` together — or
none of them do (#484 clause 4). Closes the work-graph map's #499.

Five closed nodes of map #495 ride in this branch, each already reviewed as a
decision on the tracker:

| Commit | Node | What lands |
| --- | --- | --- |
| `a573886` | #497 | `GraphStore` seam + GitHub backend — every rule in `src/work-graph.ts`, I/O only in `src/work-graph-github.ts` |
| `29e2715` | #498 | The five `soma graph` verbs, the probe runner, the four-conjunct attestation derivation |
| `22938bf` | #524 | DD-16 **Amendment A** — tracker content may parameterise a probe, never introduce code |
| `4b03f1e` | #526 | The probe registry gate that enforces Amendment A |
| `7c09176` | #499 | **orienteer** — the wayfinder doctrine re-based onto the verbs |

## What this is

A typed primitive for cross-session effort topology: nodes of work joined by
blocking edges, walked by agent sessions, closed only through checkpoint gates.
Spec: `docs/work-graph.md`. Stance and rationale: DD-16 in
`design/design-decisions.md`.

Three properties are the point:

- **A gate is deterministic iff it reads facts the agent cannot author.**
  `assertClosable` refuses a hollow close: no attached checkpoint, a declared
  probe that never ran, a probe that ran and failed, or no externally checkable
  evidence pointer.
- **Evidence is typed by lifecycle, not by assertion.** A probe is `specified`
  at creation and flips to `probed` only when the runtime actually ran it.
  Timeout expiry is failure; an unrunnable probe is failure; a probe the
  registry refuses is failure. An unrun probe may never read as a passed one.
- **`attestation` is derived, never configured.** Four conjuncts at close time,
  and it is a *label rather than a gate* — refusing on `unverified` would
  deadlock a bootstrap whose credential-separation nodes are themselves
  `approve`-class. Every receipt on this deployment reads `unverified`, honestly,
  until #510/#511 land.

## The trust gate (#524 → #526)

`soma graph close` executes `command` probes that live in tracker content. Since
`soma graph` runs against whatever repo an adopter points it at, soma cannot
know that repo's visibility or collaborator set — so no rule of the form "trust
authors who are X" is available, and only a content-structural rule survives:

- `command` — refused unless the exact `run` **and** the resolved absolute `cwd`
  are declared in the adopter's soma-home registry for that repo.
- `url` — refused unless the target host is declared (ungated, it is a blind
  SSRF oracle whose result the receipt publishes to a world-readable tracker).
- The three argv probe types stay ungated.

Deny by default. `soma policy probes` reads the registry; **no verb writes it**,
because adding an entry is a loosening mutation on the gate itself.

## Verification

- `bun test` — 2208 pass / 0 fail. `bunx tsc --noEmit` — clean.
- All five verbs walked against live scratch nodes #514–#523 during #498.
- The registry gate walked live in #526: it refused that node's own probes until
  the registry declared them, then closed it with a real 2/2 receipt.
- `soma init --apply --substrate claude-code` into a temp home projects all six
  orienteer files byte-identically and catalogs the skill.
- Diff is 6311 insertions / 10 deletions. Every deletion is a line replaced in
  place, not removed work — but "all docs and comments" would be wrong, so, named:
  2 prose lines reflowed in `docs/work-graph.md`; 2 TypeScript lines in
  `src/cli/policy.ts` (the `ParsedPolicyArgs` union and the `usage: […].join()`
  line, both widened for the new `probes` subcommand); 4 comment lines and **2
  changed assertions** in `test/install.test.ts` (projected-file counts 29→35 and
  21→27, from the six orienteer files). Nothing pre-existing is reverted.

## Known limits, recorded rather than hidden

- Enforcement runs from the **installed binary**, not the dev tree (§1 clause 5).
  A dev-tree close warns on stderr rather than refusing — refusing would make the
  primitive undevelopable.
- Conjunct 2 of the attestation derivation is a **detector, not a defence**: it
  runs inside the environment it judges. `verified` rests on a trusted base the
  runtime cannot check (#511 plus an unshimmed binary).
- The probe registry rests on the same base. That is #511's invariant, not this
  branch's.
- Phase 2 (headless tick, close auditor) is spec'd at §5 and deliberately not
  built.
- **Ratification binds to a comment id, not to its text.** Nothing hashes the
  proposal body, so a proposal ratified and *then edited* still closes on that
  👍. "Materially amended proposals need fresh ratification" is doctrine the
  operator holds, not a property the runtime checks.
- **Any non-proposer's 👍 closes a HITL node.** The root author is preferred, but
  the fallback is the first other reaction; only `attestation` records that it
  was not the right human.
- The ungated argv probes are **not** bounded to the local tree — `repo` is
  tracker content resolved without containment (→ #529). The read path is
  strictly serial over `gh` subprocesses (→ #530).

### DIFF
--- a/design/design-decisions.md
+++ b/design/design-decisions.md
@@ -385,6 +385,131 @@ skill (#487). Full spec: `docs/work-graph.md`.
 [#477](https://github.com/the-metafactory/soma/issues/477), decision tickets
 #483–#487, #491, #492; grilled with the principal 2026-08-02.
 
+#### DD-16 Amendment A: tracker content may parameterise a probe, never introduce code
+
+**Status:** Decided (2026-08-04), amends DD-16.
+
+**Context:** #498 shipped the `soma graph` verbs and the probe runner, and in
+doing so made a latent tension operational. A `command` probe is a shell string
+that lives in the node block — **tracker content** — and `soma graph close`
+executes it on the closing machine. Clause 1 of the determinism dividing line
+says a gate is deterministic iff it reads facts the agent cannot author; a
+`command` probe is agent-authored content that the gate then *runs*. The P1
+safeguard (evidence is `specified` up front and only later `probed`) defends
+against a session inventing a passing result after the fact. It does not defend
+against a session — or anyone else with issue-write — specifying
+`run: "curl … | sh"` at creation and having the close path execute it later,
+under a different identity.
+
+The decisive framing came from the principal: **the tracker is a parameter, not
+a known store.** `soma graph` takes `--repo` / `SOMA_GRAPH_REPO` and, like the
+wayfinder skill it descends from, runs against whatever repo the adopter points
+it at. Soma therefore cannot know that repo's visibility, collaborator set, or
+issue-creation policy. (The soma repo is itself public with two forks, so any
+GitHub account can author a node block today — but the argument does not rest on
+that, and must not: a rule justified by *this* repo's collaborator list is a rule
+that breaks on the next adopter's.) Only two rule shapes survive that
+constraint: **self-relative** rules, which need no knowledge of who is trusted,
+and **content-structural** rules, which deny the capability outright.
+
+Candidates:
+- **(a) Unrestricted**, compensated by confinement (sandbox execution).
+- **(b) Pattern allowlist** in runtime policy.
+- **(c) Named probe registry** — the node declares a name; the name→command
+  mapping lives locally.
+- **(d) Hash-bound ratification**, mirroring DD-7's inbound-content model.
+- **(e) Drop the `command` variant** entirely.
+
+**Decision:** **(c), in its literal-match form.** Tracker content may
+parameterise a probe but may never introduce executable code or a network
+destination.
+
+- **`command`** — the probe keeps its literal `run` string in the node block,
+  and the runner refuses to execute it unless that exact `run` **and `cwd`**
+  pair is declared in the adopter's local probe registry for this repo.
+  Deny-by-default exact match. `cwd` is part of the match and not incidental: a
+  declared `bun test` executed in an attacker-chosen directory is a different
+  command.
+- **`url`** — the target host must appear in the declared host set. Without
+  this, a `url` probe is a blind SSRF oracle: it issues a request from the
+  closing machine to a tracker-supplied address, and the receipt publishes the
+  observed status back to the tracker, which may be world-readable.
+- **`git-ref-exists`, `git-merged-into`, `artifact-exists`** — ungated. They
+  execute as argv with no shell, cause no egress, and their effect is bounded to
+  existence checks in a local tree.
+
+The registry lives in **soma-home only, scoped by repo identity**, under the
+existing `soma policy` inspection surface — §4 already forbids a parallel policy
+registry. Home rather than repo-local because clause 5 puts enforcement out of
+the agent's reach and off the tree it guards: a committed `.soma/probes.json`
+sits inside the guarded tree and is writable by any agent holding Write (#510).
+
+The rule is **uniform** across autonomy classes and applies equally to the
+phase-2 headless tick. The registry already answers *whose code is this* — the
+adopter's — and headlessness changes who is watching, not what is authorised. A
+machine with no declaration refuses those closes; fail-closed. Reading is
+unaffected: `soma graph node` and `frontier` still read any node, because a node
+is data. Only execution gates.
+
+**Rejected:**
+- (a) — confinement is blast-radius management, not authorisation, and answers
+  no part of "whose code is this". macOS offers no cheap reliable sandbox
+  primitive, and a probe that cannot reach the network or the real tree usually
+  cannot prove what it was written to prove. Retained as possible
+  defence-in-depth, not as the gate.
+- (b) — `docs/runtime-command-inspection.md` states Soma v0 "keeps explicit
+  allow as the absence of findings" and defers bypass-style trust lists; an
+  allowlist over arbitrary shell strings also requires the shell parsing that
+  same doc explicitly disclaims. Exact-match on a literal the adopter wrote is
+  not a pattern allowlist and does not reopen this.
+- (d) — maximum consistency with DD-7, and probes would stay self-describing,
+  but it puts a human decision in front of every distinct new command. That
+  places a human in the AFK path, which is the one thing the `auto` class exists
+  to avoid. Exact-match recovers DD-7's *exact-bytes* property anyway — editing
+  the probe on the tracker breaks the match — without needing a scanner or a
+  ratification step.
+- (e) — safest, and nothing would run a shell. Rejected because no remaining
+  probe type can express "the test suite passed", which is the most useful close
+  evidence there is; the `auto` class would lose its principal evidence source.
+- **Name-only registry** (the form (c) was first proposed in) — cleanest
+  separation, since tracker content could not even *look* like code. Rejected
+  because the node would no longer show what closing it executes: a human
+  reading the issue would have to consult a machine-local file, and only the
+  receipt, written afterwards, would record the resolved command. The literal
+  form keeps the tracker self-describing and gets exact-bytes semantics as a
+  side effect.
+
+**Implications:**
+- `docs/work-graph.md` §2.2, §4 and §5 updated (this commit). Spec status for
+  the probe runner moves from "shipped unrestricted" to "registry-gated".
+- Implementation is **not** in this amendment. #524 records the decision; the
+  implementation ticket inherits the blocking edge on #499, so the
+  ships-with-consumer merge stays gated until the gate exists rather than
+  shipping the behaviour this amendment rejects.
+- Existing nodes carrying undeclared `command` probes become unclosable by
+  machine until their command is declared. This is the intended migration: the
+  refusal is specific and actionable, and no node silently changes meaning.
+
+**Residual limits — recorded rather than solved:**
+- An attacker who can write a node body can still choose *which* of the
+  adopter's declared commands runs, and can set `timeoutSec` and `expectExit`
+  freely. Exact-match bounds the blast radius to commands the adopter authored;
+  it gives no control over which one fires, or when.
+- The whole construction rests on the registry being unreachable from the agent.
+  That is #511's invariant, not this amendment's, and the same honest limit
+  §3.2 already records for conjunct 2 applies here: a check that runs inside the
+  environment it judges can be shimmed by a session that sets out to do so.
+- Argument order matters for the ungated types: they are ungated because they
+  are argv and read-only *today*. Adding a probe type that writes, or that
+  passes tracker content to a shell, would reopen this decision rather than
+  inherit it.
+
+**Discussion:** Ticket
+[#524](https://github.com/the-metafactory/soma/issues/524) on map
+[#495](https://github.com/the-metafactory/soma/issues/495); grilled with the
+principal 2026-08-04. Related: DD-7 (inbound content untrusted until a
+hash-bound decision exists), `docs/runtime-command-inspection.md`.
+
 ## 6. Policy & Security
 
 ### DD-7: Soma owns inbound-content security; scanners provide evidence
--- a/docs/soma-home-layout.md
+++ b/docs/soma-home-layout.md
@@ -25,9 +25,16 @@ re-projection rewrites.
 ├── isa/                    # Verification State Artifacts, one <slug>.md per project/task
 │   └── .templates/         # VSA scaffolding templates
 ├── policy/                 # substrate policy declarations
+│   └── probe-registry.json # work-graph probe authorisations (see docs/work-graph.md §2.2)
 ├── imports/                # migration manifests and portability reports
 └── projections/            # cached generated projections (codex, claude-code, …)
 ```
+
+`policy/probe-registry.json` is **not** created by `soma init`: it authorises
+`command` and `url` probes to run on this machine, and its absence is the
+fail-closed default — those probes refuse until an adopter writes it by hand. See
+[`work-graph.md` §2.2](work-graph.md) for the format, or run `soma policy probes`
+to see the current state.
 
 On a fresh machine the profile files start as a **starter profile**
 (`status: starter-profile` in `principal.md`). Replace them with your own
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -112,6 +112,20 @@ evidence is typed `specified` at node creation and flips to `probed` only
 when the runtime has executed the probe and recorded its output
 (`ProbeResult` above).
 
+Runner semantics settled while implementing #498:
+
+- `repo` on the git and artifact probes is a **local working-tree path**
+  (relative to the runner's cwd), not `owner/name` — `artifact-exists`'s
+  `path` + `atRef` pair resolves as `git cat-file -e <atRef>:<path>`, which
+  needs a checkout rather than an API.
+- **A probe that cannot run is a failed probe, never a skipped one.** Runner
+  errors (network refused, git absent, bad path) record `outcome: "fail"` with
+  the reason in `observed`, so the close refuses. The one thing a runner may
+  not do is let an unrun probe read as a passed one.
+- Git probes execute as argv, never through a shell, so a ref name cannot
+  inject. `command` probes are shell strings by definition — see the trust
+  question that opens on #524.
+
 ### 2.3 Edge
 
 Blocking only: `blocks(a, b)` means `b` is not frontier until `a` is closed.
@@ -193,7 +207,25 @@ soma graph close <node>            # runs declared probes; refuses a hollow clos
 
 `close` enforcement lives in the **installed** soma binary, never the dev tree
 (#483 clause 5). Bypass via raw `gh` remains visible-but-unprevented in
-phase 1; the phase-2 auditor (§5) makes it detected.
+phase 1; the phase-2 auditor (§5) makes it detected. A close run from a dev
+tree warns on stderr rather than refusing — refusing would make the primitive
+undevelopable, and the warning keeps the gap visible state rather than silent.
+
+HITL closes are two-phase, inside the same verb rather than a sixth one:
+`close --propose` posts the proposal comment and stops; `close
+--proposal-comment <id>` reads its reactions and derives the receipt. Two seam
+addenda fell out of implementing it (#498), both the same class as the
+`CreateNodeSpec` addendum on #497:
+
+- `GraphStore.readComment` — conjunct 3 needs the *proposal's* author from the
+  API, and the two phases are separate process invocations, so the author must
+  come back from the backend rather than ride on the command line where it
+  would be caller-authored.
+- The parent edge is read over **GraphQL**. `GET /repos/{repo}/issues/{n}`
+  carries no `parent` key (only the child direction, `/sub_issues`, is in
+  REST), so a REST-only read resolves every node as its own root — which
+  silently degrades conjunct 4 from "the graph root's author may ratify" into
+  "a ticket's own author may ratify its close".
 
 ### 2.7 planSteps bridge
 
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -197,8 +197,16 @@ declare. **Reading is not executing:** `soma graph node` and
 `soma graph frontier` read any node regardless, because a node is data. Only the
 close path gates.
 
-Exact match also yields DD-7's *exact-bytes* property without a scanner: editing
-a probe on the tracker breaks its match and the close refuses.
+Exact match yields DD-7's *exact-bytes* property for the two fields it covers:
+editing a probe's `run` or `cwd` on the tracker breaks its match and the close
+refuses, with no scanner involved. It does **not** extend to the rest of the
+probe. `expectExit` and `timeoutSec` remain tracker content the gate never reads
+— which is the residual DD-16 Amendment A already records ("an attacker who can
+write a node body … can set `timeoutSec` and `expectExit` freely"). Concretely:
+flipping a declared `bun test` from `expectExit: 0` to `expectExit: 1` makes a
+*failing* suite record as a passed probe. The registry bounds **whose code
+runs**, not what counts as success; widening it to the whole probe would be a new
+decision, not an implementation detail of this one.
 
 **Migration:** existing nodes carrying undeclared `command` probes are
 unclosable by machine until their command is declared. Intended — the refusal is
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -127,10 +127,10 @@ Runner semantics settled while implementing #498:
 
 #### Probe registry (DD-16 Amendment A)
 
-**Status: specified, not yet enforced.** The runner shipped by #498 executes
-`command` and `url` probes ungated. The gate below lands with
-[#526](https://github.com/the-metafactory/soma/issues/526), which blocks the
-ships-with-consumer merge (#499) so this section cannot reach `main` unenforced.
+**Status: enforced** ([#526](https://github.com/the-metafactory/soma/issues/526)),
+in `src/work-graph-probe-registry.ts` — the gate — and `src/work-graph-probes.ts`,
+which consults it before dispatch so a probe type added later cannot inherit
+"ungated" by omission.
 
 Probe declarations are **tracker content**, so they may parameterise a probe but
 may never introduce executable code or a network destination. The tracker is a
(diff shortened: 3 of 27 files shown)


jcfischer (commented):

17 finding(s): 2 blocker, 5 important, 6 suggestion, 4 nit; 2 lens(es) failed to run: Architecture, ContextDrift.


### CodeQuality

- **[blocker]** `src/cli/graph.ts:706` — **Hand-written --evidence satisfies the close gate**
  `const evidence: CloseEvidence[] = [...parsed.options.evidence];` feeds caller-supplied entries straight into the receipt, and `assertClosable` only checks `kind` + non-empty `pointer` — so the closing session declares its own agent-external evidence, contradicting this module's own contract ("`--evidence` adds entries; it cannot substitute for either"). The 👎 test at test/cli-graph.test.ts:455 only refuses because no `approved` entry exists.
  Fix: Reject `--evidence` entries whose kind is in `agentExternalEvidenceKinds(node.autonomy)`, or have `assertClosable` count only derived entries (probe-derived `probed`, ratification-derived `approved`).

- **[important]** `src/cli/graph.ts:660` — **--dry-run is ignored on the --propose path**
  `if (parsed.options.propose === true)` returns after `await graph.postComment(ref, body)`, before the `dryRun` branch, so `close <id> --propose --body … --dry-run` posts a public tracker comment while references/closing.md documents `--dry-run` as "preview the verdict and receipt, write nothing".
  Fix: In `parseCloseArgs`, reject `--propose` together with `--dry-run`, or short-circuit the propose branch to render the comment body without posting.

- **[suggestion]** `src/cli/graph.ts:736` — **Missing proposal author silently satisfies conjunct 3**
  `author: proposalRef.author ?? ""` turns "the backend reported no author" into an empty string, so `deriveAttestation`'s `proposer === undefined` and `proposer === ratifier` checks both pass vacuously — the opposite of the fail-safe downgrade the surrounding comments promise.
  Fix: Leave `author` undefined when the API returns none, and let the "no proposal comment recorded" reason fire.

- **[suggestion]** `src/work-graph-github.ts:204` — **Node block round-trip breaks on "-->" in the payload**
  `decodeNodeBlock` ends the block at the first `body.indexOf(NODE_BLOCK_CLOSE, …)`, but `encodeNodeBlock` writes raw `JSON.stringify`, which does not escape `-->`; a probe like `{"run":"echo --> x"}` yields truncated JSON, so the node soma just created reads back as `typed: false` / approve-class forever.
  Fix: Escape or base64 the payload in `encodeNodeBlock` (or reject `-->` in `parseProbe`) so encode/decode round-trips.

- **[nit]** `src/cli/graph.ts:74` — **close accepts --json but never renders JSON**
  `parseCloseArgs` routes through `readShared`, which sets `options.json = true`, yet `runClose` never reads it and the close usage string omits `[--json]` — the flag is accepted and silently dropped.
  Fix: Either honour `--json` in `runClose` or reject it in `parseCloseArgs`.

- **[suggestion]** `src/skills/orienteer/references/map.md:88` — **Skill asserts a policy-floor clamp the runtime does not implement**
  "It is declared over a projected policy floor and clamped never-below-floor at creation" is shipped as enforced runtime behaviour, but no `src/work-graph*.ts` path reads a floor — `parseNodeSpec` only validates the literal `autonomy` value.
  Fix: Mark the clamp as not-yet-implemented (phase 2) in the skill, as the other known limits are.

  claude extraction recovered JSON but it is not lens-shaped (no `summary` or `findings` field) — extractor=raw

### ContextDrift — DID NOT RUN

- **[important]** `(lens output):0` — **ContextDrift: model deviated from JSON contract**
  claude extraction recovered JSON but it is not lens-shaped (no `summary` or `findings` field) — extractor=raw

### EcosystemCompliance

- **[suggestion]** `src/cli/policy.ts:283` — **`soma policy probes` accepts two competing home overrides**
  The parser takes both `--home-dir` and `--soma-home` and passes both through to `loadProbeRegistry`, but the documented override surface is one flag: "Concretely `~/.soma/policy/probe-registry.json` (override with `soma graph close --soma-home`)". Two overrides on an authorisation-list reader make it ambiguous which path an adopter is actually inspecting.
  Fix: Document the precedence of `--home-dir` vs `--soma-home` in §2.2, or drop `--home-dir` from `policy probes` to match the single documented override.

- **[nit]** `docs/soma-home-layout.md:25` — **Layout tree entry lacks the fail-closed marker used in prose**
  The tree line `│   └── probe-registry.json # work-graph probe authorisations` reads as a standard soma-home artifact, while the prose two lines later states it is "**not** created by `soma init`". A reader scanning only the tree would expect `soma init` to create it.
  Fix: Annotate the tree entry, e.g. `# work-graph probe authorisations (hand-written, not created by init)`.

### Maintainability

- **[suggestion]** `src/cli/graph.ts:63` — **EVIDENCE_KINDS hand-lists a union it cannot stay in sync with**
  `const EVIDENCE_KINDS: readonly WorkGraphEvidenceKind[] = ["specified", "probed", "tested", "judged", "approved"]` duplicates the `WorkGraphEvidenceKind` union from another module with no exhaustiveness check, while two lines above the same file derives its type from the tuple (`const GRAPH_ACTIONS = [...] as const; type GraphAction = (typeof GRAPH_ACTIONS)[number]`). Adding a sixth evidence kind compiles clean and silently drops it from CLI validation.
  Fix: Export the `as const` tuple from `work-graph.ts` and derive `WorkGraphEvidenceKind` from it; import the tuple here instead of re-listing.

- **[nit]** `src/cli/policy.ts:82` — **Policy usage string re-lists every subcommand constant**
  `usage: [POLICY_CHECK_USAGE, POLICY_SCAN_USAGE, POLICY_PROMOTE_USAGE, POLICY_INSPECT_USAGE, POLICY_GUARD_USAGE, POLICY_PROBES_USAGE].join("\n")` restates the same six constants that the adjacent `subcommands` record already holds, so each new subcommand needs two edits and only one is type-checked.
  Fix: Build `subcommands` first, then `usage: Object.values(subcommands).join("\n")`.

- **[nit]** `docs/work-graph.md:140` — **Amendment A's normative rules stated twice, near-verbatim**
  `docs/work-graph.md` §2.2 and `design/design-decisions.md` DD-16 Amendment A carry the same rules in near-identical prose — "`cwd` is part of the match: a declared command run in an attacker-chosen directory is a different command" vs "`cwd` is part of the match and not incidental: a declared `bun test` executed in an attacker-chosen directory is a different command" — so a future gate change has to land in both.
  Fix: Keep the §2.2 table normative and have the DD state rationale plus a link, rather than restating the rules.

### HonestOracle

- **[blocker]** `src/cli/graph.ts:685` — **close --soma-home lets the caller supply its own probe registry**
  The description claims "Deny by default. `soma policy probes` reads the registry; **no verb writes it**, because adding an entry is a loosening mutation on the gate itself", and DD-16 Amendment A puts it in soma-home precisely because a repo-local file "is writable by any agent holding Write (#510)" — but close parses `--soma-home <dir>` and hands it to `deps.loadProbeRegistry(repo, parsed.options.somaHome)`, and `probeRegistryPath` resolves `options.somaHome ?? join(home, ".soma")` with no further check. The threat actor named in the amendment needs no new capability: write a registry anywhere, pass `--soma-home`, and the exact `run`/`cwd` match authorises itself.

- **[important]** `test/install.test.ts:151` — **"every deletion is a doc marker or test comment" is false**
  The description states "6311 insertions, 10 deletions, and every deletion is a replaced doc marker or an updated test comment"; of the 10, two are TypeScript source lines in `src/cli/policy.ts` (the `ParsedPolicyArgs` union and the `usage: [...].join("\n")` line) and two are changed assertions (`expect(result.substrateHome.files).toHaveLength(29)` / `(21)` → 35 / 27), not comments. A reviewer taking "diff is additive" at face value skips the only pre-existing behaviour the branch modifies.

- **[important]** `docs/work-graph.md:145` — **Ungated git probes are not bounded to "a local tree"**
  The gate table justifies ungating with "Ungated — argv, no shell, no egress, bounded to existence checks in a local tree", but `probeCwd` is `resolve(fallbackCwd, probeRepo)` with no containment check (src/work-graph-probes.ts:180) and `repo` is tracker content, so `repo: "../../.."` or an absolute path probes any directory on the closing machine. The result string (`` `${probe.ref} does not resolve in ${cwd}` ``) lands in the receipt on the same possibly world-readable tracker whose disclosure risk is the stated reason `url` probes are gated — a counter-case the amendment never engages.

- **[suggestion]** `docs/work-graph.md:130` — **"Status: enforced" omits the dev-tree warn-only path**
  §2.2 reads "**Status: enforced**" and "A machine with no declaration refuses those closes; fail-closed", while the branch's own close path only warns for dev-tree runs — "a close gated by the tree it guards is not gated" (src/cli/graph.ts:679). The qualifier appears in the PR body's known limits but not in the spec section a future reader will consult for the gate's status.


---

jcfischer (commented):

17 finding(s): 6 important, 8 suggestion, 3 nit; 2 lens(es) failed to run: Architecture, ContextDrift.


### CodeQuality

- **[suggestion]** `src/cli/policy.ts:322` — **`soma policy probes` resolves repo via graph CLI, creating a cycle**
  `src/cli/policy.ts` imports `resolveGraphRepo` from `./graph`, while `src/cli/graph.ts` is the far heavier module; the fallback `parsed.options.repo ?? (await resolveGraphRepo())` will throw whatever `resolveGraphRepo` throws when no repo is configured, surfacing a graph-flavoured error from a `policy` subcommand.
  Fix: Catch the resolution failure and rethrow as a `policy probes`-scoped message naming `--repo`/`SOMA_GRAPH_REPO`.

- **[nit]** `src/cli/graph.ts:63` — **`close` usage string omits `--json` and `--soma-home`**
  The `close` usage line lists `[--dry-run] [--repo <owner/name>]` but not `--json`, which an earlier round confirmed the parser accepts; help text that disagrees with the parser misleads callers.
  Fix: Either drop `--json` from the close parser or add it to the usage string.

  claude extraction recovered JSON but it is not lens-shaped (no `summary` or `findings` field) — extractor=raw

### ContextDrift — DID NOT RUN

- **[important]** `(lens output):0` — **ContextDrift: model deviated from JSON contract**
  claude extraction recovered JSON but it is not lens-shaped (no `summary` or `findings` field) — extractor=raw

### EcosystemCompliance

- **[suggestion]** `src/cli/policy.ts:289` — **Inspection verb can read a registry the gate never consults**
  Spec: "`soma policy probes [--repo <owner/name>]` shows what is declared and where" — the documented signature carries no home override, yet the parser adds `--soma-home` and `--home-dir` cases. An adopter verifying a declaration via a pointed home sees a registry the close path (which "takes no flag that moves it") will never read.
  Fix: Drop `--soma-home`/`--home-dir` from `POLICY_PROBES_USAGE` and the parser, or have the output name the resolved registry path so an override is visible in what it prints.

- **[nit]** `docs/soma-home-layout.md:25` — **Layout tree cites §2.2 for a format documented elsewhere in the doc**
  The tree comment says "see docs/work-graph.md §2.2" while the prose below links `[work-graph.md §2.2](work-graph.md)` without an anchor. Two pointers to the same section, neither resolving to it.
  Fix: Use a single anchored link (`work-graph.md#22-evidence`-style) in the prose and drop the section reference from the tree comment.

### Performance

- **[important]** `src/work-graph.ts:723` — **frontier confirms candidates strictly serially**
  `for (const candidate of candidates) { … const state = await this.store.readNode(candidate); }` awaits each confirmation before starting the next, and each `readNode` is three `Bun.spawn(["gh", …])` round-trips — a 30-child map costs ~90 strictly serial CLI invocations (tens of seconds) on the verb orienteer's WalkTheMap runs first every session. The candidates are independent reads with no shared state.
  Fix: Fan out with bounded concurrency (`Promise.all` over chunks of ~8) and keep candidate order when assembling `confirmed`.

- **[important]** `src/work-graph-github.ts:286` — **readNode pays 3 serial round-trips, one always wasted**
  `const parentNumber = issue.parentNumber ?? (await this.fetchParentNumber(issue.number));` always fires the GraphQL call, because this file documents that `GET repos/{repo}/issues/{n}` "has no `parent` key at all" — so `issue.parentNumber` is never set. `frontier` and `claim` discard `parent` entirely, so that spawn is pure waste, and the `blocked_by` fetch that follows is serialised behind it despite being independent of it.
  Fix: Issue `fetchIssue` and the `blocked_by` fetch via `Promise.all`, and make the parent lookup opt-in (a flag on `readNode`) so only the root walk pays for it.

- **[suggestion]** `src/cli/graph.ts:751` — **close re-fetches the node it already read**
  `runClose` opens with `const state = await graph.readNode(ref);` and later calls `findGraphRoot(ref, async (nodeRef) => await graph.readNode(nodeRef))`, whose first hop reads the same `ref` again — three redundant `gh` spawns per close.
  Fix: Start the walk from `state.parent` (returning `state` as root when it is undefined), or pass the already-read state as the walk's seed.

- **[suggestion]** `src/cli/graph.ts:604` — **each --blocked-by edge re-walks the graph from scratch**
  `for (const blockerId of parsed.options.blockedBy) { await graph.addBlockingEdge(…) }` and every `addBlockingEdge` runs `if (await this.reaches(blocker, blocked.id))`, a serial BFS of `readNode` calls with no cache carried across edges; `soma graph add` with 5 blockers re-reads overlapping ancestors five times at three spawns per node.
  Fix: Thread a shared per-invocation `NodeState` cache through `reaches`, or validate all requested edges in one walk before writing any.

### Maintainability

- **[important]** `src/cli/graph.ts:658` — **runClose fuses three independent paths into 172 lines**
  `runClose` spans 658–829 and holds the propose branch, the usurped-evidence check, probe running and refusal, ratification, attestation, receipt build, the dry-run render and the real close. Its own comment records the harm: "`--propose` posts a public comment and returns before the dry-run branch is ever reached" — the earlier `--dry-run` bug existed because two unrelated paths share one body.
  Fix: Split `runClose` into `runProposePhase(parsed, graph, repo, deps)` (the 668–689 branch, which returns before any close machinery) and `resolveRatification(graph, ref, parsed, root)` (756–777); leave the receipt assembly in `runClose`.

- **[suggestion]** `src/work-graph.ts:487` — **toNode's two branches are byte-identical**
  `if (spec.autonomy === "auto") { const { body: _body, parent: _parent, ...rest } = spec; return { ...rest, id }; }` is followed by the same two lines verbatim, so the discrimination does nothing. Deleting the branch typechecks clean (`bunx tsc --noEmit` on the edited tree emits nothing).
  Fix: Drop the `if` and keep one destructure-and-return.

- **[suggestion]** `src/work-graph-probes.ts:199` — **Gated cwd and executed cwd are computed twice**
  `probeCwd(probe.cwd, baseCwd)` is evaluated once for the authorization call at line 199 and again for `cwd:` on the spawn at line 210; the gate's exact-match guarantee holds only while the two expressions stay identical.
  Fix: Hoist `const resolvedCwd = probe.type === "command" ? probeCwd(probe.cwd, baseCwd) : baseCwd;` above the `authorizeProbe` call and pass it to both.

- **[nit]** `src/cli/graph.ts:374` — **Dead claim branch returns the same object as the fallthrough**
  `if (action === "claim") return { command, action, target: resolvedTarget, options };` is immediately followed by an identical `return { command, action, target: resolvedTarget, options };`.
  Fix: Delete the `if`.

### HonestOracle

- **[important]** `src/skills/orienteer/references/closing.md:88` — **Re-ratification after amendment is asserted but never enforced**
  closing.md states "A materially amended proposal is re-posted and needs fresh ratification" as a runtime property, but the close path binds ratification to a comment *id* only (`proposal = { commentId, author }`, src/cli/graph.ts:769) and never hashes or re-reads the proposal body — post a proposal, collect the 👍, edit the comment to a materially different resolution, and `close --proposal-comment c1` still writes `approved` evidence naming the ratifier. The PR's "Known limits, recorded rather than hidden" section enumerates four gaps and omits this one, so a reader concludes the replay-rebind case is covered.

- **[suggestion]** `src/cli/graph.ts:654` — **Doctrine omits that any account's 👍 closes an approve-class node**
  `selectRatification` falls back to `approvals[0]` — any non-proposer reaction — so on a public tracker a stranger's 👍 produces `approved` evidence and the node closes, with only the `unverified` label recording it (asserted at test/cli-graph.test.ts:451: "a 👍 from someone other than the root author closes, but never as verified"). closing.md's HITL section says only "A 👎 from the graph root's author suppresses ratification outright", leaving an agent to read root-author approval as the closing condition.

- **[suggestion]** `src/cli/graph.ts:673` — **--propose writes to the tracker before any closability check**
  closing.md:7 promises "When it refuses, nothing reaches the tracker", but the propose phase calls `graph.postComment(ref, body)` with no checkpoint or closability precheck. An approve-class node created without `--checkpoint` gets its proposal published and only refuses in phase two, leaving a ratifiable proposal on a node that cannot close.


---

jcfischer (commented):

22 finding(s): 1 blocker, 6 important, 13 suggestion, 2 nit; 2 lens(es) failed to run: Architecture, ContextDrift.


### CodeQuality

- **[important]** `src/work-graph-github.ts:69` — **`gh api --paginate` output is not a single JSON document**
  `if (request.paginate === true) args.push("--paginate")` feeds `JSON.parse(trimmed)`, but `gh api --help` states "Each page is a separate JSON array or object. Pass `--slurp` to wrap all pages…" — so any response past page 1 is `[…][…]` and throws. Every paginated call (`/sub_issues`, `dependencies/blocked_by`, `/reactions`) is affected, and the mocked-transport tests cannot see it.
  Fix: Push `--slurp` alongside `--paginate` and flatten the outer array (`pages.flat()`) in the transport.

- **[important]** `src/work-graph-probes.ts:141` — **`url` probe has no timeout, so a hung target blocks the close**
  `const response = await fetch(target, { redirect: "manual" })` carries no `AbortSignal`, and the `url` variant is the only Probe with no `timeoutSec` while git probes get `GIT_TIMEOUT_SEC`. That contradicts this module's own rule — "**Timeout expiry is probe failure**, never a hang… a hanging command must not block the close path"; a declared host that blackholes packets left `fetch` unresolved past 40s in a local check.
  Fix: `fetch(target, { redirect: "manual", signal: AbortSignal.timeout(URL_TIMEOUT_SEC * 1000) })`, and let the existing catch record it as a failed probe.

- **[suggestion]** `src/work-graph-probes.ts:124` — **SIGKILL on the shell does not free pipes held by its children**
  `proc.kill("SIGKILL")` kills only `/bin/sh`, while `await Promise.all([new Response(proc.stdout).text(), …])` waits for EOF on stdout — an orphaned grandchild keeps that pipe open. A probe whose command backgrounds a process (`bun test` leaving a stray server) blocks the close well past `timeoutSec` even though `timedOut` is already true.
  Fix: Race the output reads against the timer (or read with a deadline) so the runner returns the timeout result without waiting for EOF.

- **[suggestion]** `src/cli/graph.ts:756` — **Missing evidence pointer silently drops the probed entry**
  `if (pointer !== undefined)` skips the `probed` evidence entirely when `defaultEvidencePointer` cannot run `git rev-parse HEAD`, and the close then fails in `assertClosable` with "needs at least one probed or tested evidence entry with an externally checkable pointer". The operator sees passing probes and a refusal that never mentions the real cause.
  Fix: Refuse (or warn) explicitly when probes passed but no pointer could be derived, naming the failed `git rev-parse HEAD`.

- **[suggestion]** `src/work-graph-attestation.ts:171` — **HITL-only conjuncts are evaluated on `auto` closes**
  `deriveAttestation` unconditionally pushes "no ratification found — nothing was attested by a second credential" and "no proposal comment recorded", but §3.2 scopes those conjuncts to `propose`/`approve`. Every `auto` receipt therefore records failures naming a proposal an `auto` node can never have.
  Fix: Pass the autonomy class in and evaluate conjuncts 3–4 only for HITL closes, or record "not applicable" instead of a failure reason.

- **[suggestion]** `src/cli/graph.ts:806` — **`close --checkpoint` can only produce a refusal**
  `checkpointId: parsed.options.checkpointId ?? state.node.checkpointId ?? ""` is the flag's only use, and `assertClosable` throws unless `receipt.checkpointId !== checkpointId` is false — so any value other than the node's own is refused and the matching value is a no-op. The usage string still advertises `[--checkpoint <id>]`.
  Fix: Drop `--checkpoint` from `close` (the `--propose` path already documents that no verb attaches one after creation).

- **[nit]** `src/cli/graph.ts:580` — **`--body` and `--body-file` together silently prefer the file**
  `if (bodyFile !== undefined) return await deps.readTextFile(bodyFile);` discards `--body` with no diagnostic, though the usage strings present them as alternatives (`[--body <text>|--body-file <path>]`). The adjacent `--propose`/`--dry-run` check shows contradictory combinations are otherwise refused.
  Fix: Throw when both are supplied, matching the `--propose` + `--dry-run` refusal.

  claude extraction recovered JSON but it is not lens-shaped (no `summary` or `findings` field) — extractor=raw

### ContextDrift — DID NOT RUN

- **[important]** `(lens output):0` — **ContextDrift: model deviated from JSON contract**
  claude extraction recovered JSON but it is not lens-shaped (no `summary` or `findings` field) — extractor=raw

### EcosystemCompliance

- **[suggestion]** `src/skills/orienteer/references/map.md:1` — **Skill directory mixes `Workflows/` and lowercase `references/`**
  The branch adds `src/skills/orienteer/Workflows/ChartTheMap.md` alongside `src/skills/orienteer/references/closing.md` — TitleCase for one subdirectory, lowercase for the other, inside one skill. Sibling skills in this ecosystem use a single casing convention for skill subdirectories, and the install projection test now hard-codes both paths.
  Fix: Pick one casing (`References/` to match `Workflows/`, or lowercase both) and update the projection list in `test/install.test.ts`.

- **[suggestion]** `docs/soma-home-layout.md:25` — **Layout doc points at work-graph.md §2.2 without an anchor**
  The tree comment says `# work-graph probe authorisations (see docs/work-graph.md §2.2)` and the prose links `[`work-graph.md` §2.2](work-graph.md)` — a bare file link with the section only in the label. Section-numbered cross-references elsewhere in the soma docs resolve to an anchor, so this one lands readers at the top of a 400+ line spec.
  Fix: Link the heading anchor, e.g. `(work-graph.md#probe-registry-dd-16-amendment-a)`.

- **[suggestion]** `src/cli/policy.ts:12` — **policy module imports from the graph CLI module**
  `import { resolveGraphRepo } from "./graph";` makes the `policy` command surface depend on the `graph` command surface; every other `src/cli/*` subcommand module in this file's import block pulls only from shared helpers (`./parse-utils`, `./substrate`) or `../` domain modules. Repo resolution is domain logic, not CLI-verb logic.
  Fix: Move `resolveGraphRepo` into `src/work-graph.ts` (or a shared `work-graph-repo.ts`) and import it from both CLI modules.

### Performance

- **[suggestion]** `src/cli/graph.ts:730` — **Close serializes four independent subprocess round-trips**
  `const identity = … await deps.resolveIdentity()`, `await deps.loadProbeRegistry(repo)`, `await findGraphRoot(ref, …)` and `const confinement = await deps.checkConfinement()` are awaited one at a time, yet none consumes another's result — identity spawns `gh api user`, confinement spawns up to three more processes, and the root walk pays a full `readNode` per ancestor. Every close and every `--dry-run` pays the sum of these latencies instead of the max.
  Fix: `const [identity, registry, confinement] = await Promise.all([...])`, and kick off `findGraphRoot` alongside them.

- **[suggestion]** `src/work-graph-attestation.ts:91` — **Confinement probes run serially despite being independent**
  `gh auth status`, `gh auth token` and `security find-generic-password` are three sequential `await deps.runCommand(...)` calls; only their aggregation into `reachable` is ordered, not their execution, and the first two are network-validating `gh` calls. Cost lands on every close and dry-run.
  Fix: Run the three with `Promise.all`, then aggregate `logins`/`tokenReachable`/keychain into `reachable` as today.

- **[suggestion]** `src/work-graph-probes.ts:124` — **Probe stdout buffered whole to keep a 1.2 KB tail**
  `new Response(proc.stdout).text()` materializes the entire output of a declared `command` probe as a JS string, while `boundObserved` then discards everything but the last `OBSERVED_TAIL_LIMIT = 1_200` characters. A verbose declared command (a full test or build run) spikes RSS by its whole log for a value that is thrown away.
  Fix: Stream the pipes and keep a rolling tail buffer of `OBSERVED_TAIL_LIMIT` chars instead of accumulating the full text.

### Maintainability

- **[suggestion]** `test/work-graph.test.ts:219` — **Two hand-copied FakeStore doubles across new test files**
  Both `test/cli-graph.test.ts:53` and `test/work-graph.test.ts:219` declare `class FakeStore implements GraphStore` with the same `readonly closed`/`created`/`edges` recorders and near-identical `createNode`/`addBlockingEdge`/`listCandidateFrontier`/`claim`/`close` bodies. `GraphStore` grew twice on this branch alone (`CreateNodeSpec` addendum, then `readComment`), and every such addendum now costs two edits.
  Fix: Extract one seeded `FakeGraphStore` into a shared test helper and let each suite override only the behaviour it exercises.

- **[suggestion]** `src/skills/orienteer/references/closing.md:124` — **Resolution-recording procedure restated near-verbatim in two skill files**
  closing.md's "Recording the resolution" repeats WalkTheMap.md steps 4–5 almost word for word — both say "Post the answer as a **resolution comment** on the node", "Append a one-line gist plus link to the map's **Decisions so far**", and "clearing each graduated patch from **Not yet specified** so it lives only as its new node". The skill's own invariant is "The map is an index, not a store. A decision lives in exactly one place."
  Fix: Keep the ordered procedure in WalkTheMap.md; in closing.md keep only what is close-specific (receipt vs resolution comment) and point back.

- **[nit]** `src/work-graph-github.ts:175` — **Free readComment shadowed by the GraphStore method of the same name**
  `function readComment(value: unknown, context: string): GitHubComment` sits in the same file as `async readComment(ref: CommentRef)`, whose body calls the free one — `const comment = readComment(await this.transport({...}), "readComment")` reads as self-recursion at a glance.
  Fix: Rename the free function to `parseComment` (its sibling `readIssue` has no such collision).

### HonestOracle

- **[blocker]** `docs/work-graph.md:200` — **Exact-match claim does not cover expectExit or timeoutSec**
  The diff states "Exact match also yields DD-7's *exact-bytes* property without a scanner: editing a probe on the tracker breaks its match and the close refuses", while `authorizeCommand` compares only `entry.run === run && entry.cwd === resolvedCwd` — `expectExit` and `timeoutSec` stay tracker content the gate never reads, and `parseNodeSpec` accepts any integer for `expectExit`.

- **[important]** `src/cli/graph.ts:446` — **"Externally checkable" pointer is a local HEAD sha**
  `evidencePointer` is documented in the diff as "An externally re-checkable anchor for probe evidence — the commit the probes ran against", and it is the sole thing satisfying `assertClosable`'s demand for "evidence entry with an externally checkable pointer" on an `auto` node; `defaultEvidencePointer` returns `HEAD ${sha}` from `git rev-parse HEAD` in the CLI's own cwd, which is neither pushed nor necessarily the tree any probe ran in.

- **[important]** `src/work-graph-probes.ts:12` — **url probes carry no timeout; "never a hang" is false**
  The module header claims "**Timeout expiry is probe failure**, never a hang and never a pass — a hanging command must not block the close path", and `closing.md:37` repeats it beside a `url` example, yet the `url` variant is `{ type: "url"; target: string; expectStatus: number }` with no `timeoutSec` and `defaultFetchStatus` calls `fetch(target, { redirect: "manual" })` with no AbortSignal.

- **[suggestion]** `(PR description):0` — **Verification block describes the branch two commits ago**
  The description states "Diff is 6311 insertions / 10 deletions" and "`bun test` — 2208 pass / 0 fail", but `git diff --shortstat main...HEAD` reports 6458 insertions and the head commit's own message reports 2212 pass — the evidence predates the two commits that changed the close gate (`--evidence` refusal, `--soma-home` removal).


---

Issue #542, opened 2026-08-06
Skill loading strategy belongs in the install spec, not skill frontmatter (skillsLoading)


jcfischer:
## Problem

Substrates disagree about how much of a skill they pull into model context, and
Soma has no way to say which behaviour a substrate has.

Claude Code loads skill bodies on demand — a session receives the registry plus a
name/description listing. Pi.dev loads bodies eagerly. Same `~/.soma/skills`,
same symlinks, same registry. Nothing about any skill differs. Only the loader
does.

Measured on `~/.soma/skills`, 2026-08-06:

| Quantity | Measured |
| --- | --- |
| Skills in the registry | 114 |
| Total `SKILL.md` lines | 18,644 |
| Total `SKILL.md` bytes | 799,571 |
| Estimated tokens (soma V0 estimator, `bytes / 4`) | ~200,000 |
| Mean skill size | 163 lines |
| Skills over 500 lines | 3 |

An eager substrate pays ~200K tokens before the first user turn. An on-demand
substrate pays the registry: 201 rendered lines, inside the existing
`SKILL_REGISTRY_LINE_BUDGET` of 300.

## What already exists

The registry tier is shipped and needs no change:

| Piece | Location |
| --- | --- |
| Registry projection | `renderSkills()`, `src/adapters/shared/index.ts:199` |
| Entry renderer | `renderSkillRegistryEntry()`, `src/adapters/shared/skill-registry.ts:173` |
| Line budget | `SKILL_REGISTRY_LINE_BUDGET = 300`, `skill-registry.ts:31` |
| Trigger / anti-trigger extraction | `extractUseWhenTriggers()`, `extractAntiTriggers()`, `skill-registry.ts:85,110` |
| Catalog auto-refresh | `refreshSkillCatalogs()`, `src/skill-projection.ts:316` |
| Per-substrate loader root | `skillsLoaderDir()`, `src/install-spec.ts:107`, 6 adapters |
| Body projection | `ensureSymlink()`, `src/skill-projection.ts:163` |

The router tier is not built: `SkillRoute`, `estimatedTokens`, and `defaultLoad`
each have 0 occurrences in `src/`.

## Proposal

Add one field to `SubstrateInstallSpec` (`src/install-spec.ts:79`), beside the
loader root it already owns (`:107`, soma#356):

```ts
interface SubstrateInstallSpec<S extends InstallSubstrate = InstallSubstrate> {
  skillsLoaderDir(substrateHome: string): string;
[...]

Issue #356, opened 2026-06-25
project-skill: move per-substrate loader projection behind an adapter/install-spec method


jcfischer:
Follow-up from sage's review of #355 (soma#354 slice 1).

`src/skill-projection.ts` currently hard-codes the per-substrate projection builders (`const projectionBuilders: Record<InstallSubstrate, ...>`) and derives each substrate's skill-loader root via `dirname(installSpecFor(substrate).vsaSkillProjection.destinationDir(substrateHome))`. Per `docs/architecture.md`, Skills core should only define discovery/routing contracts while each substrate adapter decides how to load/execute skills.

So adding or changing a substrate's skill projection currently means editing this cross-substrate module rather than the owning adapter/spec.

**Proposed:** add a `skillLoaderRoot(substrateHome)` (and, if needed, a catalog-file locator) to `SubstrateInstallSpec` / the adapter surface, and have `project-skill` / `unproject-skill` call that substrate-owned method instead of reaching into `vsaSkillProjection` + a hard-coded builder map.

Deferred from #355 to keep slice 1 scoped to the working primitive; the claude-code path is correct, this is a boundary cleanup.

PR #545, opened 2026-08-06
docs(skills): loading mode belongs to the install spec, not skill frontmatter (#542)

jcfischer:
Revises `docs/progressive-skill-loading.md` against what is actually built and
measured, and folds in a superseded Pi-side draft that proposed a parallel stub
format.

Docs only — no code change. The spec change it decides is tracked as #542.

## Why

The original spec was written pre-implementation. Since then the registry tier
shipped and the router tier did not, so the document no longer described the
system it specifies. Separately, a Pi-side draft was proposing to build a second
stub format alongside the registry Soma already projects.

## What changed

**Implementation status is now grounded.** New table with `file:line` for each
layer:

- Shipped: `renderSkills()`, `renderSkillRegistryEntry()`,
  `SKILL_REGISTRY_LINE_BUDGET = 300`, trigger/anti-trigger extraction,
  `refreshSkillCatalogs()`, `skillsLoaderDir()` across 6 adapters,
  `ensureSymlink()` body projection.
- Not built: `SkillRoute`, `estimatedTokens`, `defaultLoad` — 0 occurrences in
  `src/`.
- Typed but never materialized: `SomaSkillManifest` (`src/types.ts:958`), with
  **0** `soma-skill.json` files on disk.

**Six decisions recorded (DD-A…DD-F).** The load-bearing ones:

| | Decision |
| --- | --- |
| DD-A | Loading strategy is a substrate capability. `skillsLoading: "on-demand" \| "eager"` goes on `InstallSpec` beside `skillsLoaderDir`, not as a `scope` field in 114 portable skill files. |
| DD-B | An eager substrate needs *generated* stubs, because a symlink cannot be a partial view of its target. |
| DD-C | Promotion arrives as a tool result, never a system-prompt edit, or it invalidates the prompt cache. |
| DD-D | Derive routing metadata from frontmatter; defer `soma-skill.json`. |
| DD-E | Measure kernel context before gating it per skill. |
| DD-F | Routing accuracy is measured against a labelled query set, not asserted. |

**Migration plan reordered** so instrumentation lands before the router — it is
the baseline every later claim is measured against.

## Measurements behind the revision

Taken 2026-08-06 against `~/.soma/skills` and `~/.claude/rules/soma/`:

| Quantity | Measured |
| --- | --- |
| Skills in the registry | 114 |
| Total `SKILL.md` lines | 18,644 |
| Estimated tokens (soma V0 estimator) | ~200,000 |
| Mean skill size | 163 lines |
| Rendered registry | 201 lines (budget 300) |
| Whole always-on Soma projection | 427 lines |
| `PROFILE.md` lines byte-identical to `CONTEXT.md` | 41 of 51 |

The superseded draft assumed a 2,000-line mean and a 234,000-line corpus. The
problem is real and roughly 12x smaller than it claimed, which changes what the
fix is worth.

DD-E notes a separable, shippable-today cleanup that falls out of the last two
rows: collapsing the duplicated identity projections and dropping `README.md`
from the loaded `rules/` path removes ~21% of always-on context with no new
types. Not in this PR.

## Verification

- Every `file:line` reference in the document was re-read in the same session it
  was written.
- Every quantity in the tables is a measurement, not an estimate, except the two
  labelled as `bytes / 4` V0 token estimates.

## Related

- #542 — the `skillsLoading` spec change (DD-A, DD-B, DD-C).
- #356 — introduced `skillsLoaderDir` as the adapter-owned loader root; this
  extends the same seam.
- Superseded draft `~/work/pi/subagents/context-aware-skill-loading.md` now
  carries a pointer here.

### DIFF
--- a/docs/progressive-skill-loading.md
+++ b/docs/progressive-skill-loading.md
@@ -133,8 +133,14 @@ optional workspace overlays. It is safe to project into substrate homes because
 it is small and descriptive.
 
 The registry is **the** stub contract. Substrates consume it; they do not define
-their own. Its entry shape is owned by `renderSkillRegistryEntry()` and its size
-is bounded by `SKILL_REGISTRY_LINE_BUDGET`.
+their own. Its entry shape is owned by `renderSkillRegistryEntry()`.
+
+`SKILL_REGISTRY_LINE_BUDGET` (300) is a **design target, not a runtime cap**.
+`renderSkills()` emits every skill, so the real catalog can cross it as skills
+accumulate; the constant is asserted in `test/skill-registry.test.ts` against
+synthetic fixtures, not against the shipped catalog. It measured ~122 lines at
+~104 skills and 201 lines at 114, so the budget needs watching rather than
+assuming.
 
 ### Skill Loader
 
@@ -163,11 +169,11 @@ loads bodies on demand — its session receives the registry plus a name and
 description listing, not 18,644 lines. Pi does not. Same registry, same
 symlinks, different loader. Nothing about any skill changed.
 
-Add one field to the install spec, beside the loader root it already owns
-(`src/install-spec.ts:107`, soma#356):
+Add one field to `SubstrateInstallSpec` (`src/install-spec.ts:79`), beside the
+loader root it already owns (`:107`, soma#356):
 
 ```ts
-interface InstallSpec {
+interface SubstrateInstallSpec<S extends InstallSubstrate = InstallSubstrate> {
   skillsLoaderDir(substrateHome: string): string;
   skillsLoading: "on-demand" | "eager";   // NEW
 }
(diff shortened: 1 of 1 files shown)


jcfischer:
Self-review before implementing #542 against this spec. Three corrections; the
third is the one that actually bit during implementation.

## 1. The type is `SubstrateInstallSpec`, not `InstallSpec`

DD-A and the migration plan both name `InstallSpec`. No such type exists. The
declaration is:

```ts
export interface SubstrateInstallSpec<S extends InstallSubstrate = InstallSubstrate> {
```

`src/install-spec.ts:79`. The `skillsLoaderDir` line reference (`:107`) is
correct. Same error in #542's body — worth fixing in both, since an implementer
grepping for `InstallSpec` finds nothing.

## 2. The 300-line registry budget is a design target, not a runtime cap

The doc says the rendered registry sits "inside the existing
`SKILL_REGISTRY_LINE_BUDGET` of 300", which reads as enforcement. The constant's
own docstring is explicit that it is not:

> NOT a runtime-enforced cap — `renderSkills` emits every skill, so a
> pathologically large tree could exceed it

`src/adapters/shared/skill-registry.ts:24-31`. It is asserted in
`test/skill-registry.test.ts` against synthetic fixtures, not against the shipped
catalog. Nothing stops the real catalog crossing 300 as skills accumulate — the
doc should say so, because "inside budget" invites the reader to treat headroom
as guaranteed.

Related datapoint: that docstring records ~122 lines for ~104 skills; I measured
201 for 114. Both are correct at their date, but the growth rate is steeper than
per-skill count alone implies, which strengthens the case for the budget being
watched rather than assumed.

## 3. DD-B is underspecified: a stub is a *directory*, not a catalog line

DD-B says the stub "renders from the same source as the registry entry, so the
two cannot drift". That conflates two different artifacts.

`renderSkillRegistryEntry()` emits one catalog **line**:

```
- **calendar** — Read and write calendar events → /Users/…/.soma/skills/calendar
  triggers: …
  not: …
```

But an eager substrate's loader scans `skillsLoaderDir(home)/<name>/` and reads
`SKILL.md` from inside it (`skillsLoaderUnder("agent")` for pi-dev,
`src/adapters/pi-dev/install.ts:44`). So a stub cannot be a catalog line. It has
to be a **directory containing a frontmatter-only `SKILL.md`**, plus a pointer to
the real body under the registry.

Both artifacts derive from the same `SomaSkill` frontmatter, so they still cannot
drift on *content* — but they are two renderers, and DD-B should say which one it
means. Implementation follows the corrected reading.

## Verified unchanged

- `renderSkills()` `src/adapters/shared/index.ts:199` ✓
- `renderSkillRegistryEntry()` `skill-registry.ts:173` ✓
- `SKILL_REGISTRY_LINE_BUDGET = 300` `skill-registry.ts:31` ✓
- `extractUseWhenTriggers` / `extractAntiTriggers` `skill-registry.ts:110,85` ✓
- `refreshSkillCatalogs()` `src/skill-projection.ts:316` ✓
- `ensureSymlink()` `src/skill-projection.ts:163` ✓
- `skillsLoaderDir()` `src/install-spec.ts:107`, 6 adapters ✓
- `SomaSkillManifest` `src/types.ts:958`, 0 `soma-skill.json` on disk ✓
- `SkillRoute` / `estimatedTokens` / `defaultLoad` — 0 occurrences in `src/` ✓

One thing the review strengthened rather than weakened: `skill-registry.ts:8-13`
states outright that the catalog is eager context in every session and that only
a skill's *body* loads on demand. That is the premise DD-A rests on, written
independently.

Pushing fixes for 1–3, then implementing #542 on `feat/skills-loading`.

Issue #547, opened 2026-08-06
scratch: posture close


jcfischer:
Scratch node verifying the single-operator ratification path.

jcfischer:
Scratch proposal — mechanical test of the declared single-operator path, not a judgement about any real work.

jcfischer:
## Close receipt

- **checkpoint:** `cp-scratch-546`
- **closed by:** jcfischer
- **at:** 2026-08-06T08:05:11.138Z
- **attestation:** `self-attested`

### Evidence

- `approved` — ratified by jcfischer via reaction on proposal comment 5202037632 (https://github.com/the-metafactory/soma/issues/547#issuecomment-5202037632)

### Attestation facts

```json
{
  "backendCapability": "verifiable",
  "confinement": {
    "checked": true,
    "reachableIdentities": [
      "jcfischer",
      "keychain:gh:github.com"
    ],
    "at": "2026-08-06T08:05:10.591Z",
    "probes": [
      {
        "name": "gh auth status (token env stripped)",
        "observed": "exit 0; identities: jcfischer"
      },
      {
        "name": "gh auth token (token env stripped)",
        "observed": "printed a credential"
      },
      {
        "name": "security find-generic-password -s gh:github.com",
        "observed": "keychain item readable"
      }
    ]
  },
  "proposal": {
    "commentId": "5202037632",
    "author": "jcfischer"
  },
  "ratification": {
    "kind": "reaction",
    "id": "393732844",
    "author": "jcfischer"
  },
  "root": {
    "nodeId": "546",
    "author": "jcfischer"
  },
  "reasons": [
    "credentials other than jcfischer are reachable from this session: keychain:gh:github.com",
    "proposal and ratification share an author (jcfischer) — one credential, not two",
    "graph root 546 is authored by the acting identity (jcfischer)"
  ]
}
```

Issue #546, opened 2026-08-06
scratch: operator-posture walk (delete me)


jcfischer:
Scratch root for verifying feat/graph-posture end-to-end. Closed immediately.

jcfischer:
Scratch root for the feat/graph-posture acceptance walk. Node #547 closed `self-attested` as designed.

PR #532, opened 2026-08-05
feat(work-graph): labels as the human index for a map

jcfischer:
Charting a real map in another repo surfaced this immediately: ten issues, no
labels, the root indistinguishable from its nine children in `gh issue list`.

The orienteer re-base dropped wayfinder's `wayfinder:map` / `wayfinder:<type>`
convention on the grounds that no verb reads a label — topology comes from native
edges, `kind` from the typed node block. Runtime-correct, human-wrong. Labels
were never a convention the *verbs* depended on; they were the **reader's
index**, and the node block cannot do that job because it is an HTML comment:
invisible in list views, in the tracker UI, in search.

The tell was in this repo's own history — three nodes on map #495 got
`wayfinder:grilling` applied by hand while walking it, compensating for something
the same session had just removed.

## What changes

`soma graph add` gains a repeatable `--label`, carried on `CreateNodeSpec` and
written at issue creation. That is the whole change.

## The invariant that makes it safe

Labels are **write-only**. `readNode` derives `kind` and `autonomy` from the
typed block and nothing else, so a label that is mistyped, edited, stale, or
absent changes what a human sees and never what a verb decides. `toNode` strips
`labels` alongside `body` and `parent` so they never become node state, and a
test pins that a node labelled `orienteer:task` whose block says `grilling`
reports **grilling**.

Nothing derives a label from `kind` — the caller supplies both, independently.
An earlier revision called this "a projection of the node block", which claimed a
derivation that does not exist; it is now described as what it is.

## Why not derive `orienteer:${kind}` in the runtime?

Deriving would eliminate the drift this PR concedes, which is a real advantage.
It is rejected because **`kind` is deliberately free-form**: §2.1 fixes that the
runtime "never interprets its MEANING but normalizes its FORM", and the work-kind
vocabulary is doctrine owned by orienteer, not by the graph. Deriving would put a
skill's vocabulary — including its `orienteer:` prefix — inside the runtime, and
mint whatever a caller put in `kind` into the repo's label set, silently, because
GitHub creates unknown labels rather than rejecting them (verified below).

The map root settles it: it has no `kind` at all, so derivation cannot label it,
and `orienteer:map` — the one label that makes a map findable — would still need
an explicit path. A rule that covers every node except the most important one is
not worth having.

## One doc claim was wrong and is corrected

The shipped doctrine said "create the `orienteer:*` labels before the first `add`
— the create call fails on a label that does not exist." **Untested, and false.**
[...]

Issue #511, opened 2026-08-02
Cut agent sessions off from the principal's GitHub credential


jcfischer:
## Question

Graduated from the map's **"verified-receipt cutover ceremony"** fog, made specifiable by #496's resolution.

#496 achieved authorship separation — proposal by `ivy-agent`, ratification by `jcfischer`, both API-attested — and its receipt is still `attestation: unverified`. The reason: the session that posted the proposal can also resolve the principal's credential through the local `gh` keyring, so it could have produced the ratifying 👍 itself. The derivation rule under decision in #502 cannot detect this; it would mark such a receipt `verified` and be **wrong**.

Setting `GH_TOKEN` to the agent PAT is not sufficient and never was. It changes which identity `gh` *prefers*, not which identities the session can *reach* — #496's own probe script demonstrates the gap, using `GH_TOKEN=""` to fall back to the keyring and act as the principal deliberately. Any agent session can do the same, plus `gh auth token` prints the principal's credential outright.

**Decide how agent sessions become structurally incapable of acting as the principal.** Candidates, each with a real cost:

- **Separate macOS user account** for agent sessions — its own login keychain, no principal credential in it. Strongest isolation; heaviest ergonomics (file access, tooling, GUI).
- **Container / VM** with only the agent PAT injected. Clean boundary; friction with local dev tooling and the soma dev tree.
- **`GH_CONFIG_DIR` pointing at an agent-only config** plus keychain-ACL hardening so `security find-generic-password` is refused for the `gh` item. Cheapest; weakest — an ACL prompt is a speed bump, not a wall.
- **Harness-level denial** — sandbox or permission rules refusing `gh auth token`, `security find-generic-password`, and bare `GH_TOKEN=` invocations. Enforcement lives in the harness rather than the OS; note §1 clause 5 wants enforcement out of the agent's reach, and a permission rule the agent can prompt around does not qualify.
- **Physical split** — the principal's `gh` login lives only on a machine that never runs agent sessions.

### Why this is not optional

Until it holds, every HITL receipt this deployment writes is honestly `unverified`, no matter how clean the authorship separation looks. #496 is the worked example: it looks exactly like a verified gate and is not one. A rule that cannot tell those two apart, deployed without this invariant, produces *confident* wrong answers — worse than the degraded mode it replaces.

### Relationship to #502

[...]

PR #548, opened 2026-08-06
feat(work-graph): declare operator posture, and say so in the receipt

jcfischer:
Closes the deadlock that forced a hand override on
[#499](https://github.com/the-metafactory/soma/issues/499).

On a machine run by one person, `soma graph close --propose` posts under their
login, so no second account exists to ratify and a HITL node can **never** close.
That makes `attestation` a *gate* when §3.2 is explicit that it is a label.

## Declared, because it is not derivable

Two derivations were tried on
[#532](https://github.com/the-metafactory/soma/pull/532) and both failed review:

- **graph-root authorship** — self-conferred. `findGraphRoot` returns the node
  itself when it has no parent, so anyone who opens a parentless node holds the
  role.
- **one reachable credential** — universal. Any contributor running a single
  personal token looks identical to a solo adopter.

"Am I the only person working here" is a property of the *deployment*, and
nothing observable distinguishes it. So the adopter states it once:
`~/.soma/policy/graph-posture.json`, in the probe registry's shape and for the
same reasons — soma-home keeps it out of the agent's reach (§1 clause 5), and
`soma policy posture` only **reads** it, because declaring it loosens a gate
(§4). Absent, unparsable, or `false` ⇒ declares nothing, and the close refuses
exactly as before.

**Per home, not per repo.** The claim is about who sits at this machine and does
not change with the repository a map lives in. (The probe registry is repo-scoped
because a *command* is only safe in context; an operator is not.)

## A third receipt state, not a softened second

A declared solo close records **`self-attested`**.

`unverified` means *we do not know who approved this* — it conflates a reachable
second credential, a wrong ratifier, and no ratification at all. A declared solo
close is none of those: it is known exactly. `self-attested` is weaker than
`verified` and more precise than `unverified`, and is reachable from neither —
nothing upgrades it, and a clean four-conjunct close is unaffected.

### The rule is stated positively, and that matters

The first draft keyed on "the shared-author reason is the *only* one left". That
would have meant the label **never fires on the deployment it exists for**,
because two of the reasons are *expected* there: the operator's own keyring is
reachable, and the operator authored the map. So it is stated as what must hold —
a backend that can attest, a proposal, a ratification, and a ratifier who is both
the proposer and the map's owner.

## The hand override is gone from doctrine

#499 was closed past a refusal by hand, with the refusal recorded on the node.
`closing.md` now says plainly that closing past a refusal produces a receipt
claiming a ratification the graph never saw. That override should be the last.

## Verification

- `bun test` — 2226 pass / 0 fail. `bunx tsc --noEmit` — clean. eslint clean on
  the changed files.
- Nine unit tests: parsing and per-home resolution; only an explicit `true`
  declares anything; a self-👍 ratifies **only** when declared; it never outranks
  a real second credential; the root author's 👎 still suppresses everything;
  `self-attested` covers no other failure mode (stranger ratifier, wrong root
  author, non-attesting backend all stay `unverified`); and a genuinely verified
  close stays `verified`.
- **Walked live on this deployment** against scratch root #546 → node #547:
  declared the posture, `--propose`, self-👍, close. Result:
  `attestation: self-attested`. The same sequence refused on #499 before this
  branch. Both scratch issues closed.

## Known limits, carried forward

- The declaration is only as trustworthy as soma-home is unreachable from the
  agent — #511's invariant, the same base the probe registry rests on.
- `self-attested` asserts that one declared operator approved their own work. It
  is not evidence a second human looked, and does not pretend to be.

### DIFF
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -345,6 +345,20 @@ reaction on that specific comment ID — or a principal-authored comment, which
 wins when amending — verified via the API author field. A materially amended
 proposal is re-posted and needs fresh ratification (the replay-rebind lesson).
 
+**Selecting the ratifier.** A 👎 from the root node's author suppresses
+ratification outright. Otherwise any 👍 from someone other than the proposer
+ratifies — the root author's if present, else the first by author order.
+
+A proposer's own 👍 ratifies in exactly one case: soma-home declares
+`singleOperator` (`policy/graph-posture.json`, read by `soma policy posture`).
+That declaration is **per home, not per repo** — the claim is about who sits at
+this machine, and it does not change with the repository a map lives in — and it
+is *declared* rather than derived because the condition is not observable. Two
+derivations were tried and both failed: graph-root authorship is self-conferred
+(`findGraphRoot` returns the node itself when it has no parent), and a single
+reachable credential describes every solo-token contributor. Absent, unparsable,
+or `false` ⇒ the self-👍 ratifies nothing and the close refuses.
+
 #### Deriving `attestation` (#502)
 
 `attestation` is **derived at close time, never configured**. There is no flag,
@@ -372,7 +386,14 @@ ratified this* — not merely that two logins appear. All four must hold:
    authorization question. Root unreachable, or root authored by the acting
    agent identity → `unverified`.
 
-Any conjunct failing yields `unverified`. `attestation` is a **label, not a
+Any conjunct failing yields `unverified` — with one named exception. Where
+soma-home declares one operator, the backend can attest, and the sole operator
+ratified their own proposal on their own map, the receipt reads
+**`self-attested`**. That is a third state, not a softening of the second:
+`unverified` means *we do not know who approved this*, and conflates a reachable
+second credential, a wrong ratifier, and no ratification at all. A declared solo
+close is none of those — it is known exactly, and weaker than `verified` in a way
+the label states. It never becomes `verified` by any route. `attestation` is a **label, not a
 gate**: `close` proceeds either way. Refusing on `unverified` would deadlock
 the bootstrap — the nodes that establish credential separation are themselves
 `approve`-class, so they could never close.
--- a/src/cli/graph.ts
+++ b/src/cli/graph.ts
@@ -57,6 +57,11 @@ import {
   type ProbeRegistry,
 } from "../work-graph-probe-registry";
 import { allProbesPassed, runCommand, runProbes as defaultRunProbes } from "../work-graph-probes";
+import {
+  declaresSingleOperator,
+  loadGraphPosture as defaultLoadGraphPosture,
+  type GraphPosture,
+} from "../work-graph-posture";
 import { SomaCliError } from "./errors";
 import { readOption } from "./parse-utils";
 
@@ -401,6 +406,12 @@ export interface GraphCliDeps {
    * never per invocation.
    */
   loadProbeRegistry: (repo: string) => Promise<ProbeRegistry>;
+  /**
+   * Whether soma-home declares one operator on this machine (§3.2). Per home,
+   * not per repo — the claim is about who sits here, and that does not change
+   * with the repository a map lives in.
+   */
+  loadGraphPosture: () => Promise<GraphPosture>;
   runProbes: (probes: readonly Probe[], registry: ProbeRegistry) => Promise<ProbeResult[]>;
   checkConfinement: () => Promise<ConfinementResult>;
   /**
@@ -472,6 +483,7 @@ function defaultDeps(): GraphCliDeps {
     resolveRepo: resolveGraphRepo,
     resolveIdentity: async () => await gh(["api", "user", "--jq", ".login"]),
     loadProbeRegistry: async (repo) => await defaultLoadProbeRegistry({ repo }),
+    loadGraphPosture: async () => await defaultLoadGraphPosture(),
     runProbes: async (probes, registry) => await defaultRunProbes(probes, { registry }),
     checkConfinement: async () =>
       await defaultCheckConfinement({
@@ -654,6 +666,8 @@ export function selectRatification(
   reactions: readonly Reaction[],
   proposalAuthor: string,
   rootAuthor: string | undefined,
+  /** Set only from the soma-home declaration; never derived (see work-graph-posture.ts). */
+  singleOperatorDeclared = false,
 ): { kind: "reaction"; id: string; author: string } | undefined {
   if (
     rootAuthor !== undefined &&
@@ -662,13 +676,24 @@ export function selectRatification(
     return undefined;
   }
 
-  const approvals = reactions
-    .filter((reaction) => reaction.content === "+1" && reaction.author !== proposalAuthor)
-    .sort((left, right) => (left.author < right.author ? -1 : left.author > right.author ? 1 : 0));
+  const byAuthor = (left: Reaction, right: Reaction): number =>
+    left.author < right.author ? -1 : left.author > right.author ? 1 : 0;
+  const thumbsUp = reactions.filter((reaction) => reaction.content === "+1");
 
-  if (approvals.length === 0) return undefined;
-  const preferred = approvals.find((reaction) => reaction.author === rootAuthor) ?? approvals[0];
-  return { kind: "reaction", id: preferred.id, author: preferred.author };
+  const approvals = thumbsUp.filter((reaction) => reaction.author !== proposalAuthor).sort(byAuthor);
+  if (approvals.length > 0) {
+    const preferred = approvals.find((reaction) => reaction.author === rootAuthor) ?? approvals[0];
+    return { kind: "reaction", id: preferred.id, author: preferred.author };
+  }
+
+  // Last resort, and only on a machine that DECLARED one operator. Absent the
+  // declaration this returns nothing and the close refuses — the correct answer
+  // everywhere else, because a proposer thumbing their own work is not a second
+  // human looking at it.
+  if (!singleOperatorDeclared) return undefined;
+  const selfApproval = thumbsUp.filter((reaction) => reaction.author === proposalAuthor).sort(byAuthor).at(0);
+  if (selfApproval === undefined) return undefined;
+  return { kind: "reaction", id: selfApproval.id, author: selfApproval.author };
 }
 
 async function runClose(
@@ -780,6 +805,11 @@ async function runClose(
 
   const root = await findGraphRoot(ref, async (nodeRef) => await graph.readNode(nodeRef));
 
+  // Read before the ratification path: whether this machine declares one
+  // operator decides whether a self-👍 counts, and it feeds the receipt.
+  const posture = await deps.loadGraphPosture();
+  const singleOperator = declaresSingleOperator(posture);
+
   let proposal: { commentId: string; author: string } | undefined;
   let ratification: { kind: "reaction" | "comment"; id: string; author: string } | undefined;
 
@@ -798,7 +828,7 @@ async function runClose(
     const proposalRef: CommentRef = await graph.readComment({ id: commentId, nodeId: ref.id });
     proposal = { commentId: proposalRef.id, author: proposalRef.author ?? "" };
     const reactions = await graph.readCommentReactions(proposalRef);
-    ratification = selectRatification(reactions, proposal.author, root?.author);
+    ratification = selectRatification(reactions, proposal.author, root?.author, singleOperator);
     if (ratification !== undefined) {
       evidence.push({
         kind: "approved",
@@ -816,6 +846,7 @@ async function runClose(
     ...(proposal === undefined ? {} : { proposal }),
     ...(ratification === undefined ? {} : { ratification }),
     ...(root === undefined ? {} : { root }),
+    singleOperatorDeclared: singleOperator,
   });
 
   const receipt: CloseReceipt = {
--- a/src/cli/policy.ts
+++ b/src/cli/policy.ts
@@ -11,6 +11,7 @@ import type {
   SomaPolicyCheckResult,
 } from "../types";
 import { loadProbeRegistry, type ProbeRegistry } from "../work-graph-probe-registry";
+import { loadGraphPosture, renderStarterPosture, type GraphPosture } from "../work-graph-posture";
 import { resolveGraphRepo } from "./graph";
 import { readOption } from "./parse-utils";
 import { parseSubstrate } from "./substrate";
@@ -58,7 +59,15 @@ export interface ParsedPolicyProbesArgs {
   json: boolean;
 }
 
+export interface ParsedPolicyPostureArgs {
+  command: "policy";
+  action: "posture";
+  options: { homeDir?: string; somaHome?: string };
+  json: boolean;
+}
+
 export type ParsedPolicyArgs =
+  | ParsedPolicyPostureArgs
   | ParsedPolicyCheckArgs
   | ParsedPolicyScanArgs
   | ParsedPolicyPromoteArgs
@@ -76,11 +85,12 @@ const POLICY_INSPECT_USAGE =
   "Usage: soma policy inspect --surface <prompt|tool_call|permission_request|config_change|governance_event> [--prompt <text>|--prompt-env <name>] [--tool-name <name> --tool-input-env <name>] [--permission-request-env <name>] [--config-change-env <name>] [--substrate <id>] [--record <all|deny|none>] [--json]";
 const POLICY_GUARD_USAGE =
   "Usage: soma policy guard --substrate <id> --tool-name <name> --tool-input-env <name> [--cwd <dir>] [--soma-home <dir>] [--home-dir <dir>] [--private-root <dir>]… [--record <all|deny|none>] [--json]";
+const POLICY_POSTURE_USAGE = "Usage: soma policy posture [--soma-home <dir>] [--home-dir <dir>] [--json]";
 const POLICY_PROBES_USAGE =
   "Usage: soma policy probes [--repo <owner/name>] [--soma-home <dir>] [--home-dir <dir>] [--json]";
 
 export const POLICY_COMMAND_HELP: { usage: string; subcommands: Record<ParsedPolicyArgs["action"], string> } = {
-  usage: [POLICY_CHECK_USAGE, POLICY_SCAN_USAGE, POLICY_PROMOTE_USAGE, POLICY_INSPECT_USAGE, POLICY_GUARD_USAGE, POLICY_PROBES_USAGE].join("\n"),
+  usage: [POLICY_CHECK_USAGE, POLICY_SCAN_USAGE, POLICY_PROMOTE_USAGE, POLICY_INSPECT_USAGE, POLICY_GUARD_USAGE, POLICY_PROBES_USAGE, POLICY_POSTURE_USAGE].join("\n"),
   subcommands: {
     check: POLICY_CHECK_USAGE,
     scan: POLICY_SCAN_USAGE,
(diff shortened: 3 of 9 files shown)


jcfischer (commented):

18 finding(s): 2 blocker, 4 important, 8 suggestion, 4 nit; 1 lens(es) failed to run: Performance.


### CodeQuality

- **[important]** `src/cli/graph.ts:691` — **Self-👍 gate omits the map-ownership check the label requires**
  `if (!singleOperatorDeclared) return undefined;` is the only guard on the self-ratification path, while `deriveAttestation` additionally requires `ratifier === root?.author`. A contributor whose home declares `singleOperator` can therefore self-ratify and close HITL nodes on a shared repo's map they did not author, and the receipt records the same `unverified` as a stranger's thumb.
  Fix: Gate the self-approval fallback on `proposalAuthor === rootAuthor` too, so the declaration only loosens closes on the operator's own map.

- **[suggestion]** `src/cli/graph.ts:486` — **Posture dep ignores soma-home overrides passed to sibling deps**
  `loadGraphPosture: async () => await defaultLoadGraphPosture(),` passes no options, unlike the adjacent `defaultCheckConfinement({...})` which is threaded home options. A `graph close` run against an alternate soma-home reads the real `~/.soma/policy/graph-posture.json`.
  Fix: Thread the same home/soma-home resolution into `defaultLoadGraphPosture({...})`.

- **[suggestion]** `src/work-graph-attestation.ts:240` — **self-attested won't be invalidated by future conjuncts**
  `const selfAttested = ... proposer === ratifier && ratifier === root?.author;` re-derives the state positively and ignores `reasons` entirely, so any conjunct added to §3.2 later will silently be covered by `self-attested` instead of downgrading to `unverified`.
  Fix: Also assert the surviving `reasons` are a subset of the two expected ones (own keyring reachable, operator authored the map).

- **[nit]** `src/cli/graph.ts:693` — **Dead sort on single-author reaction list**
  `thumbsUp.filter((reaction) => reaction.author === proposalAuthor).sort(byAuthor).at(0)` sorts a list whose entries all share one author, so `byAuthor` never reorders anything.
  Fix: Drop `.sort(byAuthor)` and use `.find((reaction) => reaction.author === proposalAuthor)`.

- **[nit]** `src/cli/policy.ts:443` — **Template literal with no interpolation**
  `` `status: declared` `` uses backticks with no substitution, unlike the plain strings around it.
  Fix: Use `"status: declared"`.

### Security

- **[blocker]** `src/cli/graph.ts:690` — **Declared posture lets the agent satisfy the HITL gate unaided**
  The self-ratification path accepts a 👍 whose author equals `proposalAuthor` — and per `closing.md`, "the agent posts the proposal under their login", so the agent already holds that identity. With one repo token it can propose, react 👍, and close, producing `self-attested`, a receipt whose documented meaning is "the sole operator ratified their own proposal" — a claim nothing in the flow substantiates. The stated limits cover only soma-home reachability and "not evidence a second human looked", not "no human need be involved at all".
  Fix: Bind the loosening to a signal the agent cannot mint: require the ratification to be an out-of-band operator action (e.g. a reaction posted under a credential the agent's environment does not carry), or record on the receipt that proposer and ratifier shared the agent's own credential.

- **[important]** `src/cli/graph.ts:694` — **Self-ratification not restricted to the operator's own map**
  `selectRatification` gates the self-👍 solely on `singleOperatorDeclared` and `reaction.author === proposalAuthor`, with no root-author check — while `deriveAttestation` requires `ratifier === root?.author`. On a declared machine, a HITL node under a map rooted by an upstream maintainer now closes on the operator's own thumb, with `approved` evidence recorded and a bare `unverified` label indistinguishable from ordinary unverified.
  Fix: Apply the same `proposalAuthor === rootAuthor` conjunct in `selectRatification` that `deriveAttestation` already enforces, so the gate is loosened exactly where `self-attested` can fire.

### Architecture

- **[suggestion]** `src/cli/graph.ts:666` — **Ratification rule lives in the CLI layer, not the domain**
  `selectRatification(reactions, proposal.author, root?.author, singleOperator)` is now a policy-bearing rule — it decides when a self-👍 counts — but it sits in `src/cli/graph.ts` and is imported by the test from `../src/cli/graph`, while its sibling rule `deriveAttestation` lives in `src/work-graph-attestation.ts`. The same declared-single-operator fact is consumed in two layers, so the rule can drift between them.
  Fix: Move `selectRatification` next to `deriveAttestation` in the work-graph domain modules; keep `src/cli/graph.ts` as the wiring that reads the posture and passes it in.

- **[suggestion]** `src/cli/policy.ts:417` — **Posture prose and remediation text hardcoded in the CLI**
  `formatGraphPosture` embeds the domain explanation ("a self-👍 is not a second human", "recorded `self-attested` — never `verified`") in `src/cli/policy.ts`, while `renderStarterPosture()` — the one piece that already lives in the domain module — is imported. The split means the same contract is stated in two places and only half of it moves with the module.
  Fix: Move the status-explanation strings into `work-graph-posture.ts` alongside `renderStarterPosture`, leaving `formatGraphPosture` to assemble lines.

### Performance — DID NOT RUN

- **[important]** `(lens output):0` — **Performance: model deviated from JSON contract**
  claude extraction failed (exit-nonzero) exit=1
--- output ---
{"is_error":true,"duration_api_ms":0,"num_turns":1,"stop_reason":"stop_sequence","session_id":"23fd61db-3595-4dee-9562-873184f93c45","total_cost_usd":0,"usage":{"input_tokens":0,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":0,"server_tool_use":{"web_search_requests":0,"web_fetch_requests":0},"service_tier":"standard","cache_creation":{"ephemeral_1h_input_tokens":0,"ephemeral_5m_input_tokens":0},"inference_geo":"","iterations":[],"speed":"standard"},"modelUsage":{},"permission_denials":[],"terminal_reason":"api_error","fast_mode_state":"off","fast_mode_disabled_reason":"sdk_opt_in_required","subtype":"success","api_error_status":529,"result":"API Error: 529 Overloaded. This is a server-side issue, usually temporary — try again in a moment. If it persists, check https://status.claude.com.","type":"result","duration_ms":236280,"uuid":"f9d65f90-656f-4f29-b598-28fc49466370"}


### Maintainability

- **[suggestion]** `src/work-graph-posture.ts:14` — **Derivation-failure rationale maintained in four places**
  The same explanation — "graph-root authorship — self-conferred. `findGraphRoot` returns the node itself when it has no parent" and "one reachable credential — universal" — appears verbatim-in-substance in `work-graph-posture.ts`, `docs/work-graph.md` ("graph-root authorship is self-conferred (`findGraphRoot` returns the node itself when it has no parent)"), and again in the `unverified` gloss repeated across `work-graph.ts`, `work-graph-attestation.ts`, `closing.md` and `policy.ts`. Any future change to the rule means editing four to six copies, and they will drift.
  Fix: Keep the full argument in docs/work-graph.md §3.2; reduce the module docstring and the `deriveAttestation` comment to one line plus a pointer ("why declared, not derived: docs/work-graph.md §3.2").

- **[suggestion]** `src/cli/policy.ts:325` — **parsePolicyPostureArgs duplicates parsePolicyProbesArgs flag loop**
  `parsePolicyPostureArgs` is `parsePolicyProbesArgs` minus `--repo`: same `for` loop, same `case "--home-dir": options.homeDir = readOption(rest, index, arg); index += 1;` / `--soma-home` / `--json` arms, same trailing `default: throw new Error(...)`. A new home-scoped flag now has to be added twice.
  Fix: Extract `parseHomeAndJsonFlags(rest, usage)` returning `{ homeDir?, somaHome?, json, rest }`; probes layers `--repo` on top.

- **[nit]** `src/cli/graph.ts:668` — **Trailing boolean param makes the callsite unreadable**
  `selectRatification(reactions, proposal.author, root?.author, singleOperator)` reads as an anonymous flag at the callsite; the meaning lives only in the `singleOperatorDeclared = false` declaration. Adding a second policy input will push this toward a positional-boolean pile.
  Fix: Take the fourth argument as an options object: `selectRatification(reactions, proposer, rootAuthor, { singleOperatorDeclared })`.

- **[nit]** `src/cli/graph.ts:486` — **Redundant async wrapper around zero-arg default**
  `loadGraphPosture: async () => await defaultLoadGraphPosture(),` adds a closure with no arguments to bind, unlike the neighbouring `runProbes` wrapper which does supply `{ registry }`.
  Fix: `loadGraphPosture: defaultLoadGraphPosture,`

### HonestOracle

- **[blocker]** `src/work-graph-attestation.ts:216` — **self-attested claims a human approved; evidence is only a token**
  The comment asserts a declared solo close "is a *known* state rather than an unknown one" and the receipt reads `self-attested`, but the only observable is a 👍 under the proposer's login — the same login the PR says the agent posts under ("`soma graph close --propose` posts under their login"). Nothing in the conjuncts distinguishes the operator reacting from the agent reacting with the operator's token, so the receipt asserts human approval it did not witness; the "Known limits" section names only "not evidence a second human looked", never "not evidence any human looked".

- **[important]** `src/skills/orienteer/references/closing.md:101` — **Deadlock premise is a config choice stated as a law**
  "the agent posts the proposal under their login, so no second account exists to ratify and the close would refuse forever" is load-bearing for the entire new state, yet it holds only because the agent shares the operator's credential. The counter-case — the agent proposing under a distinct bot identity, which restores a genuine second credential and keeps `verified` reachable — is never engaged; the PR considers only two failed *derivations* of solo-ness, not alternatives to the self-👍.

- **[suggestion]** `src/skills/orienteer/references/closing.md:113` — **"there is no hand override" is doctrine presented as enforcement**
  "Undeclared, the refusal stands and **there is no hand override**" reads as a system property, but nothing in the diff prevents a node being closed outside `soma graph close` — #499 was closed exactly that way. An agent reading this reference may treat a prohibition as an impossibility.

- **[suggestion]** `src/cli/policy.ts:350` — **policy posture can report a posture the close path never reads**
  `runPolicyCli` honours `--home-dir`/`--soma-home` when loading the posture, while `graph.ts` wires `loadGraphPosture: async () => await defaultLoadGraphPosture()` with no options. `closing.md` tells the operator to "Read it with `soma policy posture`", so an invocation with either flag can print "status: declared / singleOperator: true" for a file `graph close` will never consult.


---

jcfischer (commented):

14 finding(s): 1 blocker, 3 important, 6 suggestion, 4 nit; 1 lens(es) failed to run: HonestOracle.


### CodeQuality

- **[important]** `src/cli/graph.ts:808` — **Close silently discards an invalid posture declaration**
  `const posture = await deps.loadGraphPosture(); const singleOperator = declaresSingleOperator(posture);` collapses `absent` and `invalid` into one boolean, so a typo'd `graph-posture.json` produces the same silent refusal as no file at all. `loadGraphPosture` carefully computes `reason` for exactly this case and the close path throws it away.
  Fix: When `posture.status === "invalid"`, surface `posture.reason` and `posture.path` in the refusal message (or a warning line on the receipt path).

- **[suggestion]** `src/work-graph.ts:210` — **AttestationState doc omits two of the four self-attested conditions**
  The type doc says `self-attested` means "the machine **declares one operator** ... and the proposer ratified their own proposal", but `deriveAttestation` additionally requires `backendCapability === "verifiable"` and `ratifier === root?.author`. This is the doc consumers read from the type, and it claims a looser rule than the code enforces.
  Fix: Mirror docs/work-graph.md's wording: declared operator, attesting backend, and the operator ratified their own proposal *on their own map*.

- **[nit]** `src/cli/policy.ts:438` — **Invalid-posture output omits the starter document**
  The `absent` branch ends with `renderStarterPosture()`, but the `invalid` branch prints only "Nothing is declared until the document parses." The user holding a broken file gets less help than the user holding none.
  Fix: Append `renderStarterPosture()` to the invalid branch too.

- **[nit]** `src/work-graph-posture.ts:118` — **Unknown-key rejection breaks on $schema in a hand-authored file**
  `Object.keys(record).filter((key) => key !== "version" && key !== "singleOperator")` invalidates the whole document for any extra key, and soma ships no verb that writes this file — adopters hand-author it, where a `$schema` line is a common habit.
  Fix: Allowlist `$schema` (and optionally `_comment`) alongside the two recognised keys.

### Security

- **[blocker]** `src/work-graph-posture.ts:60` — **$HOME redirection forges the posture declaration**
  `const home = resolve(options.homeDir ?? homedir());` — the close path calls `defaultLoadGraphPosture()` with no options, and `homedir()` honours `$HOME`, so `HOME=/tmp/x soma graph close` reads an attacker-placed `graph-posture.json`. The stated guarantee "Absent, unparsable, or unreadable ⇒ **not declared**" fails against any process that can set an env var — including the agent §1 clause 5 keeps out of soma-home.
  Fix: Resolve the posture path from a confinement-verified soma-home rather than `homedir()`, and record the resolved path on the receipt so a redirected read is visible.

- **[important]** `src/work-graph-posture.ts:45` — **Declaration carries no operator identity despite §4 claim**
  The header cites "§4: loosening is identity-bound and fail-closed", but the shape is `{ status: "declared"; path: string; singleOperator: boolean }` — it names no operator, and `unknownKeys` rejects any attempt to add one. A file declared by operator A keeps relaxing the gate for every later user of that home (shared workstation, baked container image), whose self-👍 then ratifies their own proposals.
  Fix: Add `operator: "<login>"` to the document and require it to equal both the proposer and the root author before a self-👍 ratifies.

- **[suggestion]** `src/cli/graph.ts:833` — **Self-👍 recorded as indistinguishable `approved` evidence**
  `evidence.push({ kind: "approved", ...})` fires for a self-ratification exactly as for a second-party one; only the separate `attestation` field carries the difference. A downstream consumer auditing `evidence` alone reads an approval no second party gave.
  Fix: Tag the evidence entry when proposer === ratifier, or omit `approved` evidence for a self-ratification.

### Architecture

- **[suggestion]** `src/work-graph.ts:221` — **Third attestation state widens an exported union**
  `export type AttestationState = "verified" | "self-attested" | "unverified"` adds a member to a type re-exported from `src/index.ts`; any external consumer with an exhaustive switch or a `attestation === "unverified"` negative check now silently mishandles the new value. The change is on the public surface, so it is effectively irreversible without a major bump.
  Fix: Note the union widening in the changelog/release notes as a consumer-visible surface change.

- **[suggestion]** `src/cli/policy.ts:417` — **policy.ts imports from graph.ts and work-graph-posture.ts**
  `import { loadGraphPosture, renderStarterPosture, type GraphPosture } from "../work-graph-posture"` plus the existing `import { resolveGraphRepo } from "./graph"` makes the policy CLI a second consumer of graph-owned modules, and `formatGraphPosture` renders work-graph doctrine ("§3.2", `self-attested`) from the policy command. The reading surface for graph posture now lives in two CLI modules.
  Fix: Move `formatGraphPosture` next to `GraphPosture` in `work-graph-posture.ts` so the rendering lives with the type it describes.

### Maintainability

- **[suggestion]** `src/work-graph-posture.ts:12` — **Derivation-failure rationale copy-pasted across three files**
  The same two-bullet argument — "graph-root authorship — self-conferred. `findGraphRoot` returns the node itself when it has no parent" and "one reachable credential — universal. Any contributor running a single personal token looks identical to a solo adopter" — appears verbatim in the module header, in `docs/work-graph.md` ("Two derivations were tried and both failed: graph-root authorship is self-conferred..."), and again condensed in `closing.md`. Three copies drift independently on the next edit.
  Fix: Keep the full argument in `docs/work-graph.md` §3.2; have the module header and `closing.md` cite it in one line rather than restate it.

- **[suggestion]** `src/cli/policy.ts:425` — **formatGraphPosture embeds multi-paragraph remediation prose**
  `formatGraphPosture` inlines four lines of user-facing doctrine ("A HITL node cannot close where the proposer is the only account available to / ratify: `soma graph close` refuses, because a self-👍 is not a second human...") plus two more variants in the declared branch, making the formatter a prose store rather than a renderer. `formatProbeRegistry` sits directly below with the same shape, so the pattern will keep growing.
  Fix: Hoist the three message blocks to module-level string constants (or next to `renderStarterPosture` in work-graph-posture.ts) and leave the formatter as pure branch-and-join.

- **[nit]** `src/cli/policy.ts:325` — **parsePolicyPostureArgs duplicates the probes flag loop**
  `parsePolicyPostureArgs` repeats `parsePolicyProbesArgs`'s exact `for`/`switch` over `--home-dir`, `--soma-home`, `--json` with only the `--repo` case and usage string differing. Third such subcommand parser makes the copy the established shape.
  Fix: Extract a shared `parseHomeAndJsonFlags(rest, usage)` helper; `probes` layers `--repo` on top of it.

- **[nit]** `src/cli/graph.ts:668` — **Fourth positional boolean on selectRatification**
  `selectRatification(reactions, proposal.author, root?.author, singleOperator)` now takes four positionals ending in an opaque boolean; the callsite reads as a flag with no name attached, which is why the parameter needs the `/** Set only from the soma-home declaration... */` comment to be legible.
  Fix: Take an options object for the trailing args: `selectRatification(reactions, { proposalAuthor, rootAuthor, singleOperatorDeclared })`.

  claude extraction failed (exit-nonzero) exit=1
--- output ---
{"is_error":true,"duration_api_ms":0,"num_turns":1,"stop_reason":"stop_sequence","session_id":"fde491d7-431d-45d9-9c44-3eaff6ee4339","total_cost_usd":0,"usage":{"input_tokens":0,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":0,"server_tool_use":{"web_search_requests":0,"web_fetch_requests":0},"service_tier":"standard","cache_creation":{"ephemeral_1h_input_tokens":0,"ephemeral_5m_input_tokens":0},"inference_geo":"","iterations":[],"speed":"standard"},"modelUsage":{},"permission_denials":[],"terminal_reason":"api_error","fast_mode_state":"off","fast_mode_disabled_reason":"sdk_opt_in_required","subtype":"success","api_error_status":529,"result":"API Error: 529 Overloaded. This is a server-side issue, usually temporary — try again in a moment. If it persists, check https://status.claude.com.","type":"result","duration_ms":301428,"uuid":"b23d1447-545a-48a6-8732-28651720a1c8"}



---

jcfischer:
Closing unmerged. Sage's review was right and the finding is not patchable — the
approach answers the wrong question.

## What the review found

`self-attested` is documented as "the sole operator ratified their own proposal".
The only observable behind it is a 👍 under the proposer's login — and the agent
holds that login. Nothing in the flow distinguishes *the operator reacting* from
*the agent reacting with the operator's token*, so the receipt asserts a human
approval it never witnessed.

**The acceptance walk in the PR body demonstrates the hole rather than the
feature.** That walk posted its own 👍 via `gh api ... -f content=+1` under
`jcfischer`, with no human involved, and the close returned `self-attested`. It
was written up as evidence the feature works. It is evidence it is unsound.

## Why the premise was wrong

The PR's load-bearing claim — "on a single-operator deployment no second account
exists to ratify, so the close refuses forever" — holds only because the agent
shares the principal's credential. That is a **configuration gap, not a property
of solo deployments**.

`ivy-agent` already exists (#496: outside collaborator, Triage role, PAT in
Keychain). With the agent proposing under *that* identity and the principal
ratifying as themselves, there are genuinely two credentials: conjunct 3 passes
honestly, and `verified` becomes reachable rather than being traded away.

That is what §5 specifies — "agent sessions run with `GH_TOKEN` = the bot PAT;
the principal's credential never enters an agent session" — and it is exactly
what **#511** is for.

Two derivations of "single operator" failed review on #532, and the conclusion
drawn here was *therefore declare it*. The correct conclusion was *therefore the
credential split is unfinished*. The #499 deadlock was never evidence that solo
deployments need a weaker gate; it was evidence that #511 is load-bearing.

## What is kept

- The branch `feat/graph-posture` (`be187bd`) stays for reference; nothing merges.
- The **hand-override removal** in `closing.md` was right for independent reasons
  and should return with whatever lands: closing past a refusal produces a
  receipt claiming a ratification the graph never saw.
- The general shape — an adopter declaration in soma-home, out of the agent's
  reach, fail-closed — remains the right pattern for facts that genuinely cannot
  be derived. Operator count was not one of them.

Superseded by #511.

Issue #510, opened 2026-08-02
Ivy-authored code: branch protection on main, then Write for ivy-agent


jcfischer:
## Task

`ivy-agent` should be able to author code, so the principal's contributions and the assistant's are separable — and separable **derivably** (pushed by a credential the principal does not hold), not merely declared through a commit author email. Declared authorship is forgeable by any writer including the agent, which is the evidence class §1 clause 1 rejects; only credential-bound push gives the property.

**Decided 2026-08-02:** Write role on the upstream repo, gated by branch protection. Not fork-and-PR (rejected: fork sync friction, degraded confidentiality scan on fork PRs, review-loop assumes same-repo branches). Not commit-authorship-only (rejected: declared, not derived).

### Mandatory ordering — protection first, grant second

`main` has **no branch protection today** — `GET /repos/the-metafactory/soma/branches/main/protection` → `404 Branch not protected`, verified 2026-08-02. DD-16 §4 puts main-branch writes at an `approve` floor and §1 clause 5 requires enforcement to sit out of the agent's reach. Granting Write before protection exists would leave that floor enforced by prompt text alone — precisely the failure mode the spec was written against.

1. Enable a ruleset / branch protection on `main`: require a pull request, require ≥1 approving review, dismiss stale approvals on new commits, ban force-push and deletion, and grant **no bypass** to the agent identity.
2. Only then raise the role: `PUT /repos/the-metafactory/soma/collaborators/ivy-agent -f permission=push`. Costs nothing — outside collaborators are billed only on private repos.
3. Probe both directions: agent **can** push a feature branch and open a PR; agent **cannot** push to `main`.

### Questions this ticket answers

- What the ruleset requires exactly — approval count, required status checks, linear history. Note `confidentiality-gate` is deliberately *not* a required check today (warn-only burn-in, by design); decide whether that changes when a non-human identity starts authoring code.
- Whether `enforce_admins` binds the principal too. If admins bypass, the floor is real for the agent and advisory for the human — possibly intended, but it should be a decision rather than a default.
- How the agent's git identity gets configured (`user.email` → `ivy-agent@users.noreply.github.com`) so GitHub attributes commits to the account, and whether soma's session launcher sets it.
- Whether `Co-Authored-By: Claude` trailers retire once real authorship exists.

### Consequences to record when this lands

[...]

Issue #500, opened 2026-08-02
Autonomy policy attachment on the runtime-policy surface (spec §4)


jcfischer:
## Task

Implement `docs/work-graph.md` §4: the action-taxonomy policy document under the existing `soma policy` inspection surface (no parallel registry) — node-category → minimum-class floor (comms / credentials / main / self-governance ⇒ `approve`), never-below-floor clamp at createNode, asymmetric mutation (self-tighten free; loosening = identity-bound `approved`, fail-closed), and graph-mutation events as an inspection surface beside `governance_event`. Merges after the primitive lands on main (its enforcement point is the verb layer).

Node: `kind: task` · `autonomy: auto` (branch work; the merge rides normal PR review) · probes (close evidence): `bun test` exit 0, `bunx tsc --noEmit` exit 0, clamp + loosening-refusal covered by tests with recorded output.

jcfischer:
## Ruled out of scope — 2026-08-06

Closing unresolved. §4's autonomy floors have no consumer left, and §1 clause 3
says a gate that names none gets deleted rather than built.

**What changed.** [#549](https://github.com/the-metafactory/soma/pull/549)
removes the ratification requirement from HITL closes. `propose` and `approve`
are now behaviourally identical — both close on the closing session's say-so,
both may optionally use the proposal flow. The only distinction the runtime still
draws is `auto` versus not-`auto`: an `auto` node must declare probes and they
must pass.

So the taxonomy this node would implement — comms / credentials / main /
self-governance mapped to minimum classes, clamped never-below-floor — would
clamp between three classes of which two are the same thing. The floor's one
remaining bit is "may this be `auto`?", and nothing consumes even that: **nothing
runs unattended**. A node is only closed by a human-run session, so `auto` versus
HITL is a labelling choice made by the person creating the node, not a gate on
anyone.

**When it comes back.** The consumer of a floor is the **phase-2 headless tick**
(§5) — a scheduler closing `auto` nodes with nobody watching. There, "may this
category be closed unattended" is a real question with a real asker. If phase 2
is built, this returns with that consumer named, and probably in a smaller form
than §4 sketches: one bit about `auto`, not a four-category taxonomy.

**What is kept from §4 regardless.** The probe registry already lives on the
`soma policy` surface and is unaffected — it gates *whose code runs*, which is a
question at any headcount. The asymmetric-mutation rule (tightening additive,
loosening identity-bound and fail-closed) proved its worth there and is the
[...]

Issue #499, opened 2026-08-02
Orienteer re-base + ships-with-consumer merge to main


jcfischer:
## Task

Re-base the wayfinder doctrine as the bundled **orienteer** skill (`src/skills/orienteer/`, invoked `/orienteer`, #487): the walk procedure expressed in `soma graph` verbs instead of raw `gh`, plus the doctrine lines from #492 (scaffold nodes attach below their spawning ticket; work-kinds are orienteer vocabulary). Then land the **single ships-with-consumer merge**: primitive (#497) + verbs (#498) + orienteer arrive on main together — no reader, no merge (#484 clause 4). PR goes through sage review (ContextDrift caveat, sage#104).

HITL: the merge to main is the consuming act — principal calls it after review.

Node: `kind: task` · `autonomy: approve` (main write floor) · close evidence: merge commit SHA on main + `soma install` projecting the skill.

jcfischer:
### Finding from #496 (2026-08-02): orienteer should express edges natively

Native blocking is available on this tracker and writable by the provisioned Triage-role agent identity (see #496 probe 6, and the note on #497). The wayfinder doctrine this skill re-bases already prefers native blocking — "essential because it renders the frontier *visually* in the tracker's own UI" — and falls back to body conventions only for trackers that lack it.

So the re-based orienteer walk should create edges via `soma graph add`'s native path rather than teaching agents to write `Blocked by: #N` into issue bodies. The body-line convention stays only as a reader for legacy nodes.

jcfischer:
## Status — orienteer built, PR open, sage round 3 pending commit

**Not a resolution.** This node stays open: its close evidence is a merge commit
on `main`, and the merge is the principal's consuming act.

### Where things are

- **PR [#528](https://github.com/the-metafactory/soma/pull/528)** is open and
  mergeable; CI green (`Deterministic portability`, `gate / confidentiality-gate`).
  GitHub reads `REVIEW_REQUIRED` — a self-authored PR cannot collect an `APPROVE`
  from its own author's token, so sage posts as `COMMENTED` and the bus verdict is
  the authoritative one.
- **Branch head `5c5f2b7`** — last commit with a *verified* full suite
  (2212 pass / 0 fail, `bunx tsc --noEmit` clean).
- **orienteer** ships as `src/skills/orienteer/` — a 108-line router plus two
  Workflows and three references, projecting byte-identically into
  `~/.claude/skills/orienteer/` (verified against a temp home).

### Sage rounds

[...]

Issue #496, opened 2026-08-02
Provision the principal's agent identity (work-graph identity split)


jcfischer:
## Task

Provision the dedicated GitHub identity for work-graph execution, per `docs/work-graph.md` §5.1: create the machine-user account, grant it repo access, issue a fine-grained PAT (issues RW + contents R on target repos, no admin scopes), and store it where agent sessions resolve `GH_TOKEN` (env injection at session spawn — never in a repo or projection).

**Decisions taken while resolving (2026-08-02):**

- **Per-principal, not per-org.** The account is the external credential for *a principal's assistant* — `ivy-agent` for this principal; a second principal provisions their own by the same recipe. Soma already binds an assistant identity to a principal (`CONTEXT.md`); a shared org bot severs that binding exactly where it becomes auditable. It also keeps assignee-based claims (§2.4) distinguishable between sessions, and makes #502's authorized-ratifier binding 1:1 instead of 1:N — under a shared bot, principal B reacting to a proposal authored by principal A's session would pass an authorship-difference check and read `verified`, while §3.2 says *the principal's* 👍.
- **Login: `ivy-agent`** (availability confirmed 2026-08-02).
- **Access model: outside collaborator with Write, public repos only** — not org member. `the-metafactory` is on the Team plan with 2/2 seats filled; GitHub bills outside collaborators only on *private* repos, so a public-repo collaborator costs no seat. Narrower blast radius too: the account is not an org member and sees no private repo. Deviates from §5.1's literal "one seat, org-invited"; the contract it serves (dedicated machine user, assignable, fine-grained PAT, principal credential never in an agent session) is unchanged.
- **Accepted trade-offs:** private repos (`pilot`, `grove`, `forge`, `signal`) stay out of reach without upgrading to a seat. Outside collaborators cannot be added to org **teams**, so each agent identity needs its own per-repo grant — N principals × M repos of clicks, the price of the free route. 2FA custody is per-principal: lose the seed and that identity is unrecoverable by anyone else.
- **Assignability is verified, not assumed** — an outside collaborator with Write has push access and should appear in the repo's assignee list, but that is what the probe checks. If it fails, the fallback is org membership (a seat) and a re-probe.

HITL: account creation and PAT issuance are principal-only acts (credentials sit above the autonomy line). The agent side prepares the exact click-path checklist and verifies the result (`gh api user` under the agent token; assignee-capability probe on a scratch issue).

[...]

PR #549, opened 2026-08-06
fix(work-graph): a HITL node closes on the session's say-so

jcfischer:
`assertClosable` no longer requires ratified evidence on `approve`/`propose`
nodes. A HITL node closes when the closing session closes it.

## Why

The rule named no consumer (§1 clause 3). Where one person walks a map there is
nobody else to ratify, so requiring a second party's 👍 did not verify those
closes — it prevented them.
[#499](https://github.com/the-metafactory/soma/issues/499) is the worked example:
finished, merged, evidenced work that the gate refused, closed by hand with the
refusal recorded on the node.

## What changes

- **HITL nodes close on the session's say-so.** `--proposal-comment` is optional.
- **`--proposal-comment` is keyed on the flag, not on autonomy class**, so it
  takes effect on `auto` nodes too: a root-author 👎 refuses an `auto` close, and
  a ratification is recorded alongside the probe evidence that gates it.
- **`auto` nodes are unchanged** — they still need evidence, derived from probes
  that ran and passed.
- **Probes and the registry gate are untouched.**
- **A 👎 from the root author is surfaced** when `--proposal-comment` is
  supplied: a reminder, not a control. Its limits are documented and not a closed
  list — CLI-level only, absent on a bare close, not bound across proposals, and
  silent when the root author does not resolve.

## Verification

- `bun test` — 2221 pass / 0 fail. `bunx tsc --noEmit` — clean. CI green.
- Tests assert the new rule rather than around it: a HITL node closes with no
  ratification; an `auto` node without derived evidence still refuses; a
  root-author 👎 refuses even with a second account's 👍 present; the #499 shape
  (proposal posted, no reactions) now closes, recording `attestation: unverified`
  with "no proposal comment recorded" / "no ratification found"; the veto's
  one-close depth is pinned; and the `auto` + `--proposal-comment` behaviour is
  pinned in both directions.

## Review history

Seven sage rounds. The code came through unchanged; every round after the first
found defects in the **prose** — a citation pointing at a document arguing the
opposite of its claim, a "three limits, all deliberate" list that was neither, an
assertion about `auto` nodes standing in for a test. The final commit removes the
rationale rather than defending it further: −85 lines of argument, +28 of
description, no behaviour or test changed.

One objection is recorded rather than resolved: sage holds that #496's existing
`ivy-agent` authorship separation means the deadlock could be dissolved by
finishing the credential split (#510's remaining half) instead of removing the
gate. That is a real alternative and remains open — it is a decision about what
this primitive is for, not a defect in this diff.

## Supersedes

[#548](https://github.com/the-metafactory/soma/pull/548) (operator posture),
closed unmerged; `feat/graph-posture` and `feat/self-ratification` kept for
reference. Map nodes #500 and #511 were ruled out of scope for losing their
consumer with this change.

### DIFF
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -364,10 +364,12 @@ What remains:
   reactions through the receipt and binding superseded proposals, which is
   machinery this primitive does not want.
 
-- **The receipt is unchanged.** Proposal, ratifier, root authorship, confinement
-  and `attestation` are all still recorded, so a reader can tell a ratified close
-  from an unratified one. What changed is that an unratified close is *possible*,
-  not that it is *approved*.
+- **The receipt still distinguishes the two cases.** A ratified close records the
+  proposal and ratifier; a bare one records their *absence*, in
+  `attestationFacts.reasons` ("no proposal comment recorded", "no ratification
+  found"), alongside root authorship, confinement and `attestation`. Either way a
+  reader can tell which happened. What changed is that an unratified close is
+  *possible*, not that it is *approved*.
 
 **On multi-party deployments.** The gate is removed for *every* deployment, not
 only single-operator ones, and that is deliberate rather than incidental: "how
@@ -385,10 +387,10 @@ phase-2 auditor. Justifying the removal by appealing to an auditor that does not
 exist would be the same move this change is undoing: a gate defended by a
 consumer nobody can point at.
 
-Naming this precisely matters: an earlier draft claimed the receipt carried
-"proposal, ratifier, root authorship and confinement", which is what a *ratified*
-close records. On the default path the first two are absent, and their absence —
-not their presence — is the audit signal.
+Their absence — not their presence — is the audit signal on the default path.
+Stated twice in this section on purpose: an earlier revision asserted the
+ratified-close field list as if it applied to every close, and a correction that
+is appended rather than *applied* leaves a document saying both things.
 
 That is a real trade. A team wanting closes gated on a second party's approval
 does not get that from this primitive today; they get an auditable record that
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -370,10 +370,17 @@ What remains:
 only single-operator ones, and that is deliberate rather than incidental: "how
 many people are watching" is not derivable — two attempts to derive it failed
 review (#532) — so a rule conditioned on it would be a rule conditioned on a
-guess. What a multi-party deployment gets instead is the receipt: proposal,
-ratifier, root authorship and confinement are all still recorded, so an
-unratified close is *visible* rather than *prevented*, and a reviewer or the
-phase-2 auditor can act on it after the fact.
+guess. What a multi-party deployment gets instead is a receipt that **records the
+absence**. On a bare close no proposal is read, so there is no ratifier to name —
+and `attestationFacts.reasons` says exactly that: *"no proposal comment
+recorded"*, *"no ratification found"*, alongside the root authorship and
+confinement facts. An unratified close is therefore *visible* rather than
+*prevented*, and a reviewer or the phase-2 auditor can act on it after the fact.
+
+Naming this precisely matters: an earlier draft claimed the receipt carried
+"proposal, ratifier, root authorship and confinement", which is what a *ratified*
+close records. On the default path the first two are absent, and their absence —
+not their presence — is the audit signal.
 
 That is a real trade. A team wanting closes gated on a second party's approval
 does not get that from this primitive today; they get an auditable record that
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -338,12 +338,29 @@ quality ratifies when a downstream HITL node consumes the artifact (#485).
 
 ### 3.2 HITL (`propose` / `approve`)
 
-`approved` evidence is an attestation verifiable as coming from **a credential
-the agent does not hold**. Concretely (#486): the executing bot posts the
-resolution as a **proposal comment**; the receipt is the principal's 👍
-reaction on that specific comment ID — or a principal-authored comment, which
-wins when amending — verified via the API author field. A materially amended
-proposal is re-posted and needs fresh ratification (the replay-rebind lesson).
+A HITL node **closes on the closing session's say-so**. The human in the loop is
+the person running the session; requiring a second party to ratify named no
+consumer where one person walks the map, and the rule did not verify those closes
+— it prevented them (#499 is the worked example: finished, merged, evidenced work
+that the gate refused, protecting nobody).
+
+What remains:
+
+- **The proposal flow is available, not required.** `close --propose` posts a
+  proposal and `close --proposal-comment <id>` reads its reactions, for when a
+  second opinion is genuinely wanted. Ratification is admissible evidence; its
+  absence is not a defect.
+- **An explicit refusal still stops the close.** A 👎 from the graph root's
+  author refuses it. Dropping "must be approved" is not "may close over being
+  refused".
+- **The receipt is unchanged.** Proposal, ratifier, root authorship, confinement
+  and `attestation` are all still recorded, so a reader can tell a ratified close
+  from an unratified one. What changed is that an unratified close is *possible*,
+  not that it is *approved*.
+
+Where a receipt genuinely must be attestable — a headless phase-2 tick, a
+multi-party deployment — the machinery below still derives it. It is a label on
+the record, never a gate on the close.
 
 #### Deriving `attestation` (#502)
 
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -350,17 +350,20 @@ What remains:
   proposal and `close --proposal-comment <id>` reads its reactions, for when a
   second opinion is genuinely wanted. Ratification is admissible evidence; its
   absence is not a defect.
-- **An explicit refusal stops the close — within the proposal flow.** When
-  `--proposal-comment` is supplied, a 👎 from the graph root's author on that
-  comment refuses the close, naming who refused. Dropping "must be approved" is
-  not "may close over being refused".
+- **An explicit refusal is surfaced, not enforced.** When `--proposal-comment` is
+  supplied, a 👎 from the graph root's author on that comment refuses *that*
+  close, naming who refused.
 
-  Scope, stated precisely because the earlier draft overstated it: a refusal
-  lives *on a proposal*, so a bare `close` reads no reactions and cannot be
-  refused by one. Someone who was 👎'd and then closes without the flag is not
-  stopped by the runtime. That is a real limit, and the honest one for a
-  primitive whose gate is the human running it — the alternative is hunting prior
-  comments on every close, which is the machinery this change exists to remove.
+  It is a speed bump and the spec says so rather than overselling it. Three
+  limits, all deliberate: it lives in the CLI, not in `assertClosable`, so a
+  non-CLI consumer of the seam gets none of it; a bare `close` reads no reactions
+  and cannot be refused by one; and nothing binds a new proposal to a refused
+  one, so `--propose` on an unchanged body mints a fresh comment that closes
+  cleanly. It catches the honest case — told no, closed anyway by reflex — and
+  stops nobody who means to proceed. Making it real would mean threading
+  reactions through the receipt and binding superseded proposals, which is
+  machinery this primitive does not want.
+
 - **The receipt is unchanged.** Proposal, ratifier, root authorship, confinement
   and `attestation` are all still recorded, so a reader can tell a ratified close
   from an unratified one. What changed is that an unratified close is *possible*,
@@ -375,7 +378,12 @@ absence**. On a bare close no proposal is read, so there is no ratifier to name 
 and `attestationFacts.reasons` says exactly that: *"no proposal comment
 recorded"*, *"no ratification found"*, alongside the root authorship and
 confinement facts. An unratified close is therefore *visible* rather than
-*prevented*, and a reviewer or the phase-2 auditor can act on it after the fact.
+*prevented*, and a person reading the node can act on it after the fact.
+
+The consumer named here is a **human reading the record**, deliberately — not the
+phase-2 auditor. Justifying the removal by appealing to an auditor that does not
+exist would be the same move this change is undoing: a gate defended by a
+consumer nobody can point at.
 
 Naming this precisely matters: an earlier draft claimed the receipt carried
 "proposal, ratifier, root authorship and confinement", which is what a *ratified*
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -354,12 +354,14 @@ What remains:
   supplied, a 👎 from the graph root's author on that comment refuses *that*
   close, naming who refused.
 
-  It is a speed bump and the spec says so rather than overselling it. Three
-  limits, all deliberate: it lives in the CLI, not in `assertClosable`, so a
-  non-CLI consumer of the seam gets none of it; a bare `close` reads no reactions
-  and cannot be refused by one; and nothing binds a new proposal to a refused
-  one, so `--propose` on an unchanged body mints a fresh comment that closes
-  cleanly. It catches the honest case — told no, closed anyway by reflex — and
+  It is a speed bump and the spec says so rather than overselling it. Its limits,
+  all deliberate: it lives in the CLI, not in `assertClosable`, so a non-CLI
+  consumer of the seam gets none of it; a bare `close` reads no reactions and
+  cannot be refused by one; nothing binds a new proposal to a refused one, so
+  `--propose` on an unchanged body mints a fresh comment that closes cleanly; and
+  the predicate needs the root author to resolve, so where the root walk fails
+  nothing refuses — **silently**. Not a closed list: it is a reminder, and
+  anything that treats it as a control will be disappointed somewhere else too. It catches the honest case — told no, closed anyway by reflex — and
   stops nobody who means to proceed. Making it real would mean threading
   reactions through the receipt and binding superseded proposals, which is
   machinery this primitive does not want.
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -350,17 +350,35 @@ What remains:
   proposal and `close --proposal-comment <id>` reads its reactions, for when a
   second opinion is genuinely wanted. Ratification is admissible evidence; its
   absence is not a defect.
-- **An explicit refusal still stops the close.** A 👎 from the graph root's
-  author refuses it. Dropping "must be approved" is not "may close over being
-  refused".
+- **An explicit refusal stops the close — within the proposal flow.** When
+  `--proposal-comment` is supplied, a 👎 from the graph root's author on that
+  comment refuses the close, naming who refused. Dropping "must be approved" is
+  not "may close over being refused".
+
+  Scope, stated precisely because the earlier draft overstated it: a refusal
+  lives *on a proposal*, so a bare `close` reads no reactions and cannot be
+  refused by one. Someone who was 👎'd and then closes without the flag is not
+  stopped by the runtime. That is a real limit, and the honest one for a
+  primitive whose gate is the human running it — the alternative is hunting prior
+  comments on every close, which is the machinery this change exists to remove.
 - **The receipt is unchanged.** Proposal, ratifier, root authorship, confinement
   and `attestation` are all still recorded, so a reader can tell a ratified close
   from an unratified one. What changed is that an unratified close is *possible*,
   not that it is *approved*.
 
-Where a receipt genuinely must be attestable — a headless phase-2 tick, a
-multi-party deployment — the machinery below still derives it. It is a label on
-the record, never a gate on the close.
+**On multi-party deployments.** The gate is removed for *every* deployment, not
+only single-operator ones, and that is deliberate rather than incidental: "how
+many people are watching" is not derivable — two attempts to derive it failed
+review (#532) — so a rule conditioned on it would be a rule conditioned on a
+guess. What a multi-party deployment gets instead is the receipt: proposal,
+ratifier, root authorship and confinement are all still recorded, so an
+unratified close is *visible* rather than *prevented*, and a reviewer or the
+phase-2 auditor can act on it after the fact.
+
+That is a real trade. A team wanting closes gated on a second party's approval
+does not get that from this primitive today; they get an auditable record that
+one did not happen. If such a consumer appears, the gate returns as an opt-in
+naming them — which is what §1 clause 3 asks of any gate.
 
 #### Deriving `attestation` (#502)
 
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -373,9 +373,14 @@ What remains:
 
 **On multi-party deployments.** The gate is removed for *every* deployment, not
 only single-operator ones, and that is deliberate rather than incidental: "how
-many people are watching" is not derivable — two attempts to derive it failed
-review (#532) — so a rule conditioned on it would be a rule conditioned on a
-guess. What a multi-party deployment gets instead is a receipt that **records the
+many people are watching" is not derivable, so a rule conditioned on it would be
+a rule conditioned on a guess. Two attempts were made and both failed review;
+neither survives in a merged commit, so they are cited where they actually live:
+**graph-root authorship** (self-conferred — `findGraphRoot` returns the node
+itself when it has no parent) and **one reachable credential** (universal — true
+of any solo-token contributor), both on branch `feat/self-ratification` at
+`b8acfe4`, with the reasoning in the closing comment of
+[#548](https://github.com/the-metafactory/soma/pull/548). What a multi-party deployment gets instead is a receipt that **records the
 absence**. On a bare close no proposal is read, so there is no ratifier to name —
 and `attestationFacts.reasons` says exactly that: *"no proposal comment
 recorded"*, *"no ratification found"*, alongside the root authorship and
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -350,59 +350,24 @@ What remains:
   proposal and `close --proposal-comment <id>` reads its reactions, for when a
   second opinion is genuinely wanted. Ratification is admissible evidence; its
   absence is not a defect.
-- **An explicit refusal is surfaced, not enforced.** When `--proposal-comment` is
-  supplied, a 👎 from the graph root's author on that comment refuses *that*
-  close, naming who refused.
+- **An explicit refusal is surfaced, not enforced.** With `--proposal-comment`, a
+  👎 from the graph root's author on that comment refuses that close, naming who
+  refused.
 
-  It is a speed bump and the spec says so rather than overselling it. Its limits,
-  all deliberate: it lives in the CLI, not in `assertClosable`, so a non-CLI
-  consumer of the seam gets none of it; a bare `close` reads no reactions and
-  cannot be refused by one; nothing binds a new proposal to a refused one, so
-  `--propose` on an unchanged body mints a fresh comment that closes cleanly; and
-  the predicate needs the root author to resolve, so where the root walk fails
-  nothing refuses — **silently**. Not a closed list: it is a reminder, and
-  anything that treats it as a control will be disappointed somewhere else too. It catches the honest case — told no, closed anyway by reflex — and
-  stops nobody who means to proceed. Making it real would mean threading
-  reactions through the receipt and binding superseded proposals, which is
-  machinery this primitive does not want.
+  A reminder, not a control, and its limits are not a closed list: it lives in
+  the CLI rather than `assertClosable`, so a non-CLI consumer of the seam gets
+  none of it; a bare `close` reads no reactions; nothing binds a new proposal to
+  a refused one; and it needs the root author to resolve, so where the root walk
+  fails nothing refuses, silently.
 
-- **The receipt still distinguishes the two cases.** A ratified close records the
-  proposal and ratifier; a bare one records their *absence*, in
+- **The receipt distinguishes the two cases.** A ratified close records the
+  proposal and ratifier; a bare one records their absence in
   `attestationFacts.reasons` ("no proposal comment recorded", "no ratification
-  found"), alongside root authorship, confinement and `attestation`. Either way a
-  reader can tell which happened. What changed is that an unratified close is
-  *possible*, not that it is *approved*.
+  found"), alongside root authorship, confinement and `attestation`.
 
-**On multi-party deployments.** The gate is removed for *every* deployment, not
-only single-operator ones, and that is deliberate rather than incidental: "how
-many people are watching" is not derivable, so a rule conditioned on it would be
-a rule conditioned on a guess. Two attempts were made and both failed review;
-neither survives in a merged commit, so they are cited where they actually live:
-**graph-root authorship** (self-conferred — `findGraphRoot` returns the node
-itself when it has no parent) and **one reachable credential** (universal — true
-of any solo-token contributor), both on branch `feat/self-ratification` at
-`b8acfe4`, with the reasoning in the closing comment of
-[#548](https://github.com/the-metafactory/soma/pull/548). What a multi-party deployment gets instead is a receipt that **records the
-absence**. On a bare close no proposal is read, so there is no ratifier to name —
-and `attestationFacts.reasons` says exactly that: *"no proposal comment
-recorded"*, *"no ratification found"*, alongside the root authorship and
-confinement facts. An unratified close is therefore *visible* rather than
-*prevented*, and a person reading the node can act on it after the fact.
-
-The consumer named here is a **human reading the record**, deliberately — not the
-phase-2 auditor. Justifying the removal by appealing to an auditor that does not
-exist would be the same move this change is undoing: a gate defended by a
-consumer nobody can point at.
-
-Their absence — not their presence — is the audit signal on the default path.
-Stated twice in this section on purpose: an earlier revision asserted the
-ratified-close field list as if it applied to every close, and a correction that
-is appended rather than *applied* leaves a document saying both things.
-
-That is a real trade. A team wanting closes gated on a second party's approval
-does not get that from this primitive today; they get an auditable record that
-one did not happen. If such a consumer appears, the gate returns as an opt-in
-naming them — which is what §1 clause 3 asks of any gate.
+The removal is unconditional — it is not scoped to single-operator deployments.
+A deployment wanting closes gated on a second party's approval does not get that
+from this primitive; it gets a record of whether one happened.
 
 #### Deriving `attestation` (#502)
 
--- a/src/cli/graph.ts
+++ b/src/cli/graph.ts
@@ -786,8 +786,15 @@ async function runClose(
   // A proposal comment is optional now: a HITL node closes on the session's
   // say-so (§3.2). When one IS supplied, its reactions still carry weight —
   // a ratification as admissible evidence, and a root-author 👎 as a refusal.
+  //
+  // Keyed on the comment id alone, deliberately. `assertClosable` no longer
+  // distinguishes `propose` from `approve` from anything else non-`auto`, so a
+  // CLI-side autonomy test would be the only place in the system still drawing a
+  // line the core does not — a distinction living in one consumer is worse than
+  // no distinction at all. Supplying a proposal on an `auto` node is unusual
+  // rather than wrong: its reactions are read on the same terms.
   const commentId = parsed.options.proposalComment;
-  if (state.node.autonomy !== "auto" && commentId !== undefined) {
+  if (commentId !== undefined) {
     const proposalRef: CommentRef = await graph.readComment({ id: commentId, nodeId: ref.id });
     proposal = { commentId: proposalRef.id, author: proposalRef.author ?? "" };
     const reactions = await graph.readCommentReactions(proposalRef);
--- a/src/cli/graph.ts
+++ b/src/cli/graph.ts
@@ -792,10 +792,15 @@ async function runClose(
     proposal = { commentId: proposalRef.id, author: proposalRef.author ?? "" };
     const reactions = await graph.readCommentReactions(proposalRef);
 
-    // Ratification is no longer required, but an explicit refusal is still
-    // honoured. Dropping "you must be approved" must not become "you may ignore
-    // being refused" — a 👎 from the map's owner is the one reaction that
-    // unambiguously means *no*, and it is cheap to respect.
+    // Ratification is no longer required; an explicit refusal is still surfaced.
+    //
+    // Deliberately a speed bump, not a control, and it lives here in the CLI
+    // rather than in `assertClosable` because it is not a contract rule. It
+    // catches the honest case — you were told no and closed anyway by reflex —
+    // and it stops nobody who means to proceed: a fresh `--propose` mints a new
+    // comment with no reactions, and passing that id closes cleanly. Binding
+    // proposals to superseded ones, or threading reactions through the receipt
+    // so the core could enforce it, is machinery this primitive does not want.
     const vetoed = reactions.some(
       (reaction) => reaction.content === "-1" && root?.author !== undefined && reaction.author === root.author,
     );
@@ -803,7 +808,7 @@ async function runClose(
       throw new SomaCliError(
         [
           `Node ${ref.id} was refused: ${root?.author ?? "the graph root's author"} left a 👎 on proposal comment ${proposal.commentId}.`,
-          `Address it and post a new proposal rather than closing over the refusal.`,
+          `Nothing here can stop you closing anyway — this is a reminder, not a gate.`,
         ].join("\n"),
         1,
       );
--- a/src/cli/graph.ts
+++ b/src/cli/graph.ts
@@ -783,18 +783,11 @@ async function runClose(
   let proposal: { commentId: string; author: string } | undefined;
   let ratification: { kind: "reaction" | "comment"; id: string; author: string } | undefined;
 
-  if (state.node.autonomy !== "auto") {
-    const commentId = parsed.options.proposalComment;
-    if (commentId === undefined) {
-      throw new SomaCliError(
-        [
-          `Node ${ref.id} is ${state.node.autonomy}-class: it closes on a ratified proposal, not on a self-report (§3.2).`,
-          `Post one with:  soma graph close ${ref.id} --propose --body-file <path>`,
-          `Then close with: soma graph close ${ref.id} --proposal-comment <id>`,
-        ].join("\n"),
-        1,
-      );
-    }
+  // A proposal comment is optional now: a HITL node closes on the session's
+  // say-so (§3.2). When one IS supplied, its reactions still carry weight —
+  // a ratification as admissible evidence, and a root-author 👎 as a refusal.
+  const commentId = parsed.options.proposalComment;
+  if (state.node.autonomy !== "auto" && commentId !== undefined) {
     const proposalRef: CommentRef = await graph.readComment({ id: commentId, nodeId: ref.id });
     proposal = { commentId: proposalRef.id, author: proposalRef.author ?? "" };
     const reactions = await graph.readCommentReactions(proposalRef);
--- a/src/cli/graph.ts
+++ b/src/cli/graph.ts
@@ -798,6 +798,24 @@ async function runClose(
     const proposalRef: CommentRef = await graph.readComment({ id: commentId, nodeId: ref.id });
     proposal = { commentId: proposalRef.id, author: proposalRef.author ?? "" };
     const reactions = await graph.readCommentReactions(proposalRef);
+
+    // Ratification is no longer required, but an explicit refusal is still
+    // honoured. Dropping "you must be approved" must not become "you may ignore
+    // being refused" — a 👎 from the map's owner is the one reaction that
+    // unambiguously means *no*, and it is cheap to respect.
+    const vetoed = reactions.some(
+      (reaction) => reaction.content === "-1" && root?.author !== undefined && reaction.author === root.author,
+    );
+    if (vetoed) {
+      throw new SomaCliError(
+        [
+          `Node ${ref.id} was refused: ${root?.author ?? "the graph root's author"} left a 👎 on proposal comment ${proposal.commentId}.`,
+          `Address it and post a new proposal rather than closing over the refusal.`,
+        ].join("\n"),
+        1,
+      );
+    }
+
     ratification = selectRatification(reactions, proposal.author, root?.author);
     if (ratification !== undefined) {
       evidence.push({
--- a/src/skills/orienteer/references/closing.md
+++ b/src/skills/orienteer/references/closing.md
@@ -101,10 +101,14 @@ It is a tool, not a toll. An earlier version required that 👍 before any
 the map: there was nobody else to ratify, so the rule did not verify closes, it
 prevented them.
 
-**One reaction still stops a close.** A 👎 from the graph root's author is an
-explicit refusal, and the close refuses with it — dropping "you must be approved"
-does not mean "you may close over being refused". Address it and post a new
-proposal.
+**One reaction still stops a close — if you use the flow.** A 👎 from the graph
+root's author on a proposal refuses the close when you pass
+`--proposal-comment`: dropping "you must be approved" does not mean "you may
+close over being refused". Address it and post a new proposal.
+
+A bare `close` reads no reactions, so it cannot be refused by one. That is not a
+loophole to use — if someone 👎'd your proposal, closing around it produces a
+receipt that records no refusal ever happened.
 
 Two things worth knowing when you do use the proposal flow:
 
--- a/src/skills/orienteer/references/closing.md
+++ b/src/skills/orienteer/references/closing.md
@@ -102,7 +102,9 @@ the map: there was nobody else to ratify, so the rule did not verify closes, it
 prevented them.
 
 **A 👎 is surfaced, not enforced.** If you pass `--proposal-comment` and the
-graph root's author left a 👎 on it, the close refuses and names them. Treat it
+graph root's author left a 👎 on it, the close refuses and names them — provided
+the root author could be resolved at all. Where the root walk fails, no reaction
+matches and nothing refuses, silently. Treat it
 as a reminder: it catches you closing by reflex after being told no, and it stops
 nothing else. A bare `close` reads no reactions at all, and re-proposing produces
 a fresh comment that closes cleanly — nothing binds it to the refused one.
--- a/src/skills/orienteer/references/closing.md
+++ b/src/skills/orienteer/references/closing.md
@@ -101,14 +101,15 @@ It is a tool, not a toll. An earlier version required that 👍 before any
 the map: there was nobody else to ratify, so the rule did not verify closes, it
 prevented them.
 
-**One reaction still stops a close — if you use the flow.** A 👎 from the graph
-root's author on a proposal refuses the close when you pass
-`--proposal-comment`: dropping "you must be approved" does not mean "you may
-close over being refused". Address it and post a new proposal.
+**A 👎 is surfaced, not enforced.** If you pass `--proposal-comment` and the
+graph root's author left a 👎 on it, the close refuses and names them. Treat it
+as a reminder: it catches you closing by reflex after being told no, and it stops
+nothing else. A bare `close` reads no reactions at all, and re-proposing produces
+a fresh comment that closes cleanly — nothing binds it to the refused one.
 
-A bare `close` reads no reactions, so it cannot be refused by one. That is not a
-loophole to use — if someone 👎'd your proposal, closing around it produces a
-receipt that records no refusal ever happened.
+So the honest reading is that being refused is *recorded*, not *prevented*. If
+you close around a 👎, the receipt shows a close with no refusal in it, and the
+person who said no finds out by reading the node.
 
 Two things worth knowing when you do use the proposal flow:
 
--- a/src/skills/orienteer/references/closing.md
+++ b/src/skills/orienteer/references/closing.md
@@ -103,11 +103,12 @@ prevented them.
 
 **A 👎 is surfaced, not enforced.** If you pass `--proposal-comment` and the
 graph root's author left a 👎 on it, the close refuses and names them — provided
-the root author could be resolved at all. Where the root walk fails, no reaction
-matches and nothing refuses, silently. Treat it
-as a reminder: it catches you closing by reflex after being told no, and it stops
-nothing else. A bare `close` reads no reactions at all, and re-proposing produces
-a fresh comment that closes cleanly — nothing binds it to the refused one.
+the root author resolves at all; where the root walk fails, nothing refuses and
+nobody is named.
+
+Treat it as a reminder: it catches you closing by reflex after being told no, and
+stops nothing else. A bare `close` reads no reactions, and re-proposing produces a
+fresh comment that closes cleanly — nothing binds it to the refused one.
 
 So the honest reading is that being refused is *recorded*, not *prevented*. If
 you close around a 👎, the receipt shows a close with no refusal in it, and the
--- a/src/skills/orienteer/references/closing.md
+++ b/src/skills/orienteer/references/closing.md
@@ -80,30 +80,40 @@ downstream HITL node consumes the artifact.
 
 ## HITL close (`propose` / `approve`)
 
-Two phases, one verb:
+A HITL node **closes when you close it**. There is no ratification requirement:
+you are the human in the loop, and you are present.
+
+```bash
+soma graph close <id>
+```
+
+The two-phase flow remains for when a second opinion is actually wanted — a
+decision you want someone else to sign off, or a record of who agreed:
 
 ```bash
 soma graph close <id> --propose --body-file <path>   # posts the proposal, closes nothing
-#  … the principal reacts 👍 on that comment …
+#  … someone reacts 👍 on that comment …
 soma graph close <id> --proposal-comment <comment-id>
 ```
 
-The receipt is the ratification, read from the API's author field — never from
-body text. A 👎 from the graph root's author suppresses ratification outright.
+It is a tool, not a toll. An earlier version required that 👍 before any
+`approve`-class node could close, which named no consumer when one person walks
+the map: there was nobody else to ratify, so the rule did not verify closes, it
+prevented them.
 
-Two things the runtime does **not** do here, both of which are yours to hold:
+**One reaction still stops a close.** A 👎 from the graph root's author is an
+explicit refusal, and the close refuses with it — dropping "you must be approved"
+does not mean "you may close over being refused". Address it and post a new
+proposal.
 
-- **Any non-proposer's 👍 closes the node.** The ratifier is preferred to be the
-  graph root's author, but the fallback is the first other reaction — so on a
-  public tracker a stranger's thumb produces `approved` evidence and the node
-  closes. Only the `attestation` label records that it was not the right human.
-  Read root-author approval as the *thing you must obtain*, not as a condition
-  the verb enforces.
+Two things worth knowing when you do use the proposal flow:
+
+- **Any non-proposer's 👍 is taken as the ratification**, root author preferred.
+  On a public tracker that includes a stranger's. Only `attestation` records
+  that it was not the right human.
 - **Ratification binds to a comment id, not to its text.** Nothing hashes the
-  proposal body, so a proposal that is ratified and *then edited* still closes
-  on that 👍. "A materially amended proposal is re-posted and needs fresh
-  ratification" is doctrine you follow, not a property the runtime checks: if
-  the resolution changes materially, post a new comment and get a new 👍.
+  proposal body, so a proposal that is ratified and *then edited* still closes on
+  that 👍. If the resolution changes materially, post a new comment.
 
 ## Attestation
 
--- a/src/work-graph.ts
+++ b/src/work-graph.ts
@@ -633,34 +633,11 @@ export function assertClosable(node: WorkGraphNode, receipt: CloseReceipt): void
   // Evidence is required of `auto` nodes only, and there it costs nothing: the
   // `probed` entry is derived from probes that already ran and passed.
   //
-  // HITL nodes close on the session's say-so. Requiring a ratified proposal here
-  // was the original rule, and it was wrong for what this primitive is: a way to
-  // structure work whose route is unclear, walked by a human who is *present*.
-  // A gate naming no consumer is ceremony (§1 clause 3), and where one person
-  // walks a map the ratification gate named none — nobody else could ratify, so
-  // it did not verify the close, it only prevented it. #499 is the worked
-  // example: finished, merged, evidenced work that the gate refused, protecting
-  // no one.
-  //
-  // The removal is UNCONDITIONAL, not scoped to single-operator deployments, and
-  // that is the deliberate part. "How many people are watching" is not derivable
-  // — two attempts to derive it failed review — so conditioning the gate on it
-  // would condition it on a guess. A multi-party deployment instead gets a
-  // receipt that RECORDS THE ABSENCE: on a bare close no proposal is read, so
-  // there is no ratifier to name, and `attestationFacts.reasons` says exactly
-  // that — "no proposal comment recorded", "no ratification found". Their
-  // absence is the audit signal, not their presence. An unratified close is
-  // therefore visible rather than prevented, and a person reading the node can
-  // act on it. If a consumer appears who needs it gated, it returns as an
-  // opt-in naming them.
-  //
-  // The receipt still records everything it recorded before — proposal, ratifier,
-  // root authorship, confinement, `attestation`. Nothing is hidden; a reader can
-  // still tell a ratified close from an unratified one. What changed is that an
-  // unratified close is now *possible*, not *approved*.
-  //
-  // The two-phase `--propose` flow remains for when a second opinion is wanted.
-  // It is a tool, no longer a toll.
+  // HITL nodes close on the session's say-so, unconditionally — not scoped to
+  // single-operator deployments. The receipt still distinguishes the two cases:
+  // a bare close records the absence of a proposal and ratification in
+  // `attestationFacts.reasons`, so an unratified close is visible rather than
+  // prevented. See §3.2.
   if (node.autonomy !== "auto") return;
 
   const admissible = agentExternalEvidenceKinds(node.autonomy);
--- a/src/work-graph.ts
+++ b/src/work-graph.ts
@@ -630,6 +630,27 @@ export function assertClosable(node: WorkGraphNode, receipt: CloseReceipt): void
     }
   }
 
+  // Evidence is required of `auto` nodes only, and there it costs nothing: the
+  // `probed` entry is derived from probes that already ran and passed.
+  //
+  // HITL nodes close on the session's say-so. Requiring a ratified proposal here
+  // was the original rule, and it was wrong for what this primitive is: a way to
+  // structure work whose route is unclear, walked by a human who is *present*.
+  // A gate naming no consumer is ceremony (§1 clause 3), and on a
+  // single-operator deployment the ratification gate named none — there was
+  // nobody else to ratify, so it did not verify the close, it only prevented it.
+  // #499 is the worked example: a finished, merged, evidenced piece of work that
+  // the gate refused, and the refusal protected no one.
+  //
+  // The receipt still records everything it recorded before — proposal, ratifier,
+  // root authorship, confinement, `attestation`. Nothing is hidden; a reader can
+  // still tell a ratified close from an unratified one. What changed is that an
+  // unratified close is now *possible*, not *approved*.
+  //
+  // The two-phase `--propose` flow remains for when a second opinion is wanted.
+  // It is a tool, no longer a toll.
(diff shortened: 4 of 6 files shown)


jcfischer (commented):

11 finding(s): 2 blocker, 3 important, 4 suggestion, 2 nit.


### CodeQuality

- **[blocker]** `src/skills/orienteer/references/closing.md:84` — **Documented bare close still refused by the CLI**
  closing.md now instructs `soma graph close <id>` for HITL nodes ("A HITL node **closes when you close it**"), but `runClose` is unchanged above the diff: for `state.node.autonomy !== "auto"` a missing `--proposal-comment` still throws "Node X is approve-class: it closes on a ratified proposal, not on a self-report (§3.2)" — and test/cli-graph.test.ts:381 still asserts that refusal. Only `assertClosable` was relaxed; the CLI gate that #499 actually hit remains.
  Fix: Drop the `commentId === undefined` refusal in runClose (make `--proposal-comment` optional for non-auto nodes), update the stale test at cli-graph.test.ts:381, and add a test for `graph close <id>` with no proposal at all.

- **[suggestion]** `src/work-graph.ts:652` — **assertClosable JSDoc still states clause 3 unconditionally**
  `if (node.autonomy !== "auto") return;` makes clause 3 of the function's own contract — "at least one agent-external evidence entry carries a pointer someone else can check (§3.1)" — false for HITL nodes, but the JSDoc directly above is untouched. The contract a caller reads now contradicts the code.
  Fix: Scope clause 3 to `auto` nodes in the JSDoc; the 20-line inline rationale below belongs in the doc/commit, not duplicated in three files.

- **[nit]** `src/cli/graph.ts:807` — **Unreachable fallback in the veto message**
  The predicate requires `root?.author !== undefined`, so when `vetoed` is true `root.author` is always defined and `${root?.author ?? "the graph root's author"}` can never take its fallback branch.
  Fix: Hoist `const owner = root?.author;` out of the predicate, test `owner !== undefined && reaction.author === owner`, and interpolate `owner` directly.

### Architecture

- **[important]** `src/cli/graph.ts:803` — **Veto enforced in CLI, not in the core close gate**
  `assertClosable` now does `if (node.autonomy !== "auto") return;`, and the only thing that stops a refused close is the `const vetoed = reactions.some(...)` / `throw new SomaCliError` block in `runClose`. CONTEXT.md's work-graph entry says the runtime enforces the autonomy class "at close time" over "a typed `GraphStore` seam"; any non-CLI consumer of the seam (library mode, MCP surface, the phase-2 auditor) now sees a HITL node as unconditionally closable and silently loses the 👎 refusal.
  Fix: Move the refusal check into `assertClosable`, deriving it from the receipt's recorded ratifier/root-author fields; keep the CLI block only for the human-readable message.

- **[suggestion]** `src/work-graph.ts:650` — **`propose` and `approve` now behave identically at close**
  `if (node.autonomy !== "auto") return;` collapses the three-value autonomy class to two runtime behaviours, while CONTEXT.md still states "A node carries an autonomy class (`auto` / `propose` / `approve`)" and that "the spec obligates the runtime to enforce the autonomy class at close time". The exported type keeps a distinction nothing now consumes.
  Fix: State in `docs/work-graph.md` what distinguishes `propose` from `approve` post-change, or fold them into one value.

### Maintainability

- **[important]** `src/cli/graph.ts:803` — **Veto predicate duplicates selectRatification's suppression guard**
  The new `reactions.some((reaction) => reaction.content === "-1" && root?.author !== undefined && reaction.author === root.author)` restates the identical rule already in `selectRatification` (`reaction.content === "-1" && reaction.author === rootAuthor`, graph.ts:660); since the new block throws first, that older guard is now unreachable from its only production caller while its doc comment still claims it suppresses ratification.
  Fix: Extract `export function rootAuthorVetoed(reactions, rootAuthor)`, call it from the new throw site, and delete the now-dead guard (and its stale comment) from `selectRatification`. Hoist `root?.author !== undefined` out of the `.some()` callback.

- **[suggestion]** `src/work-graph.ts:650` — **18-line rationale comment on a one-line guard, duplicated in four other files**
  The comment block ending "A gate naming no consumer is ceremony (§1 clause 3)... #499 is the worked example" guards a single statement `if (node.autonomy !== "auto") return;`, and the same argument is re-told in docs/work-graph.md, closing.md, the cli/graph.ts comment, and two test comments. Five copies of one rationale drift independently on the next change.
  Fix: Keep two or three lines here stating the rule plus a pointer to docs/work-graph.md; let the doc own the argument.

- **[nit]** `test/cli-graph.test.ts:464` — **Test name no longer describes the mechanism it exercises**
  The name `"a 👎 from the root author blocks ratification, so the close is refused"` is unchanged, but the assertion moved from `"needs at least one approved evidence"` to `expect(message).toContain("was refused")` — the refusal is now an explicit veto throw, not an absent ratification.
  Fix: Rename to "a 👎 from the root author refuses the close outright".

### HonestOracle

- **[blocker]** `src/cli/graph.ts:803` — **Veto enforced only when a proposal comment is passed**
  The new `vetoed` check sits inside the branch that reads `reactions`, i.e. only when `--proposal-comment <id>` is supplied, yet docs/work-graph.md states unqualified: "**An explicit refusal still stops the close.** A 👎 from the graph root's author refuses it." A closer facing a 👎 just runs the bare `soma graph close <id>` that closing.md now documents, and no reaction is ever read.

- **[important]** `src/work-graph.ts:650` — **Single-operator rationale, unconditional removal**
  The in-code justification is scoped — "on a single-operator deployment the ratification gate named none" — but `if (node.autonomy !== "auto") return;` drops the gate for every deployment, and the PR states two attempts to derive "single operator" failed. docs/work-graph.md names the counter-case ("a multi-party deployment") only to wave it off as "a label on the record, never a gate on the close", without engaging what enforces the gate there.

- **[suggestion]** `test/cli-graph.test.ts:478` — **Headline path untested; test routes through proposal flow**
  The test named "without a 👎, a HITL node closes on the session's say-so" closes via `--propose` then `--proposal-comment c1`, not the bare `soma graph close <id>` that closing.md presents as the way a HITL node now closes. The PR's "Rewritten tests assert the new rule" therefore covers a proxy path, and leaves the veto-free path unexercised.


---

jcfischer (commented):

9 finding(s): 4 important, 4 suggestion, 1 nit; 1 lens(es) failed to run: Maintainability.


### CodeQuality

- **[suggestion]** `src/work-graph.ts:651` — **Guard returns from the whole function, not just the evidence check**
  `if (node.autonomy !== "auto") return;` exits `assertClosable` entirely, so every check that lives below it — now and later — is skipped for HITL nodes, even ones unrelated to ratification. Scoping the skip to the evidence block keeps the exemption to what the rationale actually argues for.
  Fix: Wrap the external-evidence check in `if (node.autonomy === "auto") { ... }` instead of early-returning.

- **[nit]** `test/work-graph.test.ts:218` — **Test name promises more than the assertion checks**
  `test("an auto node still needs evidence its probes actually produced")` asserts only on `receipt({ evidence: [] })` — an empty list, not an entry that failed the summary/pointer derivation. The surviving gate's actual shape (non-empty summary and pointer) goes unexercised.
  Fix: Either rename to "an auto node still needs evidence", or add a case with an `approved`/`probed` entry carrying a blank pointer.

### Architecture

- **[important]** `docs/work-graph.md:338` — **Spec deletes the definition of `approved`, which is still live**
  The hunk removes "`approved` evidence is an attestation verifiable as coming from **a credential the agent does not hold**" and the #486 mechanics, but `approved` remains a receipt evidence kind (`ratification = selectRatification(...); if (ratification !== undefined) { evidence.push({`) and the "#### Deriving `attestation` (#502)" section directly below still derives from it. The canonical spec now names a term it no longer defines.
  Fix: Keep the `approved` definition paragraph under "What remains", qualified with "admissible, no longer required".

- **[suggestion]** `docs/work-graph.md:341` — **CONTEXT.md work-graph entry diverges from the spec it cites**
  CONTEXT.md § `work graph` states "The spec obligates the runtime to enforce the autonomy class at close time" and cites `docs/work-graph.md` as the spec; the spec now says "A HITL node **closes on the closing session's say-so**", leaving enforcement on `auto` only. The glossary definition of the primitive is no longer what the spec says.
  Fix: Scope the close-time enforcement clause in CONTEXT.md § `work graph` to `auto` nodes.

- **[suggestion]** `docs/work-graph.md:340` — **Non-canonical actor nouns in docs/ and src/**
  CONTEXT.md § `principal` locks the term: "Do not use `user`, `owner`, `human`, `operator` in glossary or formal docs... Never in `docs/`, `src/`." The new rule is stated as "The human in the loop is the person running the session", with "on a single-operator deployment" in the `src/work-graph.ts` comment — so the actor the whole change turns on is named in three killed synonyms.
  Fix: Say "the closing session's principal".

### Maintainability — DID NOT RUN

- **[important]** `(lens output):0` — **Maintainability: model deviated from JSON contract**
  claude extraction failed (exit-nonzero) exit=1
--- output ---
{"is_error":true,"duration_api_ms":0,"num_turns":1,"stop_reason":"stop_sequence","session_id":"8881808d-e161-4634-b033-8846088b61e4","total_cost_usd":0,"usage":{"input_tokens":0,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":0,"server_tool_use":{"web_search_requests":0,"web_fetch_requests":0},"service_tier":"standard","cache_creation":{"ephemeral_1h_input_tokens":0,"ephemeral_5m_input_tokens":0},"inference_geo":"","iterations":[],"speed":"standard"},"modelUsage":{},"permission_denials":[],"terminal_reason":"api_error","fast_mode_state":"off","fast_mode_disabled_reason":"sdk_opt_in_required","subtype":"success","api_error_status":529,"result":"API Error: 529 Overloaded. This is a server-side issue, usually temporary — try again in a moment. If it persists, check https://status.claude.com.","type":"result","duration_ms":245606,"uuid":"805777d1-f08b-4d44-86b4-de60a9964b56"}


### HonestOracle

- **[important]** `docs/work-graph.md:338` — **"One change" hides doctrine deleted from the canonical doc**
  The description asserts "One change: `assertClosable` no longer requires ratified evidence", but this hunk also deletes the #486 definition of `approved` evidence ("verifiable as coming from **a credential the agent does not hold** ... verified via the API author field") and the replay-rebind rule ("A materially amended proposal is re-posted and needs fresh ratification"), neither mentioned anywhere in the PR. A reviewer approving on "one change" never audits the removed normative rules, which still govern the proposal flow the PR says survives.

- **[important]** `PR description:1` — **"Nothing gates on verified" asserted without evidence, re-scopes #510/#511**
  "Nothing gates on `verified` any more, so they are optionality rather than route" is a repo-wide absence claim that a six-file diff touching only the HITL branch of `assertClosable` cannot establish — no grep, no test, no citation is offered. The conclusion is load-bearing: it downgrades map nodes #510 and #511 (per test/work-graph.test.ts:203, #511 is credential confinement) from route to optionality.

- **[suggestion]** `docs/work-graph.md:357` — **Derived label offered where attestability is called mandatory**
  "Where a receipt genuinely must be attestable — a headless phase-2 tick, a multi-party deployment — the machinery below still derives it. It is a label on the record, never a gate on the close" contradicts itself in one paragraph: if a receipt *must* be attestable, deriving a label does not make it so. The diff leaves a multi-party deployment no mechanism to require ratification, only to record its absence.


---

jcfischer (commented):

11 finding(s): 1 blocker, 1 important, 7 suggestion, 2 nit.


### CodeQuality

- **[suggestion]** `src/cli/graph.ts:801` — **Veto silently no-ops when the root author is unresolved**
  `reaction.content === "-1" && root?.author !== undefined && reaction.author === root.author` makes `vetoed` false whenever the root's author could not be read, so a real 👎 passes unnoticed instead of the runtime saying it could not evaluate the refusal.
  Fix: If `root?.author` is undefined while a proposal comment was supplied, fail or warn rather than treating the refusal check as satisfied.

- **[suggestion]** `src/cli/graph.ts:789` — **`--proposal-comment` silently ignored on auto nodes**
  `if (state.node.autonomy !== "auto" && commentId !== undefined)` drops a supplied proposal comment for `auto` nodes with no message; now that the flag is optional rather than mandatory-for-HITL, a user passing it to the wrong node gets a close whose receipt names no proposal and no explanation why.
  Fix: Reject `--proposal-comment` on an `auto` node with an explicit error.

- **[nit]** `test/cli-graph.test.ts:393` — **Optional chaining makes the proposal assertion vacuous**
  `expect(store.closed[0].receipt.attestationFacts?.proposal).toBeUndefined()` passes both when `proposal` is absent and when `attestationFacts` itself is missing, so it cannot distinguish "no proposal recorded" from "no facts recorded at all".
  Fix: Assert `attestationFacts` is defined first, then that `.proposal` is undefined.

### Architecture

- **[suggestion]** `src/work-graph.ts:656` — **`autonomy` no longer discriminates behaviour at the close gate**
  `if (node.autonomy !== "auto") return;` collapses `propose` and `approve` into one branch, so the three-valued autonomy class now carries only a two-valued distinction in the only place the spec says it is enforced ("The spec obligates the runtime to enforce the autonomy class at close time", CONTEXT.md §work graph). Reintroducing a per-class rule later means re-deriving the split rather than extending an existing one.
  Fix: Keep the branch keyed on the class explicitly (e.g. a switch over `auto` / `propose` / `approve` with the two HITL arms returning) so a future per-class rule has a seam to land in.

### Maintainability

- **[suggestion]** `src/work-graph.ts:633` — **24-line comment guards a one-line return**
  The block from "Evidence is required of `auto` nodes only" through "It is a tool, no longer a toll." is 24 comment lines above `if (node.autonomy !== "auto") return;`. Five paragraphs of decision history in a function body means future readers scroll past design rationale to find control flow.
  Fix: Keep 2-3 lines here ("HITL nodes close on the session's say-so — see docs/work-graph.md §3.2 / #499") and let the doc carry the argument.

- **[suggestion]** `src/skills/orienteer/references/closing.md:95` — **Same rationale restated in four places**
  "named no consumer when one person walks the map: there was nobody else to ratify, so the rule did not verify closes, it prevented them" is a near-verbatim echo of docs/work-graph.md ("requiring a second party to ratify named no consumer where one person walks the map") and src/work-graph.ts ("where one person walks a map the ratification gate named none"). Four copies drift independently on the next revision.
  Fix: Pick docs/work-graph.md as canonical; have closing.md and the code comments link to it rather than restate it.

- **[nit]** `src/cli/graph.ts:786` — **Two comment blocks say the same thing in one branch**
  The branch header comment ("a ratification as admissible evidence, and a root-author 👎 as a refusal") and the inner comment above `vetoed` ("Ratification is no longer required, but an explicit refusal is still honoured") state the same rule twice within eleven lines.
  Fix: Drop the inner block; the header already covers both halves.

### HonestOracle

- **[blocker]** `docs/work-graph.md:370` — **Multi-party mitigation claims receipt fields the default path never writes**
  The spec tells multi-party deployments "What a multi-party deployment gets instead is the receipt: proposal, ratifier, root authorship and confinement are all still recorded, so an unratified close is *visible* rather than *prevented*" — but on a bare `close` no proposal is read, and this PR's own test asserts `expect(store.closed[0].receipt.attestationFacts?.proposal).toBeUndefined()`. closing.md in the same diff concedes the opposite of the claim: "closing around it produces a receipt that records no refusal ever happened", so the after-the-fact audit offered as the reason the gate can be dropped for *every* deployment does not exist for the case it is offered against.

- **[important]** `PR description:1` — **Description advertises veto unconditionally after docs scoped it**
  The "What survives" table row reads "**A 👎 from the root author still refuses the close**" with consumer "an explicit human *no*", and the body says "the veto is now enforced explicitly" — no mention that it fires only when `--proposal-comment` is supplied. closing.md and work-graph.md were both amended to state that limit ("Scope, stated precisely because the earlier draft overstated it"); the PR body, which becomes the merge commit, was not.

- **[suggestion]** `src/cli/graph.ts:799` — **Refusal check fails open when root author is unreadable**
  `reaction.content === "-1" && root?.author !== undefined && reaction.author === root.author` silently skips the refusal whenever the root's author is absent, and the close proceeds — where the old rule would have refused for want of ratification. The PR names no such case; the error string's own `root?.author ?? "the graph root's author"` fallback shows the author is expected to be optional.

- **[suggestion]** `test/cli-graph.test.ts:381` — **#499 cited as two incompatible shapes**
  The bare-close test comments "the exact shape #499 was stuck in" (no proposal at all), while the description's verification section says "the #499 shape — proposal posted, no reactions at all" and the second test repeats "No reactions at all — the #499 case". A reader cannot tell which scenario the worked example actually was, and only one of the two tests can be reproducing it.


---

jcfischer (commented):

11 finding(s): 3 important, 6 suggestion, 2 nit.


### CodeQuality

- **[suggestion]** `test/work-graph.test.ts:208` — **Ratified half of the HITL test can no longer fail**
  With `if (node.autonomy !== "auto") return;` in `assertClosable`, every receipt passes for an `approve` node, so the retained `// A ratification is still admissible when one exists` block asserts nothing the preceding `not.toThrow()` did not already assert.
  Fix: Drop the second half, or move it to a test that inspects how the ratified receipt is recorded (attestation/facts) rather than whether it closes.

- **[suggestion]** `test/cli-graph.test.ts:475` — **Veto's author scoping has no negative test**
  Tests cover a root-author 👎 refusing the close, but nothing exercises `reaction.author === root.author` being false; dropping that clause — making any stranger's 👎 block the close on a public tracker — would keep CI green.
  Fix: Add a case: a 👎 from a non-root author on the proposal still closes the node.

### Architecture

- **[important]** `src/work-graph.ts:656` — **Refusal rule lives in CLI, not the core close gate**
  `assertClosable` now short-circuits with `if (node.autonomy !== "auto") return;` while the veto (`const vetoed = reactions.some(...)`) is implemented in `src/cli/graph.ts`'s `runClose`. Any non-CLI consumer of the typed `GraphStore` seam gets no refusal enforcement, so the rule the docs state as a runtime property is only a property of one caller.
  Fix: Move the root-author-👎 check into `assertClosable` (feed reactions through the receipt) so every close path enforces it, with the CLI reduced to message formatting.

- **[suggestion]** `docs/work-graph.md:338` — **Spec states a runtime rule the core does not own**
  The spec says "**An explicit refusal stops the close — within the proposal flow.** When `--proposal-comment` is supplied…" — a CLI-flag-scoped rule written into the primitive's spec. `docs/work-graph.md` is the contract for the primitive, not for one verb surface.
  Fix: State the refusal rule in the CLI section and note in the primitive spec that the core gate enforces evidence for `auto` only.

- **[suggestion]** `src/work-graph.ts:656` — **`autonomy` no longer discriminates at the close gate**
  After `if (node.autonomy !== "auto") return;`, `propose` and `approve` are indistinguishable to `assertClosable`, collapsing a three-valued class to a binary at the only place the spec said it was enforced. Re-introducing the promised opt-in gate later means re-deriving that distinction rather than reusing it.
  Fix: Keep the branch keyed on the specific classes (`if (node.autonomy === "propose" || node.autonomy === "approve") return;`) so the discriminator survives for a future opt-in.

### Maintainability

- **[suggestion]** `test/cli-graph.test.ts:480` — **HITL store setup copy-pasted across tests**
  The three-line fixture `.seed("495", { node: autoNode("495"), author: "jcfischer" }).seed("530", { node: { id: "530", title: "hitl", autonomy: "approve", checkpointId: "cp-530" }, parent: "495" })` appears verbatim in both the rewritten bare-close test and the new "closes on the session's say-so" test. The file already establishes the extraction shape with `autoNode("495")`.
  Fix: Add `hitlNode(id)` beside `autoNode`, or a `hitlStore()` fixture returning the seeded FakeStore.

- **[nit]** `docs/work-graph.md:362` — **Receipt-contents sentence stated twice in one section**
  Bullet 3 says "Proposal, ratifier, root authorship, confinement and `attestation` are all still recorded, so a reader can tell a ratified close from an unratified one"; eight lines later the multi-party paragraph repeats "proposal, ratifier, root authorship and confinement are all still recorded, so an unratified close is *visible* rather than *prevented*". Two edits needed whenever the receipt fields change.
  Fix: Keep the bullet; have the multi-party paragraph reference it instead of restating the field list.

- **[nit]** `docs/work-graph.md:352` — **Review-cycle meta-commentary baked into the canonical spec**
  "Scope, stated precisely because the earlier draft overstated it" and "two attempts to derive it failed review (#532)" describe this PR's review history, not the rule; a reader six months out has no draft to compare against.
  Fix: State the scope directly and move the drafting history to the PR/ADR.

### HonestOracle

- **[important]** `docs/work-graph.md:373` — **Multi-party mitigation names a consumer the PR says does not exist**
  The doc justifies removing the gate for every deployment because "a reviewer or the phase-2 auditor can act on it after the fact", while the PR description states the phase-2 headless tick "does not exist, and may not". The change deletes a gate for naming no consumer, then defends the deletion by naming a consumer that is not there.

- **[important]** `src/cli/graph.ts:808` — **Refusal message prescribes the bypass of the refusal**
  The veto throws with `Address it and post a new proposal rather than closing over the refusal.`, but nothing binds a new proposal to the refused one — `--propose` on an unchanged body yields a fresh comment id with no reactions, which closes cleanly and records no refusal. The PR's table sells this row as "an explicit human *no*"; the runtime enforces one round of it and documents the exit.

- **[suggestion]** `src/work-graph.ts:632` — **"Costs nothing" asserted for auto evidence, not shown**
  The comment claims evidence on `auto` nodes "costs nothing: the `probed` entry is derived from probes that already ran and passed", but nothing in the diff establishes that a checkpoint always carries probes — and the new test shows `receipt({ evidence: [] })` refuses. The kept gate's cost is asserted, not demonstrated.


---

jcfischer (commented):

12 finding(s): 1 blocker, 4 important, 6 suggestion, 1 nit; 1 lens(es) failed to run: ContextDrift.


### CodeQuality

- **[important]** `src/work-graph.ts:646` — **Comment repeats receipt claim the PR corrects in docs**
  The comment states "A multi-party deployment instead gets the receipt: proposal, ratifier, root authorship and confinement are all still recorded", but the same PR's docs explicitly retract this: "On the default path the first two are absent, and their absence — not their presence — is the audit signal." The authoritative in-code rationale now asserts something the spec calls wrong.
  Fix: Rewrite to match the docs: on an unratified close there is no proposal or ratifier; `attestationFacts.reasons` records their absence.

- **[suggestion]** `src/skills/orienteer/references/closing.md:118` — **Amendment rule loses the fresh-ratification requirement**
  The old text said "post a new comment and get a new 👍"; the new text is "If the resolution changes materially, post a new comment." With the proposal flow still live, a reader following this re-posts and closes without a new reaction — the rule's point.
  Fix: Restore "…post a new comment and obtain a fresh 👍 on it."

- **[suggestion]** `test/cli-graph.test.ts:504` — **Re-propose test does not assert which comment the close recorded**
  `expect(store.closed).toHaveLength(1);` is the only post-close assertion, so a close that recorded the refused `c1` instead of the fresh `c2` still passes the test that exists to pin proposal-to-close binding.
  Fix: Also assert `store.closed[0].receipt.attestationFacts?.proposal?.commentId === "c2"`.

### Architecture

- **[important]** `src/work-graph.ts:656` — **Close rule split across core seam and CLI consumer**
  `if (node.autonomy !== "auto") return;` empties the core gate for HITL nodes while the replacement refusal lives in `src/cli/graph.ts` (`const vetoed = reactions.some(...)`), so the work-graph primitive no longer owns any close-time rule for `propose`/`approve`. CONTEXT.md's `work graph` entry says the graph is "[e]xposed as `soma graph` verbs over a typed `GraphStore` seam" and "[t]he spec obligates the runtime to enforce the autonomy class at close time" — enforcement now exists only above that seam.
  Fix: Either accept the refusal check as CLI-only and amend the CONTEXT.md work-graph entry to stop claiming runtime autonomy enforcement, or pass the read reactions into `assertClosable` so the seam keeps the rule.

- **[suggestion]** `docs/work-graph.md:338` — **Spec states a CLI-only limit as a primitive property**
  The spec now documents behaviour it does not own: "it lives in the CLI, not in `assertClosable`, so a non-CLI consumer of the seam gets none of it". `docs/work-graph.md` is the primitive's spec; recording a consumer-specific limit there blurs which layer the contract binds.
  Fix: Move the CLI-scoped refusal description to the `soma graph` verb docs and leave the spec stating only what `assertClosable` guarantees.

- **[suggestion]** `src/cli/graph.ts:787` — **Refusal semantics keyed on optional flag, not node state**
  `if (state.node.autonomy !== "auto" && commentId !== undefined)` makes the entire refusal path conditional on a caller-supplied flag, so the CLI cannot later enforce a refusal recorded on a node without a source-of-truth change. Adding real refusal binding ("threading reactions through the receipt") would require reworking this branch rather than extending it.
  Fix: Read the node's latest proposal comment when `--proposal-comment` is absent, so the refusal path is keyed on node state and can be strengthened without restructuring.

### ContextDrift — DID NOT RUN

- **[important]** `(lens output):0` — **ContextDrift: model deviated from JSON contract**
  claude extraction failed (exit-nonzero) exit=1
--- output ---
{"is_error":true,"duration_api_ms":0,"num_turns":1,"stop_reason":"stop_sequence","session_id":"3384d2e5-5e6d-4c6c-bc6a-11c3b459da11","total_cost_usd":0,"usage":{"input_tokens":0,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":0,"server_tool_use":{"web_search_requests":0,"web_fetch_requests":0},"service_tier":"standard","cache_creation":{"ephemeral_1h_input_tokens":0,"ephemeral_5m_input_tokens":0},"inference_geo":"","iterations":[],"speed":"standard"},"modelUsage":{},"permission_denials":[],"terminal_reason":"api_error","fast_mode_state":"off","fast_mode_disabled_reason":"sdk_opt_in_required","subtype":"success","api_error_status":529,"result":"API Error: 529 Overloaded. This is a server-side issue, usually temporary — try again in a moment. If it persists, check https://status.claude.com.","type":"result","duration_ms":242782,"uuid":"36734c17-e30b-4e1a-866c-046b9c6598ac"}


### Maintainability

- **[suggestion]** `src/work-graph.ts:630` — **24-line rationale comment guards a one-line return**
  `if (node.autonomy !== "auto") return;` is preceded by five paragraphs restating the same argument three times — "it did not verify the close, it only prevented it", then "an unratified close is visible rather than prevented", then "an unratified close is now *possible*, not *approved*". Comment blocks that argue rather than explain rot away from the code they sit above.
  Fix: Keep two lines (what the rule is, plus the #499 pointer) and move the argument to docs/work-graph.md, which already carries it.

- **[suggestion]** `src/cli/graph.ts:786` — **Veto rationale duplicated between CLI comment and docs**
  The 8-line comment above `const vetoed = ...` ("Deliberately a speed bump, not a control ... machinery this primitive does not want") is a near-verbatim restatement of the docs/work-graph.md paragraph and the closing.md paragraph added in the same diff. Three copies of one argument means three places to update when the veto changes.
  Fix: Reduce the CLI comment to "speed bump, not a gate — see docs/work-graph.md §3.2" and let the spec hold the reasoning.

- **[nit]** `test/cli-graph.test.ts:490` — **Three-line HITL store fixture copy-pasted across new tests**
  `new FakeStore().seed("495", ...).seed("530", { node: { id: "530", title: "hitl", autonomy: "approve", checkpointId: "cp-530" }, parent: "495" })` appears identically in three of the tests touched or added here.
  Fix: Extract `function hitlStore() { ... }` next to the existing `autoNode` helper.

### HonestOracle

- **[blocker]** `src/work-graph.ts:649` — **Removal justified by receipt fields the default path never writes**
  The comment defending the unconditional removal claims "A multi-party deployment instead gets the receipt: proposal, ratifier, root authorship and confinement are all still recorded" (and repeats it at :653), yet docs/work-graph.md:388-390 in this same PR retracts exactly that sentence — "an earlier draft claimed the receipt carried 'proposal, ratifier, root authorship and confinement' ... On the default path the first two are absent." A consumer reading the seam gets the version the PR knows to be false, attached to the load-bearing argument for dropping the gate.

- **[important]** `docs/work-graph.md:367` — **Canonical spec asserts and then retracts the same receipt claim**
  Line 367 states "**The receipt is unchanged.** Proposal, ratifier, root authorship, confinement and `attestation` are all still recorded"; line 388 calls that same list an "earlier draft" error and says the first two are absent on the default path. The correction was appended rather than applied, so the spec now says both things and a reader believes whichever bullet they hit first.


---

jcfischer (commented):

7 finding(s): 2 important, 3 suggestion, 2 nit; 1 lens(es) failed to run: HonestOracle.


### CodeQuality

- **[important]** `src/cli/graph.ts:806` — **Veto throws before the dry-run branch**
  `if (vetoed) { throw new SomaCliError(...) }` runs before `if (parsed.options.dryRun === true)`, unlike the probe-refusal check just above it which is explicitly guarded by `parsed.options.dryRun !== true` ("A dry run keeps going: showing the whole receipt is the point of asking for one"). `soma graph close <id> --proposal-comment <id> --dry-run` now exits 1 and prints no receipt, even though a dry run writes nothing.
  Fix: Guard the veto the same way (`vetoed && parsed.options.dryRun !== true`) and surface the refusal in the preview verdict instead; add a test alongside "--dry-run still renders the receipt when the registry refused a probe".

- **[suggestion]** `docs/work-graph.md:353` — **"Surfaced, not enforced" contradicts the next sentence**
  The bullet reads "**An explicit refusal is surfaced, not enforced.**" and then "a 👎 ... refuses *that* close" — the CLI throws with exit code 1 and closes nothing, which is enforcement. closing.md repeats the same pairing ("A 👎 is surfaced, not enforced" / "the close refuses and names them").
  Fix: Say what it is: "An explicit refusal blocks that one close, but binds nothing after it" — the limits paragraph already carries the speed-bump framing.

### Maintainability

- **[suggestion]** `src/work-graph.ts:630` — **24-line rationale comment guards a one-line return**
  The block from `// Evidence is required of \`auto\` nodes only` through `// It is a tool, no longer a toll.` is 24 lines of prose fronting `if (node.autonomy !== "auto") return;`, and its substance ("The removal is UNCONDITIONAL... two attempts to derive it failed review") is already the canonical text in docs/work-graph.md. Two full copies of a design argument drift apart on the next edit.
  Fix: Keep 2-3 lines at the guard ("HITL closes on the session's say-so; gate named no consumer — see docs/work-graph.md §3.2 and #499") and let the doc own the argument.

- **[suggestion]** `docs/work-graph.md:366` — **Spec section states a claim, then retracts it in place**
  The section asserts "**The receipt is unchanged.** Proposal, ratifier, root authorship, confinement and `attestation` are all still recorded", then twenty lines later corrects itself: "an earlier draft claimed the receipt carried 'proposal, ratifier, root authorship and confinement'... On the default path the first two are absent". A canonical spec that contains both its wrong and right version leaves readers to work out which governs.
  Fix: Fix the first claim in place (name only what the default path writes) and delete the "Naming this precisely matters" paragraph — review history belongs in the PR, not the spec.

- **[nit]** `src/cli/graph.ts:786` — **Two adjacent comment blocks make the same point**
  The branch opens with "A proposal comment is optional now: a HITL node closes on the session's say-so (§3.2). When one IS supplied, its reactions still carry weight — a ratification as admissible evidence, and a root-author 👎 as a refusal", and six lines later the veto block restates "Ratification is no longer required; an explicit refusal is still surfaced." plus a 7-line expansion already present verbatim in docs and closing.md.
  Fix: Drop the second header line and the speed-bump paragraph; one pointer to docs/work-graph.md §3.2 covers it.

- **[nit]** `test/cli-graph.test.ts:490` — **HITL store fixture copy-pasted across three new tests**
  The identical three-line `new FakeStore().seed("495", ...).seed("530", { node: { id: "530", title: "hitl", autonomy: "approve", checkpointId: "cp-530" }, parent: "495" })` appears in the bare-close test, the fresh-proposal test, and the say-so test. Changing the node shape means editing three literals.
  Fix: Extract `function hitlStore() { ... }` alongside the existing `autoNode` helper.

  claude extraction recovered JSON but it is not lens-shaped (no `summary` or `findings` field) — extractor=claude-envelope


---

jcfischer (commented):

8 finding(s): 2 important, 5 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/skills/orienteer/references/closing.md:96` — **"Surfaced, not enforced" contradicted one sentence later**
  The heading says "**A 👎 is surfaced, not enforced.**" and the next sentence says "the close refuses and names them" — the CLI throws and exits 1, which is enforcement. An operator reading this doc expects a warning and gets a hard failure.
  Fix: Reword to what the code does, e.g. "**A 👎 refuses that close.** It is a speed bump, not a control — a fresh proposal closes cleanly."

### Maintainability

- **[suggestion]** `src/work-graph.ts:630` — **33-line comment guards a one-line return**
  The diff adds 33 lines of rationale (`// The removal is UNCONDITIONAL...`, `// The receipt still records everything it recorded before...`) in front of `if (node.autonomy !== "auto") return;`. A reader scrolls a screen of argument to reach one branch, and the argument will rot independently of the code.
  Fix: Keep two lines here ("HITL closes on the session's say-so — see docs/work-graph.md §3.2") and let the doc hold the argument.

- **[suggestion]** `docs/work-graph.md:362` — **Spec duplicates its own paragraph and annotates the duplication**
  The `attestationFacts.reasons` / "no proposal comment recorded" / "no ratification found" list appears twice in the same section, and the diff then adds `Stated twice in this section on purpose: an earlier revision asserted the ratified-close field list as if it applied to every close`. Meta-commentary about a prior revision is review residue, not spec.
  Fix: Apply the correction in place — one statement of the reasons list — and drop the paragraph explaining why it was stated twice.

- **[suggestion]** `src/cli/graph.ts:789` — **Two adjacent comment blocks state the same rule**
  The pre-branch block says `a ratification as admissible evidence, and a root-author 👎 as a refusal`; the block eight lines later re-states `Ratification is no longer required; an explicit refusal is still surfaced` plus seven more lines of the same argument already in work-graph.ts and closing.md.
  Fix: Delete the second block's prose; keep one sentence naming why the check is CLI-side.

- **[nit]** `test/cli-graph.test.ts:490` — **Three-line HITL store fixture copy-pasted across new tests**
  `new FakeStore().seed("495", ...).seed("530", { node: { id: "530", title: "hitl", autonomy: "approve", checkpointId: "cp-530" }, parent: "495" })` appears verbatim in three tests added by this diff.
  Fix: Extract `function hitlStore() { ... }` next to `autoNode`.

### HonestOracle

- **[important]** `test/cli-graph.test.ts:505` — **Re-propose test asserts a weaker fact than the PR claims it pins**
  The PR states the re-propose limit is "**pinned by a test**, so the limit cannot quietly change in either direction", but the test only asserts `expect(store.closed).toHaveLength(1)` after two closes — it never asserts the close recorded `c2` rather than `c1`, so a change that binds the new proposal to the refused one could still leave the count at 1. The claimed bidirectional pin is a surrogate for what is measured.

- **[important]** `src/cli/graph.ts:803` — **Refusal fails open when the root author is unresolved; PR does not say so**
  The predicate `root?.author !== undefined && reaction.author === root.author` yields `vetoed === false` whenever `root?.author` is unreadable, so a 👎 is silently ignored and the close proceeds; the PR's list of "three limits, all deliberate" does not name this fourth one, and the fallback string `root?.author ?? "the graph root's author"` in the message shows the author is known to be possibly-absent. A reader of the docs believes an explicit refusal is surfaced when it may not be.

- **[suggestion]** `docs/work-graph.md:391` — **Failed derivation attempts cite one issue for two attempts**
  The spec asserts "two attempts to derive it failed review (#532)" and the code comment repeats "two attempts to derive it failed review" with no citation at all; a single issue number is offered as the record for two distinct attempts, and the diff gives a reader no way to locate the second.


---

jcfischer (commented):

13 finding(s): 3 important, 7 suggestion, 3 nit.


### CodeQuality

- **[suggestion]** `src/cli/graph.ts:802` — **Veto counts the proposer's own 👎, unlike ratification**
  `reaction.content === "-1" && root?.author !== undefined && reaction.author === root.author` applies no proposer exclusion, while the line below passes `proposal.author` to `selectRatification` precisely so a self-👍 is ignored. When the operator posts the proposal under their own credential (what `close --propose` does), their own 👎 permanently refuses that comment id while their own 👍 counts for nothing.
  Fix: Either exclude `proposal.author` from the veto for symmetry, or note in the comment that self-refusal is intentionally honoured where self-ratification is not.

- **[nit]** `test/cli-graph.test.ts:495` — **Veto test hard-codes FakeStore's generated comment ids**
  `store.reactions.set("c1", ...)` and the later `--proposal-comment", "c2"` depend on FakeStore numbering comments c1, c2 in order; the reaction is injected by reaching into the fixture rather than through the CLI.
  Fix: Capture the comment id from the `--propose` output instead of assuming `c1`/`c2`.

### Architecture

- **[important]** `src/cli/graph.ts:788` — **CLI now owns the only remaining autonomy-class behaviour**
  After `if (node.autonomy !== "auto") return;` in `assertClosable`, the core seam treats every HITL node identically, while `src/cli/graph.ts` still branches on `state.node.autonomy !== "auto" && commentId !== undefined`. The typed `GraphStore` seam CONTEXT.md describes as the primitive's surface no longer distinguishes `propose` from `approve` at all; the distinction survives only in one CLI consumer.
  Fix: Either drop the `autonomy !== "auto"` test in the CLI (it now discriminates nothing the core recognizes) or keep the refusal predicate behind a core-exported helper so non-CLI consumers can reuse it.

- **[suggestion]** `docs/work-graph.md:341` — **Spec states a runtime obligation with no owner**
  The section says "A HITL node **closes on the closing session's say-so**" but the enforcement it describes next — "When `--proposal-comment` is supplied, a 👎 from the graph root's author on that comment refuses *that* close" — is a property of the CLI, not of the work-graph primitive the doc specifies. CONTEXT.md's work-graph entry frames `soma graph` verbs as one consumer over a typed `GraphStore` seam.
  Fix: Move the veto paragraph under a clearly-labelled "CLI behaviour" subsection so the primitive spec and the consumer behaviour are not read as one contract.

- **[suggestion]** `src/work-graph.ts:663` — **`agentExternalEvidenceKinds(node.autonomy)` now only ever sees `auto`**
  The new `if (node.autonomy !== "auto") return;` guard makes the following `const admissible = agentExternalEvidenceKinds(node.autonomy);` a call with a statically known argument. The helper's per-class branching is now dead surface that reads as still-live policy.
  Fix: Inline the `auto` case or add a comment at the helper marking the non-`auto` branches unreachable from `assertClosable`.

### Maintainability

- **[suggestion]** `src/work-graph.ts:630` — **27-line rationale comment guards one-line return**
  The block from `// Evidence is required of \`auto\` nodes only` through `// It is a tool, no longer a toll.` precedes a single statement, `if (node.autonomy !== "auto") return;`. A reader scrolls a screen of PR-description prose to reach one line of logic, and the prose will drift from the code on the next change.
  Fix: Keep two lines here (`HITL closes on the session's say-so; removal is unconditional — see docs/work-graph.md §3.2`) and let the spec hold the argument.

- **[suggestion]** `src/work-graph.ts:632` — **Same rationale restated in four files**
  "it did not verify the close, it only prevented it" appears in src/work-graph.ts, docs/work-graph.md ("the rule did not verify those closes — it prevented them"), closing.md ("the rule did not verify closes, it prevented them"), and test/work-graph.test.ts ("it did not verify the close, it only prevented it"). Four copies of one argument means four places to update when the rule changes.
  Fix: Make docs/work-graph.md §3.2 canonical; the other three cite it.

- **[suggestion]** `docs/work-graph.md:362` — **Spec repeats a paragraph and annotates the repetition**
  The receipt-records-the-absence point is made once under "The receipt still distinguishes the two cases" and again under "On multi-party deployments", then the text says "Stated twice in this section on purpose" and explains that an earlier revision was corrected by appending. A canonical spec carrying its own revision history as body text is harder to read and to keep true.
  Fix: Apply the correction: delete the duplicate paragraph and the meta-note explaining it.

- **[nit]** `src/cli/graph.ts:786` — **Two adjacent comments state the same rule**
  The block comment above the branch already says "A proposal comment is optional now… When one IS supplied, its reactions still carry weight — a ratification as admissible evidence, and a root-author 👎 as a refusal", and three lines later a second comment repeats "Ratification is no longer required; an explicit refusal is still surfaced."
  Fix: Drop the second line; keep the longer block.

- **[nit]** `test/cli-graph.test.ts:490` — **HITL store fixture copy-pasted across three new tests**
  The identical three-line `new FakeStore().seed("495", …).seed("530", { node: { id: "530", title: "hitl", autonomy: "approve", checkpointId: "cp-530" }, parent: "495" })` appears in the bare-close test, the re-propose test, and the say-so test.
  Fix: Extract `const hitlStore = () => new FakeStore().seed(...).seed(...)` near `autoNode`.

### HonestOracle

- **[important]** `src/skills/orienteer/references/closing.md:96` — **Docs promise a named refuser the code cannot always name**
  closing.md states "the close refuses and names them", but the CLI predicate is `root?.author !== undefined && reaction.author === root.author` — when the root author is unresolved no reaction ever matches, so nothing refuses and nobody is named. The doc describes a behaviour the reader cannot rely on and gives no hint of the precondition.

- **[important]** `docs/work-graph.md:380` — **"Two attempts to derive it failed review (#532)" cites one issue for two events**
  The canonical spec attributes both derivation failures to a single reference — "two attempts to derive it failed review (#532)" — with no second citation, so a reader cannot check either attempt. The load-bearing justification for making the removal unconditional rests on evidence that is not locatable from the citation given.

- **[suggestion]** `docs/work-graph.md:355` — **"Available, not required" overstated: bare close silently skips refusal**
  The spec frames the proposal flow as optional convenience — "The proposal flow is available, not required" — while the CLI branch `state.node.autonomy !== "auto" && commentId !== undefined` means the documented default path (`soma graph close <id>`) is precisely the one where a recorded 👎 has zero effect. The framing sells opting out as neutral when it is the path that disables the only surviving check.


---

jcfischer (commented):

12 finding(s): 3 important, 6 suggestion, 3 nit.


### CodeQuality

- **[important]** `src/cli/graph.ts:788` — **Non-auto guard now silently drops proposal comment on auto nodes**
  The condition changed from `if (state.node.autonomy !== "auto")` gating a required id to `if (state.node.autonomy !== "auto" && commentId !== undefined)`, so on an `auto` node an explicitly supplied `--proposal-comment` is never read: no `proposal` is set, no reactions fetched, and the receipt records "no proposal comment recorded" despite the operator naming one. Previously the flag was equally unused there, but the flag's meaning has now shifted from "required on HITL" to "optional evidence anywhere", making the silent discard a user-visible surprise.
  Fix: Either error on `--proposal-comment` with an `auto` node, or read the comment regardless of autonomy and let `selectRatification` decide.

- **[suggestion]** `src/cli/graph.ts:802` — **Veto predicate is satisfiable by a proposer self-👎**
  `reaction.content === "-1" && root?.author !== undefined && reaction.author === root.author` has no proposer-exclusion, unlike `selectRatification(reactions, proposal.author, root?.author)` which ignores the proposer's own reaction. When the root author is also the proposer, their own 👎 on their own proposal refuses the close.
  Fix: Exclude `proposal.author` from the veto scan, mirroring `selectRatification`'s suppression rule.

- **[suggestion]** `test/cli-graph.test.ts:496` — **Re-propose test asserts only the total close count**
  `expect(store.closed).toHaveLength(1)` passes whether the second close recorded `c2`, `c1`, or no proposal at all — so the pinned "limit" is weaker than the comment claims ("a new comment id, no reactions on it, and the close proceeds").
  Fix: Assert `store.closed[0].receipt.attestationFacts?.proposal?.commentId === "c2"`.

- **[nit]** `src/cli/graph.ts:806` — **Refusal message fallback string is unreachable**
  `vetoed` can only be true when `root?.author !== undefined`, so `${root?.author ?? "the graph root's author"}` never renders the fallback.
  Fix: Bind the author to a local before the check and interpolate it directly.

### Architecture

- **[suggestion]** `src/work-graph.ts:663` — **Autonomy class now only meaningful to the CLI consumer**
  With `if (node.autonomy !== "auto") return;` placed before `agentExternalEvidenceKinds(node.autonomy)`, that helper can only ever receive `auto`, so the `propose`/`approve` branches of the evidence-kind table are dead at this call site. The typed autonomy class remains in the core node shape while every behavioural difference between its values now lives in `src/cli/graph.ts`, inverting the documented direction where the seam owns the contract and the CLI is a consumer.
  Fix: Either narrow `agentExternalEvidenceKinds` to the `auto` case at this call site, or keep a per-class evidence policy in the core so `autonomy` still discriminates where it is declared.

### Maintainability

- **[suggestion]** `src/work-graph.ts:630` — **33-line rationale comment guards a one-line return**
  The block from "Evidence is required of `auto` nodes only" through "It is a tool, no longer a toll." precedes a single statement, `if (node.autonomy !== "auto") return;`, and restates docs/work-graph.md's §3.2 text nearly verbatim ("How many people are watching" is not derivable — two attempts to derive it failed review"). A reader of `assertClosable` must scroll past a full screen of prose to reach the next executable line, and the argument now has to be kept in sync in two places.
  Fix: Keep two lines at the guard ("HITL closes on the session's say-so; see docs/work-graph.md §3.2") and let the doc own the rationale.

- **[suggestion]** `src/cli/graph.ts:792` — **Veto predicate reimplements selectRatification's author check**
  `reactions.some((reaction) => reaction.content === "-1" && root?.author !== undefined && reaction.author === root.author)` is an inline re-derivation of the root-author-suppression logic `selectRatification` already owns — and it runs three lines above the `selectRatification(reactions, proposal.author, root?.author)` call that consumes the same inputs. Two independent copies of "is this the root author's 👎" will drift.
  Fix: Export a `rootAuthorRefusal(reactions, rootAuthor)` helper from the ratification module and call it here.

- **[nit]** `src/cli/graph.ts:786` — **Two adjacent comment blocks state the same rule**
  The block above the branch ("A proposal comment is optional now… When one IS supplied, its reactions still carry weight — a ratification as admissible evidence, and a root-author 👎 as a refusal") and the block inside it ("Ratification is no longer required; an explicit refusal is still surfaced") make the same point eight lines apart.
  Fix: Drop the second one-liner; the eight-line block below it already carries the detail.

- **[nit]** `test/cli-graph.test.ts:490` — **HITL store fixture copy-pasted across three new tests**
  `new FakeStore().seed("495", …).seed("530", { node: { id: "530", title: "hitl", autonomy: "approve", checkpointId: "cp-530" }, parent: "495" })` appears verbatim in the bare-close test, the veto-depth test, and the say-so test.
  Fix: Extract `function hitlStore() { … }` at the top of the file.

### HonestOracle

- **[important]** `src/cli/graph.ts:788` — **Veto's advertised consumer cannot reach the veto path**
  The PR names the consumer as "someone about to close by reflex after being told no", but `if (state.node.autonomy !== "auto" && commentId !== undefined)` means a reflex `soma graph close 530` — the exact command `closing.md` now promotes as the default — never reads reactions. The mechanism is gated on the user voluntarily supplying the id of the comment they were refused on, i.e. on deliberateness, which is the opposite of reflex.

- **[important]** `docs/work-graph.md:379` — **"Not derivable" asserted from two failed attempts**
  "'how many people are watching' is not derivable — two attempts to derive it failed review (#532)" converts two rejected implementations into an impossibility claim, and #532 is a review outcome, not a proof of non-derivability. This load-bearing premise justifies the unconditional removal for multi-party deployments.

- **[suggestion]** `test/cli-graph.test.ts:490` — **Re-propose test asserts a count, not the documented limit**
  The comment claims the docs' claim "is executable here so it cannot quietly become false in either direction", but the only post-condition is `expect(store.closed).toHaveLength(1)` — it never asserts the close recorded `c2` rather than the refused `c1`, so the binding property the prose describes is unpinned.


---

jcfischer (commented):

11 finding(s): 1 blocker, 2 important, 6 suggestion, 2 nit.


### CodeQuality

- **[important]** `src/cli/graph.ts:803` — **Refusal check fails open when root author is unresolved**
  `reaction.content === "-1" && root?.author !== undefined && reaction.author === root.author` evaluates to `false` for every reaction when `root?.author` is undefined, so a node whose root author cannot be read accepts any close regardless of a 👎. The error message's `root?.author ?? "the graph root's author"` fallback admits the author can be absent, yet the predicate treats absence as consent.
  Fix: Either warn on stderr when `root?.author === undefined` and a `-1` reaction exists, or treat an unresolvable root author as a refusal-eligible state rather than silently skipping the check.

- **[suggestion]** `src/cli/graph.ts:802` — **Veto counts the proposer's own 👎, unlike ratification**
  `selectRatification(reactions, proposal.author, root?.author)` excludes the proposer's own reaction, but the veto predicate has no proposer exclusion — when the root author is also the proposer, their own 👎 on their own proposal refuses the close. The two reaction readers apply inconsistent author rules to the same comment.
  Fix: Exclude `proposal.author` from the veto predicate, matching `selectRatification`'s suppression rule.

- **[suggestion]** `src/cli/graph.ts:787` — **`--proposal-comment` silently ignored on auto nodes**
  The guard `state.node.autonomy !== "auto" && commentId !== undefined` means an `auto` node passed `--proposal-comment <id>` never reads the comment, never records a proposal, and never reports that the flag was dropped. Previously the flag was only reachable on non-auto nodes, so the silent-drop path is new.
  Fix: Error on `--proposal-comment` supplied for an `auto` node rather than ignoring it.

- **[nit]** `src/cli/graph.ts:807` — **Unreachable fallback in the refusal message**
  `${root?.author ?? "the graph root's author"}` can only render the fallback when `root?.author` is undefined — but the `vetoed` predicate requires `root?.author !== undefined` to be true, so the throw is unreachable in that state.
  Fix: Drop the `??` fallback, or narrow `root.author` to a local const before the predicate so the type reflects the invariant.

### Architecture

- **[suggestion]** `src/work-graph.ts:663` — **Autonomy class collapses to a binary at the core seam**
  `if (node.autonomy !== "auto") return;` sits above `agentExternalEvidenceKinds(node.autonomy)`, so that helper now only ever receives `auto` and the three-valued autonomy class no longer discriminates anything in the core. Forward-compat: reintroducing a `propose`/`approve` distinction means re-deriving it, since the seam no longer carries it.
  Fix: Narrow the early return to the evidence check only, or note in the type/JSDoc that `agentExternalEvidenceKinds` is now `auto`-only.

- **[suggestion]** `docs/work-graph.md:352` — **Canonical spec carries review-process meta-commentary**
  "Stated twice in this section on purpose: an earlier revision asserted the ratified-close field list as if it applied to every close, and a correction that is appended rather than *applied* leaves a document saying both things" documents the doc's own edit history inside the normative spec. `docs/work-graph.md` is the cited spec for the `work graph` glossary entry; readers of a contract should not have to parse which sentence supersedes which.
  Fix: Apply the correction in place and delete the meta-paragraph; the revision history belongs in the PR, not the spec.

### Maintainability

- **[suggestion]** `src/cli/graph.ts:795` — **Proposal branch in runClose does three jobs inline**
  Inside one `if (state.node.autonomy !== "auto" && commentId !== undefined)` branch the diff reads the comment (`const proposalRef: CommentRef = await graph.readComment`), computes refusal (`const vetoed = reactions.some(`), throws, and then selects ratification (`ratification = selectRatification(reactions, proposal.author, root?.author)`) — all in an already-long close handler.
  Fix: Extract `resolveProposal(graph, ref, commentId, root)` returning `{proposal, ratification, refusedBy}`; leave `runClose` with the throw.

- **[nit]** `test/cli-graph.test.ts:508` — **New test re-walks path an adjacent test already covers**
  "without a 👎, a HITL node closes on the session's say-so" runs the same `--propose` then `--proposal-comment c1` sequence whose clean-close half is already asserted in "the veto is one close deep"; it differs only by `expect(store.closed[0].receipt.attestation).toBe("unverified")`.
  Fix: Fold the attestation assertion into the existing test, or drop the redundant propose step.

### HonestOracle

- **[blocker]** `docs/work-graph.md:339` — **"The human in the loop is the person running the session" contradicted by deleted text**
  The added line "A HITL node **closes on the closing session's say-so**. The human in the loop is the person running the session" is asserted in the same hunk that deletes "the executing bot posts the resolution as a **proposal comment**; the receipt is the principal's 👍" — which stated that the closing session is a bot and the human is the reactor. Nothing in the diff establishes that a `close` invocation comes from a human, so an agent session can now close an `approve`-class node and the spec tells readers a human did.

- **[important]** `PR description:1` — **"One change" claim refuted by the diff's own test comment**
  The description opens "One change: `assertClosable` no longer requires ratified evidence" and offers #499 as the justification, but test/cli-graph.test.ts states "Relaxing `assertClosable` alone did not fix it: this CLI branch still refused, which is the layer #499 actually hit." The named change does not unblock the cited example; a second, unnamed CLI change does.

- **[suggestion]** `PR description:1` — **"Exactly as before" claims parity with a path that never executed**
  "the #499 shape — proposal posted, no reactions at all — now closes, recording `attestation: unverified` exactly as before" asserts continuity with a prior behaviour that did not exist: before this diff that shape was refused, so no receipt was written to compare against.


---

jcfischer (commented):

12 finding(s): 3 important, 7 suggestion, 2 nit.


### CodeQuality

- **[important]** `src/cli/graph.ts:795` — **Refusal path newly reaches auto nodes, untested**
  The block moved out from under `if (state.node.autonomy !== "auto") {` and is now keyed on `if (commentId !== undefined)`, so an `auto` node passed `--proposal-comment` reads reactions and can hard-fail on a root-author 👎 — previously the flag was ignored there. No test in the diff passes `--proposal-comment` to an auto node.
  Fix: Add a test closing an `auto` node with `--proposal-comment` carrying a root-author 👎, pinning whichever outcome is intended.

- **[suggestion]** `src/cli/graph.ts:807` — **Refusal message contradicts the exit it accompanies**
  `Nothing here can stop you closing anyway — this is a reminder, not a gate.` is printed by the one code path that did stop the close (SomaCliError, exit 1), and unlike the message it replaces it names no next step.
  Fix: Say the close was refused and how to proceed, e.g. "re-propose with --propose and close on the new comment".

- **[nit]** `test/cli-graph.test.ts:493` — **Test writes FakeStore internals instead of using its API**
  `store.reactions.set("c1", [{ id: "r1", content: "-1", author: "jcfischer" }]);` reaches into the fake's field directly, while every other setup in the file goes through `.seed(...)`.
  Fix: Add a `reactOn(commentId, reaction)` helper on FakeStore and use it here.

### Architecture

- **[suggestion]** `src/work-graph.ts:662` — **Autonomy class collapses to auto/non-auto at the seam**
  `if (node.autonomy !== "auto") return;` makes `propose` and `approve` indistinguishable to every core consumer, so `agentExternalEvidenceKinds(node.autonomy)` is now only ever reached with `auto`. The three-valued type remains in the public shape but only two values carry behaviour, which invites future consumers to key on a distinction the core no longer honours.
  Fix: Either narrow the parameter to the values that still discriminate, or add a comment at `agentExternalEvidenceKinds` recording that its non-`auto` branches are now unreachable from `assertClosable`.

### Maintainability

- **[suggestion]** `src/work-graph.ts:630` — **33-line rationale comment guards a one-line return**
  The block from "Evidence is required of `auto` nodes only" through "It is a tool, no longer a toll." precedes `if (node.autonomy !== "auto") return;` — six paragraphs of history, PR argument, and consumer analysis for one guard. A reader scrolls past a page of prose to reach two lines of logic.
  Fix: Keep 2-3 lines at the guard ("HITL closes on the session's say-so; removal is unconditional — see §3.2") and move the history/consumer argument to docs/work-graph.md, which already carries it.

- **[suggestion]** `src/cli/graph.ts:783` — **Same rationale restated across five files**
  "it lives here in the CLI rather than in `assertClosable` because it is not a contract rule... a fresh `--propose` mints a new comment with no reactions" duplicates docs/work-graph.md's "it lives in the CLI, not in `assertClosable`... `--propose` on an unchanged body mints a fresh comment that closes cleanly" and closing.md's "re-proposing produces a fresh comment that closes cleanly". Five copies of one argument drift apart on the first amendment.
  Fix: Pick docs/work-graph.md §3.2 as the single home; replace the CLI and test copies with one-line pointers to it.

- **[suggestion]** `src/cli/graph.ts:783` — **Two adjacent comment blocks state the same rule**
  The first block says "When one IS supplied, its reactions still carry weight — a ratification as admissible evidence, and a root-author 👎 as a refusal"; ten lines later the second says "Ratification is no longer required; an explicit refusal is still surfaced." One branch, one rule, two headers.
  Fix: Merge into a single comment above the `if (commentId !== undefined)` branch.

- **[suggestion]** `src/cli/graph.ts:795` — **runClose proposal branch does read, refuse, and ratify inline**
  The `if (commentId !== undefined)` branch now reads the comment, builds `proposal`, reads reactions, evaluates the veto predicate, throws, then calls `selectRatification` and pushes evidence — four concerns in one inline block, mostly comment.
  Fix: Extract `resolveProposal(graph, ref, commentId, root)` returning `{proposal, ratification}` and throwing on refusal; `runClose` keeps one call.

- **[nit]** `test/cli-graph.test.ts:480` — **HITL store fixture copy-pasted across three new tests**
  `new FakeStore().seed("495", ...).seed("530", { node: { id: "530", title: "hitl", autonomy: "approve", checkpointId: "cp-530" }, parent: "495" })` appears verbatim in three of the new tests.
  Fix: Add a `hitlStore()` helper next to `autoNode` and call it from each test.

### HonestOracle

- **[important]** `src/cli/graph.ts:792` — **Refusal and ratification silently extended to auto nodes**
  Hoisting the block out of `if (state.node.autonomy !== "auto")` makes `--proposal-comment` take effect on `auto` nodes, justified only by a comment asserting "Supplying a proposal on an `auto` node is unusual rather than wrong: its reactions are read on the same terms" — no test, and the PR calls itself "One change: `assertClosable` no longer requires ratified evidence on `approve`/`propose` nodes". A root-author 👎 now refuses an `auto` close, and a ratification pushes `approved` evidence onto an `auto` receipt, for a node class the PR says keeps its machine-checkable gate.

- **[important]** `docs/work-graph.md:355` — **"Three limits, all deliberate" is not exhaustive**
  The canonical spec enumerates the refusal's weaknesses as "Three limits, all deliberate", omitting the silent no-op that closing.md in this same PR does disclose: "Where the root walk fails, no reaction matches and nothing refuses, silently." A reader who trusts the closed list in the spec gets a false picture of when the 👎 actually fires.

- **[suggestion]** `docs/work-graph.md:379` — **#548 cited for reasoning about a different branch's work**
  The spec sends readers to "branch `feat/self-ratification` at `b8acfe4`, with the reasoning in the closing comment of [#548]", while the PR description identifies #548 as "(operator posture) — closed unmerged; `feat/graph-posture` kept for reference only". The pointer for why the two derivation attempts failed lands on a PR about a different change.


---

jcfischer (commented):

15 finding(s): 1 blocker, 3 important, 7 suggestion, 4 nit.


### CodeQuality

- **[important]** `src/cli/graph.ts:806` — **Refusal check fails open when root author is unresolved**
  `reaction.author === root.author` is guarded by `root?.author !== undefined`, so when the root walk yields no author the predicate is uniformly false and every 👎 is silently ignored — the close proceeds with no signal that the refusal check did not run. `closing.md` documents this ("Where the root walk fails, no reaction matches and nothing refuses, silently") but the code emits nothing.
  Fix: When `root?.author === undefined` and reactions contain a `-1`, warn on stderr that the refusal check could not be evaluated.

- **[nit]** `src/cli/graph.ts:810` — **Unreachable fallback in refusal message**
  `${root?.author ?? "the graph root's author"}` can only be reached when `vetoed` is true, which requires `root?.author !== undefined`. The fallback branch is dead.
  Fix: Use `root.author` directly (narrow before the throw) and drop the `??` fallback.

- **[suggestion]** `src/cli/graph.ts:803` — **Refusal counts the proposer's own 👎, unlike ratification**
  `selectRatification(reactions, proposal.author, root?.author)` excludes the proposer's own reaction, but the veto predicate only checks `reaction.author === root.author` — so when the bot posting the proposal *is* the root author, its own 👎 refuses the close. The two paths read the same reaction list under different authorship rules.
  Fix: Exclude `proposal.author` from the veto predicate, matching `selectRatification`.

### Architecture

- **[important]** `src/cli/graph.ts:788` — **Autonomy class drops out of the core seam into the CLI**
  The diff removes the `state.node.autonomy !== "auto"` branch from `runClose` and keys entirely on `commentId !== undefined`, with the comment conceding "`assertClosable` no longer distinguishes `propose` from `approve` from anything else non-`auto`". CONTEXT.md's work-graph entry states the node's autonomy class is a property "the spec obligates the runtime to enforce … at close time"; after this change no layer enforces a `propose`/`approve` distinction, so the glossary's three-class model has no runtime referent.
  Fix: Either collapse `propose`/`approve` into one class in the type and spec, or state in CONTEXT.md's work-graph entry that autonomy is now advisory metadata at close time.

- **[suggestion]** `src/work-graph.ts:663` — **`agentExternalEvidenceKinds(node.autonomy)` now only ever receives `auto`**
  The new `if (node.autonomy !== "auto") return;` guard sits directly above `const admissible = agentExternalEvidenceKinds(node.autonomy)`, so the parameter is now a constant at the only surviving call path. The function's per-class branching is dead surface that will read as live policy to the next editor.
  Fix: Call `agentExternalEvidenceKinds("auto")` explicitly, or drop the parameter if no other caller remains.

- **[suggestion]** `docs/work-graph.md:355` — **Canonical spec documents a CLI-only behaviour as a primitive property**
  The spec section for the work-graph primitive now describes the refusal in terms of `--proposal-comment` and `close --propose`, while itself noting "it lives in the CLI, not in `assertClosable`, so a non-CLI consumer of the seam gets none of it". docs/work-graph.md is the primitive's spec; a consumer-specific flag behaviour documented there invites other `GraphStore` consumers to assume the seam provides it.
  Fix: Move the refusal description to the CLI/orienteer docs and leave a one-line pointer in the spec noting the primitive itself has no refusal semantics.

### Maintainability

- **[suggestion]** `src/work-graph.ts:630` — **33-line rationale comment guards a one-line return**
  The block from "Evidence is required of `auto` nodes only" through "It is a tool, no longer a toll." documents design history in front of `if (node.autonomy !== "auto") return;`. Future edits to this guard require reading and re-syncing a page of prose that belongs in the spec.
  Fix: Keep two lines ("HITL nodes close on the session's say-so — see docs/work-graph.md §3.2 and #499") and let docs/work-graph.md hold the argument.

- **[suggestion]** `src/work-graph.ts:632` — **Same removal rationale restated across five files**
  "it did not verify the close, it only prevented it" appears in src/work-graph.ts, the docs section ("it prevented them (#499 is the worked example"), closing.md ("the rule did not verify closes, it prevented them"), and again as a comment in test/work-graph.test.ts ("it did not verify the close, it only prevented it (#499)"). Five copies drift independently on the next change.
  Fix: Name docs/work-graph.md §3.2 as the single home; the other four sites cite it.

- **[suggestion]** `src/cli/graph.ts:783` — **runClose proposal branch does read, refuse, and ratify inline**
  The new `if (commentId !== undefined)` body reads the comment, computes `vetoed`, throws, then falls through to `selectRatification` — three concerns interleaved with two multi-paragraph comment blocks in one function body.
  Fix: Extract `resolveProposal(graph, ref, commentId, root)` returning `{proposal, reactions}` and a `assertNotRefused(reactions, root, proposal)` helper.

- **[nit]** `src/cli/graph.ts:786` — **Two adjacent comment blocks state the same rule**
  The first block says "a HITL node closes on the session's say-so (§3.2) … a root-author 👎 as a refusal"; the second, ten lines later, repeats "Ratification is no longer required; an explicit refusal is still surfaced." plus the same speed-bump argument.
  Fix: Merge into one block at the top of the branch.

- **[nit]** `src/cli/graph.ts:807` — **Unreachable fallback in the refusal message**
  `vetoed` is only true when `root?.author !== undefined`, so `${root?.author ?? "the graph root's author"}` can never take the fallback arm.
  Fix: Bind the resolved author to a local inside the `vetoed` check and interpolate that.

- **[nit]** `test/cli-graph.test.ts:480` — **HITL store fixture copy-pasted across three new tests**
  The identical three-line `new FakeStore().seed("495", …).seed("530", { node: { id: "530", title: "hitl", autonomy: "approve", checkpointId: "cp-530" }, parent: "495" })` appears in the bare-close, veto-depth, and say-so tests.
  Fix: Add a `hitlStore()` helper next to `autoNode`.

### HonestOracle

- **[blocker]** `docs/work-graph.md:380` — **Cited #548 comment argues against the change it is cited to support**
  The spec's multi-party justification ends "with the reasoning in the closing comment of [#548]" — that comment actually says "The #499 deadlock was never evidence that solo deployments need a weaker gate; it was evidence that #511 is load-bearing" and closes with "Superseded by #511." A reader following the canonical citation finds a rejection of the removal, not its rationale.

- **[important]** `src/work-graph.ts:640` — **"Nobody else could ratify" contradicted by #496's existing credential split**
  The core comment justifies the unconditional removal with "where one person walks a map the ratification gate named none — nobody else could ratify"; #511's body records that #496 already achieved "authorship separation — proposal by `ivy-agent`, ratification by `jcfischer`, both API-attested" on this deployment. The alternative (finish the credential split) is never engaged anywhere in the diff or description.

- **[suggestion]** `src/cli/graph.ts:793` — **Stated principle refuted fifteen lines later in the same hunk**
  The autonomy check is deleted because "a distinction living in one consumer is worse than no distinction at all" — and the same hunk then adds the refusal rule, which the docs admit "lives in the CLI, not in `assertClosable`." The comment states a principle the commit itself violates, misleading anyone who reads it as the codebase's rule.


---

jcfischer (commented):

9 finding(s): 2 important, 5 suggestion, 2 nit.


### CodeQuality

- **[important]** `src/cli/graph.ts:803` — **Refusal predicate fails open on unresolved root author**
  `root?.author !== undefined && reaction.author === root.author` means a failed root walk yields no refusal at all — a 👎 is silently ignored rather than reported. The failure is indistinguishable from "nobody refused" in the CLI output.
  Fix: When a 👎 exists but `root?.author` is undefined, emit a warning to stderr naming the unresolved root instead of proceeding silently.

- **[suggestion]** `src/cli/graph.ts:800` — **Refusal counts the proposer's own 👎; selectRatification excludes it**
  `reactions.some((reaction) => reaction.content === "-1" && ... reaction.author === root.author)` applies no proposer exclusion, while the very next line's `selectRatification(reactions, proposal.author, root?.author)` does. A root author who proposes and then 👎s their own comment refuses their own close.
  Fix: Mirror selectRatification's proposer guard: `reaction.author !== proposal.author && reaction.author === root?.author`.

- **[suggestion]** `src/cli/graph.ts:795` — **Auto-node refusal path is untested for the no-reaction case**
  The new comment says "Supplying a proposal on an `auto` node is unusual rather than wrong: its reactions are read on the same terms", and tests cover 👎 and 👍 on auto nodes — but not `--proposal-comment` with zero reactions, which is the path where `proposal` is now recorded on an auto receipt for the first time.
  Fix: Add an auto-node case with a proposal comment and no reactions, asserting the receipt records the proposal and still closes on probed evidence.

- **[nit]** `test/cli-graph.test.ts:508` — **Re-propose test asserts a count, not which comment closed**
  `expect(store.closed).toHaveLength(1)` after closing with `c2` does not verify the receipt recorded `c2` rather than the refused `c1`; the test would pass if the wrong comment id were persisted.
  Fix: Assert `store.closed[0].receipt.attestationFacts?.proposal?.commentId` equals `"c2"`.

### Security

- **[suggestion]** `src/cli/graph.ts:803` — **Refusal check fails open when root author is unresolved**
  The predicate `reaction.content === "-1" && root?.author !== undefined && reaction.author === root.author` evaluates to false for every reaction when `root?.author` is undefined, so a close proceeds silently where the refusal signal cannot be attributed. As a security control this is fail-open, though the diff and docs correctly describe it as a reminder rather than a gate.
  Fix: When `root?.author` is undefined but a `-1` reaction exists, emit a warning on stderr naming the unattributable refusal instead of closing silently.

### Maintainability

- **[suggestion]** `test/cli-graph.test.ts:495` — **Auto-node test bundles two independent scenarios**
  `a proposal comment on an auto node is read on the same terms` builds `autoGraph()` twice, runs a refusal scenario, then a second `const ratified = autoGraph()` block asserting the ratified path; a failure in either half reports as one test. Split into refusal and ratification cases so the failing scenario is named.
  Fix: Split into two tests: "a root-author 👎 refuses an auto close" and "a 👍 on an auto node records approved alongside probed".

- **[nit]** `test/cli-graph.test.ts:512` — **Three-line HITL fixture copy-pasted across new tests**
  The identical `new FakeStore().seed("495", …).seed("530", { node: { id: "530", title: "hitl", autonomy: "approve", checkpointId: "cp-530" }, parent: "495" })` block appears verbatim in three new tests plus the rewritten bare-close test. `autoGraph()` already exists as the parallel helper for the auto case.
  Fix: Add a `hitlGraph()` helper next to `autoGraph()` and call it from the four sites.

### HonestOracle

- **[important]** `docs/work-graph.md:384` — **Cited commit does not contain the attempt it is cited for**
  The spec promises the failed derivations are "cited where they actually live: **graph-root authorship** ... and **one reachable credential** ..., both on branch `feat/self-ratification` at `b8acfe4`" — but `b8acfe4` is the commit that *deleted* graph-root authorship (`- if (rootAuthor === undefined || proposalAuthor !== rootAuthor) return undefined;` → `+ if (!sessionIsSingleCredential) return undefined;`); that attempt lives at `4950dae`, and the first credential-count attempt at `8cdab3a`. A reader auditing whether "not derivable" is earned lands on the removal, not the attempt.

- **[suggestion]** `PR description:1` — **Stated test count is not this head's**
  "`bun test` — 2219 pass / 0 fail" does not match HEAD `cd2ceef`: I ran it and got 2221 pass / 0 fail, so the quoted run predates commits `380b3b9` (which changed `--proposal-comment` semantics for `auto` nodes) and `cd2ceef` (which pinned them) — the two the description flags "up front". The head is green (`bunx tsc --noEmit` clean too); the figure certifying it is stale.


---

PR #550, opened 2026-08-06
feat(skills): skillsLoading on the adapter spec; stubs for eager loaders (#542)

jcfischer:
Implements #542. Design and rationale in `docs/progressive-skill-loading.md`
DD-A and DD-B (#545, which should land first — this PR's commit message points
at it).

## Problem

A substrate whose skill loader reads every projected `SKILL.md` at session start
pays the whole corpus before the first turn — **~200K tokens** across the current
114-skill home. Claude Code does not pay it (its Skill tool loads bodies on
demand). pi-dev does. Same registry, same symlinks, different loader — and Soma
had no way to express which behaviour a substrate has.

## Change

`SkillsLoadingMode` on `SubstrateInstallSpec`, beside the `skillsLoaderDir` it
already owns:

```ts
export type SkillsLoadingMode = "on-demand" | "eager";
```

| Substrate | Value | Loader slot |
| --- | --- | --- |
| claude-code, codex, cursor, grok, anthropic-cowork | `on-demand` | symlink to the body (unchanged) |
| pi-dev | `eager` | generated frontmatter-only stub |

A stub is a directory holding a `SKILL.md` with the source frontmatter verbatim,
a provenance marker, and a pointer to the real body under `~/.soma/skills/`. The
loader can still list and route the skill; the body arrives only when the
substrate resolves the pointer.

Nothing changes for on-demand substrates, and no skill file changes at all.

## Three things that were easy to get wrong

**Provenance goes after the frontmatter.** YAML frontmatter is only frontmatter
when it is the first thing in the file, so the usual leading `withProvenance`
header would have made every stub unparseable by the exact loader it is written
for.

**A stub is a real directory, so `!isSymbolicLink()` no longer means "not
ours".** `ensureSymlink`'s data-loss guard rests on that equivalence. Ownership
is decided by the stub marker instead: a directory carrying it is a previous
projection and is replaced; anything else is user data and still needs `force`.

**`removeLink` needed the same rule.** Without it, `unproject-skill` would have
demanded `--force` to delete a directory Soma itself had just generated — a
regression this PR would otherwise have shipped. Covered by a test.

Plans carry `kind: "symlink" | "stub"` from the same branch apply takes, so a dry
run cannot claim a link where apply writes a stub.

## Tests

9 new cases in `test/skill-projection.test.ts`: stub shape and content, registry
slot still a symlink, mixed on-demand/eager run, byte-idempotent reprojection
reporting `unchanged`, replacement when source frontmatter changes, the
force/no-force guard both ways, unproject removing its own stub without
`--force`, unproject still refusing a non-stub directory, and plan/apply
agreement on `kind`.

## Verification

- `bun test` — **2226 pass, 0 fail** across 146 files.
- `bun run lint` — 52 errors, all pre-existing on `main`. Two flagged files are
  also touched here (`adapters/anthropic-cowork/install.ts`,
  `adapters/grok/install.ts`), but the errors sit at lines 80 and 135–148 while
  the insertions are single lines at 96 and 188. No changed line is flagged.

## Out of scope

Per DD-D/DD-E and #542's scope section: the router, `soma-skill.json`, and the
pi-dev side of trigger-time promotion — which must deliver the body as a **tool
result, never a system-prompt edit** (DD-C), or promotion costs more than it
saves by invalidating the cached prefix.

### DIFF
--- a/src/adapters/anthropic-cowork/install.ts
+++ b/src/adapters/anthropic-cowork/install.ts
@@ -93,6 +93,7 @@ export const anthropicCoworkInstallSpec: SubstrateInstallSpec<"anthropic-cowork"
   defaultHome: ANTHROPIC_COWORK_DEFAULT_HOME,
   homeFiles: ANTHROPIC_COWORK_HOME_FILE_PATHS,
   skillsLoaderDir: skillsLoaderUnder(ANTHROPIC_COWORK_SKILLS_ROOT_PATH),
+  skillsLoading: "on-demand",
   vsaSkillProjection: {
     destinationDir: (substrateHome) => resolve(substrateHome, "skills", "VSA"),
   },
--- a/src/adapters/claude-code/install.ts
+++ b/src/adapters/claude-code/install.ts
@@ -58,6 +58,7 @@ export const claudeCodeInstallSpec: SubstrateInstallSpec<"claude-code"> = {
       : []),
   ],
   skillsLoaderDir: skillsLoaderUnder(),
+  skillsLoading: "on-demand",
   vsaSkillProjection: {
     destinationDir: vsaSkillUnder(),
     // soma#329: before reprojecting VSA, prune a sibling renamed-away "ISA" skill
--- a/src/adapters/codex/install.ts
+++ b/src/adapters/codex/install.ts
@@ -68,6 +68,7 @@ export const codexInstallSpec: SubstrateInstallSpec<"codex"> = {
   // Owned (Soma-exclusive) dir — see ownedSubtrees JSDoc. (hooks/ + skills/ are shared.)
   ownedSubtrees: ["memories/soma"],
   skillsLoaderDir: skillsLoaderUnder(),
+  skillsLoading: "on-demand",
   vsaSkillProjection: {
     destinationDir: vsaSkillUnder(),
     // soma#329: before reprojecting VSA, prune a sibling renamed-away "ISA" skill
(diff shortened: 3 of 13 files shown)


jcfischer:
Review pass on the blast radius, since the change makes a loader slot something other than a symlink for the first time.

**Contained.** `skillsLoaderDir` is referenced only by the install specs and `skill-projection.ts`, and `skill-projection.ts` is the only file in `src/` that calls `readlink`. Nothing else in the tree reads a loader slot and assumes a link.

**Reconciliation will not eat the stubs.** pi-dev declares `ownedSubtrees: ["agent/soma"]`, and its own comment records that `agent/skills` is shared. Owned-subtree reconciliation removes files Soma no longer emits inside owned trees only, so stub dirs under `agent/skills/<name>/` are out of its reach. `pruneLegacyVsaSkill` matches the legacy ISA/VSA dir by name and is likewise unaffected.

**One coexistence worth recording.** pi-dev's kernel skill is projected by a different path — `vsaSkillProjection.destinationDir` resolves to `agent/skills/soma` (`src/adapters/pi-dev/skill-projection.ts:56`) and copies a *full* body. So after this change an eager loader dir is deliberately mixed: one full kernel skill plus N stubs. That is the intended shape — the kernel is the context that should load by default ("Always-On Kernel" in the design doc) — but it is not obvious from either code path alone, and a future reader could easily read it as an inconsistency and "fix" it.

PR #554, opened 2026-08-06
fix(work-graph): a ratification is a reaction; amendment is caller discipline (#525)

### LINKED ISSUE #525: Comment-based ratification: the amended-proposal path §3.2 specifies
## Task

Surfaced by #498. §3.2 names **two** admissible ratifications for a HITL close:

> the receipt is the principal's 👍 reaction on that specific comment ID — **or a
> principal-authored comment, which wins when amending** — verified via the API
> author field.

#498 shipped only the reaction path. The comment path is unimplemented, so an
amended proposal that the principal ratifies *by replying* rather than by
reacting cannot close, and the "wins when amending" precedence has no code
behind it.

### What it needs

1. A read of the node's comments after the proposal — the seam has
   `postComment`, `readComment` and `readCommentReactions`, but no
   `listComments`. That is one more §2.5 addendum (same class as
   `CreateNodeSpec.body`/`parent` on #497 and `readComment` on #498).
2. The precedence rule: when both a 👍 and a later principal-authored comment
   exist on the same proposal, the comment wins.
3. The amendment rule it exists to serve (§3.2, the replay-rebind lesson): a
   materially amended proposal is re-posted and needs *fresh* ratification —
   so a ratification must be bound to the proposal comment it answers, not to
   the node.

### Scope

Not a blocker for the ships-with-consumer merge: reaction ratification is the
primary path and works end to end. This closes the gap between what §3.2
specifies and what the verbs implement.

Node: `kind: task` · `autonomy: auto` (branch work; the merge rides normal PR
review) · probes (close evidence): `bun test` exit 0 including a test that a
principal-authored comment outranks a 👍 on the same proposal and a test that a
re-posted proposal does not inherit the prior ratification, `bunx tsc --noEmit`
exit 0.

jcfischer:
Resolves #525.

#525 asked for §3.2's second ratification path — a principal-authored comment, outranking the 👍 when amending. **The §3.2 it quoted no longer exists**: #549 (`8ba3378`) rewrote the section and dropped the clause.

So this withdraws the path rather than leaving it unimplemented:

- **Why not build it.** With ratification demoted from gate to label, its only remaining job is feeding conjuncts 3 and 4 of the attestation derivation. Reading approval out of free prose would let a root-author reply of "hold on, not this" derive `verified` — on the one field the phase-2 close auditor recomputes. A 👍 is a deliberate, unambiguous gesture; prose is not.
- **The dead type goes, and gets a name.** `ratification.kind` was `"reaction" | "comment"` spelled inline in three modules; nothing ever produced `"comment"`. It is now a single exported `Ratification` type — dropping one member took a three-file edit, and the next shape change would have taken another.
- **Amendment is caller discipline, and the docs now say so.** An amended proposal inherits nothing *when the caller passes the new id*; passing the superseded id ratifies from the superseded proposal, and nothing in the runtime rejects it. Same root as the already-documented "nothing binds a new proposal to a refused one" — no verb binds a proposal to the one it supersedes. Earlier revisions of this PR called that a structural guarantee, which the shipped §3.2 contradicts.
- **The invariant is pinned by a two-arm test.** Arm A closes against the original proposal and asserts a ratified `verified` receipt — establishing the 👍 is admissible in that exact store. Arm B amends, closes against the new comment, and asserts the close declines to reach it, with the original reaction still present and unretracted. Deleting either seed fails the test; a single-arm version did not, because every assertion followed from the amended comment simply having no reactions.

No `listComments` seam addendum. No behaviour change beyond the type narrowing.

`bun test` 2264 pass / 0 fail **on the PR head** (`70a17b3`) · `bunx tsc --noEmit` clean.

### DIFF
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -360,8 +360,10 @@ What remains:
   a refused one; and it needs the root author to resolve, so where the root walk
   fails nothing refuses, silently.
 
-- **A 👍 reaction is the only ratification (#525).** Literally that reaction:
-  `selectRatification` matches on `content === "+1"`, so no other emoji ratifies.
+- **A 👍 reaction is the only ratification (#525).** Literally that reaction —
+  matched on `+1`, so no other emoji ratifies, and (per the rule above) any
+  non-proposer's counts, with *whose* it is affecting `attestation` rather than
+  admissibility.
   This section once admitted a second form — a principal-authored *comment*,
   winning over the 👍 when amending — and the seam carried a `"comment"`
   ratification kind nothing produced. It is withdrawn, not unimplemented: with
@@ -374,9 +376,9 @@ What remains:
   guarantee**. Ratification is read from the comment id passed to
   `--proposal-comment`, so an amended proposal inherits nothing *when the new id
   is the one passed*; passing the superseded id ratifies from the superseded
-  proposal, and nothing in the runtime rejects it. Same class of limit as "a
-  fresh proposal carries no refusal" above, and it has the same root: no verb
-  binds a proposal to the one it supersedes.
+  proposal, and nothing in the runtime rejects it. Same class of limit as the
+  refusal one named above — nothing binds a new proposal to a refused one — and
+  it has the same root: no verb binds a proposal to the one it supersedes.
 
 - **The receipt distinguishes the two cases.** A ratified close records the
   proposal and ratifier; a bare one records their absence in
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -360,15 +360,23 @@ What remains:
   a refused one; and it needs the root author to resolve, so where the root walk
   fails nothing refuses, silently.
 
-- **A reaction is the only ratification (#525).** This section once admitted a
-  second form — a principal-authored *comment*, winning over the 👍 when
-  amending — and the seam carried a `"comment"` ratification kind nothing
-  produced. It is withdrawn, not unimplemented: with ratification demoted from
-  gate to label, deriving approval from free prose would let a root-author reply
-  of "hold on, not this" produce `verified` on the one field the phase-2 auditor
-  recomputes. A 👍 is a deliberate, unambiguous gesture; prose is not. The
-  amendment rule it served survives structurally — a re-posted proposal is a new
-  comment id with no reactions, so it inherits nothing.
+- **A 👍 reaction is the only ratification (#525).** Literally that reaction:
+  `selectRatification` matches on `content === "+1"`, so no other emoji ratifies.
+  This section once admitted a second form — a principal-authored *comment*,
+  winning over the 👍 when amending — and the seam carried a `"comment"`
+  ratification kind nothing produced. It is withdrawn, not unimplemented: with
+  ratification demoted from gate to label, deriving approval from free prose
+  would let a root-author reply of "hold on, not this" produce `verified` on the
+  one field the phase-2 auditor recomputes. A 👍 is a deliberate, unambiguous
+  gesture; prose is not.
+
+  The amendment rule it served is **caller discipline, not a structural
+  guarantee**. Ratification is read from the comment id passed to
+  `--proposal-comment`, so an amended proposal inherits nothing *when the new id
+  is the one passed*; passing the superseded id ratifies from the superseded
+  proposal, and nothing in the runtime rejects it. Same class of limit as "a
+  fresh proposal carries no refusal" above, and it has the same root: no verb
+  binds a proposal to the one it supersedes.
 
 - **The receipt distinguishes the two cases.** A ratified close records the
   proposal and ratifier; a bare one records their absence in
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -360,6 +360,16 @@ What remains:
   a refused one; and it needs the root author to resolve, so where the root walk
   fails nothing refuses, silently.
 
+- **A reaction is the only ratification (#525).** This section once admitted a
+  second form — a principal-authored *comment*, winning over the 👍 when
+  amending — and the seam carried a `"comment"` ratification kind nothing
+  produced. It is withdrawn, not unimplemented: with ratification demoted from
+  gate to label, deriving approval from free prose would let a root-author reply
+  of "hold on, not this" produce `verified` on the one field the phase-2 auditor
+  recomputes. A 👍 is a deliberate, unambiguous gesture; prose is not. The
+  amendment rule it served survives structurally — a re-posted proposal is a new
+  comment id with no reactions, so it inherits nothing.
+
 - **The receipt distinguishes the two cases.** A ratified close records the
   proposal and ratifier; a bare one records their absence in
   `attestationFacts.reasons` ("no proposal comment recorded", "no ratification
--- a/src/cli/graph.ts
+++ b/src/cli/graph.ts
@@ -655,7 +655,7 @@ export function selectRatification(
   reactions: readonly Reaction[],
   proposalAuthor: string,
   rootAuthor: string | undefined,
-): { kind: "reaction"; id: string; author: string } | undefined {
+): Ratification | undefined {
   if (
     rootAuthor !== undefined &&
     reactions.some((reaction) => reaction.content === "-1" && reaction.author === rootAuthor)
--- a/src/cli/graph.ts
+++ b/src/cli/graph.ts
@@ -41,6 +41,7 @@ import {
   type NodeState,
   type Probe,
   type ProbeResult,
+  type Ratification,
   type Reaction,
   type WorkGraphEvidenceKind,
 } from "../work-graph";
@@ -781,7 +782,7 @@ async function runClose(
   const root = await findGraphRoot(ref, async (nodeRef) => await graph.readNode(nodeRef));
 
   let proposal: { commentId: string; author: string } | undefined;
-  let ratification: { kind: "reaction"; id: string; author: string } | undefined;
+  let ratification: Ratification | undefined;
 
   // A proposal comment is optional now: a HITL node closes on the session's
   // say-so (§3.2). When one IS supplied, its reactions still carry weight —
--- a/src/cli/graph.ts
+++ b/src/cli/graph.ts
@@ -781,7 +781,7 @@ async function runClose(
   const root = await findGraphRoot(ref, async (nodeRef) => await graph.readNode(nodeRef));
 
   let proposal: { commentId: string; author: string } | undefined;
-  let ratification: { kind: "reaction" | "comment"; id: string; author: string } | undefined;
+  let ratification: { kind: "reaction"; id: string; author: string } | undefined;
 
   // A proposal comment is optional now: a HITL node closes on the session's
   // say-so (§3.2). When one IS supplied, its reactions still carry weight —
--- a/src/index.ts
+++ b/src/index.ts
@@ -711,6 +711,7 @@ export {
   type Probe,
   type ProbeResult,
   type ProbeType,
+  type Ratification,
   type Reaction,
   type WorkGraphAutonomy,
   type WorkGraphErrorCode,
--- a/src/skills/orienteer/references/closing.md
+++ b/src/skills/orienteer/references/closing.md
@@ -121,10 +121,13 @@ Two things worth knowing when you do use the proposal flow:
   that it was not the right human.
 - **Ratification binds to a comment id, not to its text.** Nothing hashes the
   proposal body, so a proposal that is ratified and *then edited* still closes on
-  that 👍. If the resolution changes materially, post a new comment — the new one
-  carries no reactions, so it needs fresh ratification and inherits nothing.
-- **A 👍 is the only ratification.** Replying "yes, go ahead" ratifies nothing;
-  the receipt will read *no ratification found*. An older §3.2 also admitted a
+  that 👍. If the resolution changes materially, post a new comment — and then
+  **pass the new id**. Amending is your discipline, not the runtime's: nothing
+  binds a proposal to the one it supersedes, so a close pointed at the old id
+  still ratifies from the old 👍.
+- **A 👍 is the only ratification** — that reaction specifically, matched on
+  `+1`; no other emoji counts. Replying "yes, go ahead" ratifies nothing, and the
+  receipt will read *no ratification found*. An older §3.2 also admitted a
   principal-authored comment, outranking the 👍 when amending, and #525 was to
   implement it. It was dropped instead: once ratification stopped gating a HITL
   close (#549) and became a label feeding `attestation`, reading approval out of
--- a/src/skills/orienteer/references/closing.md
+++ b/src/skills/orienteer/references/closing.md
@@ -121,7 +121,15 @@ Two things worth knowing when you do use the proposal flow:
   that it was not the right human.
 - **Ratification binds to a comment id, not to its text.** Nothing hashes the
   proposal body, so a proposal that is ratified and *then edited* still closes on
-  that 👍. If the resolution changes materially, post a new comment.
+  that 👍. If the resolution changes materially, post a new comment — the new one
+  carries no reactions, so it needs fresh ratification and inherits nothing.
+- **A 👍 is the only ratification.** Replying "yes, go ahead" ratifies nothing;
+  the receipt will read *no ratification found*. An older §3.2 also admitted a
+  principal-authored comment, outranking the 👍 when amending, and #525 was to
+  implement it. It was dropped instead: once ratification stopped gating a HITL
+  close (#549) and became a label feeding `attestation`, reading approval out of
+  free prose would let a reply of "hold on, not this" derive `verified`. React,
+  don't reply.
 
 ## Attestation
 
--- a/src/skills/orienteer/references/closing.md
+++ b/src/skills/orienteer/references/closing.md
@@ -114,7 +114,7 @@ So the honest reading is that being refused is *recorded*, not *prevented*. If
 you close around a 👎, the receipt shows a close with no refusal in it, and the
 person who said no finds out by reading the node.
 
-Two things worth knowing when you do use the proposal flow:
+Three things worth knowing when you do use the proposal flow:
 
 - **Any non-proposer's 👍 is taken as the ratification**, root author preferred.
   On a public tracker that includes a stranger's. Only `attestation` records
@@ -126,13 +126,14 @@ Two things worth knowing when you do use the proposal flow:
   binds a proposal to the one it supersedes, so a close pointed at the old id
   still ratifies from the old 👍.
 - **A 👍 is the only ratification** — that reaction specifically, matched on
-  `+1`; no other emoji counts. Replying "yes, go ahead" ratifies nothing, and the
-  receipt will read *no ratification found*. An older §3.2 also admitted a
-  principal-authored comment, outranking the 👍 when amending, and #525 was to
-  implement it. It was dropped instead: once ratification stopped gating a HITL
-  close (#549) and became a label feeding `attestation`, reading approval out of
-  free prose would let a reply of "hold on, not this" derive `verified`. React,
-  don't reply.
+  `+1`; no other emoji counts, and per the first bullet *whose* 👍 it is affects
+  `attestation`, not whether it ratifies. Replying "yes, go ahead" ratifies
+  nothing, and the receipt will read *no ratification found*. The HITL section of
+  the spec once admitted a principal-authored comment too, outranking the 👍 when
+  amending, and #525 was to implement it. It was dropped instead: once
+  ratification stopped gating a HITL close (#549) and became a label feeding
+  `attestation`, reading approval out of free prose would let a reply of "hold
+  on, not this" derive `verified`. React, don't reply.
 
 ## Attestation
 
--- a/src/work-graph-attestation.ts
+++ b/src/work-graph-attestation.ts
@@ -28,6 +28,7 @@ import type {
   ConfinementProbeRecord,
   NodeRef,
   NodeState,
+  Ratification,
 } from "./work-graph";
 import type { CommandOutcome, CommandRequest } from "./work-graph-probes";
 
@@ -133,8 +134,7 @@ export interface AttestationInputs {
   actingIdentity: string;
   confinement?: ConfinementResult;
   proposal?: { commentId: string; author: string };
-  /** Reactions only — see {@link AttestationFacts.ratification} for why the comment path is not here. */
-  ratification?: { kind: "reaction"; id: string; author: string };
+  ratification?: Ratification;
   root?: { nodeId: string; author: string };
 }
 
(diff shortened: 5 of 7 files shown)


jcfischer (commented):

5 finding(s): 1 blocker, 1 important, 2 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/work-graph.ts:246` — **`kind` is now a discriminant with one variant**
  `ratification?: { kind: "reaction"; id: string; author: string }` leaves a field every producer must write and no consumer can branch on; the evidence summary that interpolates it (`via ${ratification.kind}`) is now a constant string.
  Fix: Either drop `kind` and let the field's presence mean "reaction", or note in the docstring that it is retained deliberately as a receipt self-description field.

### Maintainability

- **[nit]** `src/work-graph.ts:233` — **Withdrawal rationale duplicated across three artifacts**
  The same argument — "inferring approval from free prose would let a root-author reply of 'hold on, not this' derive `verified`" — is restated near-verbatim in `work-graph.ts`, `docs/work-graph.md` ("deriving approval from free prose would let a root-author reply of \"hold on, not this\" produce `verified`"), and `closing.md`. Three copies drift independently when the rule changes.
  Fix: Keep the full argument in `docs/work-graph.md` §3.2; reduce the TSDoc to one line plus a pointer to that section, as `work-graph-attestation.ts` already does with its `{@link}`.

### HonestOracle

- **[blocker]** `test/cli-graph.test.ts:534` — **Test cannot detect the invariant it claims to pin**
  The comment claims "nothing would catch it becoming false ... This is the pin", but every assertion (`ratification` undefined, no `approved` evidence, `unverified`, "no ratification found") follows solely from c2 carrying no reactions; `proposal.commentId).toBe("c2")` merely echoes the id the test itself passed. Seeding the 👍 on c1 is decorative — the test would pass identically if the second `--propose` had edited c1 in place and left the ratification live, which is the exact regression the PR cites as its reason not to implement #525.

- **[important]** `docs/work-graph.md:368` — **Caller discipline documented as a structural guarantee**
  "a re-posted proposal is a new comment id with no reactions, so it inherits nothing" holds only if the operator passes the *new* id; the test's own note concedes "pointing the close at the stale comment id fails it", and nothing in the diff rejects a superseded comment id. A reader of §3.2 is told inheritance is impossible when it is one stale `--proposal-comment` argument away from deriving `verified` on the field the phase-2 auditor recomputes.

- **[suggestion]** `src/skills/orienteer/references/closing.md:125` — **"A 👍" vs "a reaction" — docs disagree on what ratifies**
  closing.md asserts "**A 👍 is the only ratification.**" while docs/work-graph.md says "**A reaction is the only ratification (#525).**" and the narrowed type carries only `kind: "reaction"`; nothing in the diff shows the emoji content is constrained. The orienteer guidance promises a specificity the seam as shown does not encode.


---

jcfischer (commented):

6 finding(s): 1 important, 4 suggestion, 1 nit.


### Architecture

- **[suggestion]** `src/cli/graph.ts:781` — **Ratification shape declared inline in three modules**
  `{ kind: "reaction"; id: string; author: string }` is re-declared verbatim in `src/cli/graph.ts`, `AttestationInputs` (`src/work-graph-attestation.ts`), and `AttestationFacts` (`src/work-graph.ts`) — this one-member narrowing required editing all three. The CLI layer restating a core domain shape means the next field change touches three sites with no compiler link between them.
  Fix: Export `interface Ratification { kind: "reaction"; id: string; author: string }` from `src/work-graph.ts` and reference it from the attestation module and the CLI.

### Maintainability

- **[suggestion]** `src/work-graph.ts:234` — **Nine-line doc comment restates PR narrative on a one-member union**
  The comment on `ratification?: { kind: "reaction"; ... }` spans nine lines covering #525's history, #549's demotion, and the "hold on, not this" hypothetical — all of which already live in `docs/work-graph.md` §3.2. Historical narrative in a type declaration goes stale independently of the code it annotates.
  Fix: Reduce to one line plus a pointer: `/** Reactions only — see docs/work-graph.md §3.2 (#525) for why the comment path was withdrawn. */`

- **[nit]** `test/cli-graph.test.ts:534` — **Eight-line test preamble duplicates the doc rationale again**
  The comment block ("#525 was going to add a rule; #549 removed the gate it would have served…") is a fourth copy of the same withdrawal story already in `work-graph.ts`, `work-graph.md`, and `closing.md`. Test comments should say what the test pins, not re-litigate the decision.
  Fix: Keep the first two lines (the amendment rule + why it holds structurally); drop the #525/#549 history.

### HonestOracle

- **[important]** `docs/work-graph.md:366` — **Withdrawal justified against a proposal #525 didn't make**
  The load-bearing argument — "deriving approval from free prose would let a root-author reply of 'hold on, not this' produce `verified`" — assumes the withdrawn path would have read *any* prose, but the quoted clause described "a principal-authored *comment*, winning over the 👍 when amending", which an explicit marker (a required token, a structured body) satisfies without prose inference. The counter-design is never engaged, yet this reasoning is now recorded in docs, source comments, and the skill reference as settled.

- **[suggestion]** `src/work-graph-attestation.ts:136` — **In-repo typecheck sold as safety of a narrowed exported type**
  `export interface AttestationInputs` and `export interface AttestationFacts` lose a union member, and the only evidence offered is "`bunx tsc --noEmit` clean" plus "No behaviour change beyond the type narrowing" — a check that sees this repo's call sites only. Any external consumer branching on or constructing `kind: "comment"` breaks, and nothing in the PR states whether these types are published.

- **[suggestion]** `src/skills/orienteer/references/closing.md:126` — **Receipt string promised unconditionally, true in one branch**
  "Replying 'yes, go ahead' ratifies nothing; the receipt will read *no ratification found*" holds only when a proposal comment was passed; the sibling doc records the other branch as "no proposal comment recorded". An agent following this reference and not finding the promised string has no way to tell which case it is in.


---

jcfischer:
Round 1 addressed in `378e19a` — all five findings, one pass.

**[blocker] the test could not detect the invariant it claimed to pin.** Correct, and worse than the review states: I checked by deleting the seeded 👍 and the test passed unchanged, so the seed was inert. It is now two arms over the *same* reaction — arm A closes against `c1` and asserts a ratified `verified` receipt (establishing the reaction is admissible in that exact store), arm B amends and closes against `c2` and asserts the close declines to reach it, with `c1`'s reaction asserted still present and unretracted. Deleting **either** seed now fails the test; verified both ways.

**[important] caller discipline documented as a structural guarantee.** Agreed. §3.2 now says the amendment rule is discipline, not a guarantee: inheritance is prevented by *which id you pass*, and a close pointed at the superseded id ratifies from the superseded proposal with nothing rejecting it. Tied to the limit it shares a root with — no verb binds a proposal to the one it supersedes. `closing.md` carries the same correction ("pass the new id").

**[suggestion] "A 👍" vs "a reaction".** The narrow claim is the true one: `selectRatification` matches `content === "+1"`, so no other emoji ratifies. §3.2 was the loose sentence, and now says so and names the check.

**[suggestion] `kind` is a one-variant discriminant.** Retained deliberately, and the docstring now says why rather than leaving it to be inferred: a receipt is a published artifact re-read years later, and *how* it was ratified is part of what it records — a bare `{id, author}` leaves a future reader inferring the mechanism from an id shape. Nothing branches on it today and nothing should need to.

**[nit] rationale triplicated.** The TSDoc is now one line plus a pointer to §3.2, matching what `work-graph-attestation.ts` already does. The full argument lives in §3.2 only.

`bun test` 2264 pass / 0 fail **on the PR head** (`378e19a`) · `bunx tsc --noEmit` clean.

jcfischer (commented):

9 finding(s): 1 important, 5 suggestion, 3 nit.


### CodeQuality

- **[nit]** `src/skills/orienteer/references/closing.md:117` — **Lead-in still says "Two things" above three bullets**
  The diff adds a third bullet ("**A 👍 is the only ratification**") under the unchanged lead-in `Two things worth knowing when you do use the proposal flow:` — base had exactly two bullets there.
  Fix: Change to "Three things worth knowing…" (or drop the count).

- **[nit]** `docs/work-graph.md:376` — **Quoted back-reference doesn't appear above**
  `Same class of limit as "a fresh proposal carries no refusal" above` presents a verbatim quotation, but the passage above reads "nothing binds a new proposal to a refused one" — the quoted phrase is a test name, not doc text, so a reader searching for it finds nothing.
  Fix: Reference the passage unquoted, or quote the wording that actually appears in §3.2.

### Architecture

- **[suggestion]** `src/work-graph.ts:236` — **New code comment pins to a section number known to rot**
  The comment says "§3.2 has the argument for why the comment path was withdrawn (#525)" — the same anchor the PR body reports as already broken ("The §3.2 it quoted no longer exists: #549 rewrote the section"). closing.md adds a second §3.2 reference in the same change.
  Fix: Anchor to the issue/PR (#525, #549) or a stable heading slug rather than a section number.

- **[suggestion]** `docs/work-graph.md:361` — **Spec names a private implementation symbol**
  The spec now states "`selectRatification` matches on `content === \"+1\"`, so no other emoji ratifies", making CONTEXT.md's declared spec doc depend on an internal function name. A rename in src silently drifts the spec with no compile-time signal.
  Fix: State the rule behaviourally ("only the 👍 reaction, matched on `+1`, ratifies") without naming the function.

### Maintainability

- **[suggestion]** `src/work-graph.ts:234` — **Ratification shape triplicated across three modules**
  `{ kind: "reaction"; id: string; author: string }` is now spelled inline in `work-graph.ts:243`, `work-graph-attestation.ts:135`, and `graph.ts:781` — this PR had to edit all three to drop one union member. The next shape change repeats that three-file sweep.
  Fix: Export a named `Ratification` type from `work-graph.ts` and reference it in the other two.

- **[nit]** `test/cli-graph.test.ts:534` — **Two seeded closes share an untyped inline node literal**
  `seeded()` correctly factors the store setup, but the arms still repeat the identical four-call `--propose` / `reactions.set("c1", ...)` / close sequence with only the body string differing. Fine at two arms; a third would warrant a helper.
  Fix: Leave as-is unless a third arm appears; then extract `proposeAndRatify(store, body)`.

### HonestOracle

- **[important]** `docs/work-graph.md:371` — **PR title claims a structural guarantee its own doc denies**
  The title says "amendment inherits nothing" and the body "The amendment rule survives structurally... an amended proposal already inherits nothing", while the shipped doc says it is "**caller discipline, not a structural guarantee**" and that "passing the superseded id ratifies from the superseded proposal, and nothing in the runtime rejects it". The merge commit becomes the permanent record and states the opposite of the file it lands.

- **[suggestion]** `src/skills/orienteer/references/closing.md:127` — **"Only ratification" stated without the author condition**
  "**A 👍 is the only ratification** — that reaction specifically, matched on `+1`; no other emoji counts" gives emoji as the sole condition, yet the same PR's threat model turns on *who* reacts ("a root-author reply"), and the only evidence is Arm A's reaction authored by `jcfischer`, the root author. No test or doc line establishes whether a third party's 👍 ratifies.

- **[suggestion]** `test/cli-graph.test.ts:536` — **"Verified falsifiable" mutation check leaves no artifact**
  The description asserts "Verified falsifiable: pointing the close at the stale comment id fails it" as fact; that run exists nowhere in the diff or the suite. The claim rests on a one-off manual mutation a reader cannot reproduce from the repo.


---

jcfischer (commented):

7 finding(s): 2 important, 3 suggestion, 2 nit.


### CodeQuality

- **[suggestion]** `test/cli-graph.test.ts:573` — **Assertion on the fake's own Map can never fail**
  `expect(amended.reactions.get("c1")).toHaveLength(1); // still there — not consumed, not retracted` reads the test's own `reactions` Map, which only the test writes — `FakeStore.readCommentReactions` is read-only and no production path mutates reactions. It reads as a guarantee about the system but carries no signal.
  Fix: Drop the assertion, or keep it as a comment noting reactions are read-only by construction.

- **[nit]** `test/cli-graph.test.ts:536` — **Test names an invariant with a term used nowhere else**
  `"an amended proposal inherits no ratification — the replay-rebind invariant"` coins "replay-rebind", which appears in no other file; the docs this PR adds call the same thing "the amendment rule" and "caller discipline".
  Fix: Rename to match the docs' vocabulary, e.g. "— §3.2's amendment rule".

### Maintainability

- **[suggestion]** `test/cli-graph.test.ts:568` — **Test asserts against undeclared minted comment ids**
  `await run(["graph", "close", "530", "--proposal-comment", "c2", "--repo", REPO], amended)` hard-codes an id that FakeStore mints implicitly from propose order; nothing in the test declares that `c1`/`c2` come from the first and second `--propose`. Inserting or reordering a propose call silently repoints the assertion.
  Fix: Capture the minted id from the propose run (or expose a `store.lastCommentId`) and pass that, instead of literal `"c1"`/`"c2"`.

- **[nit]** `test/cli-graph.test.ts:565` — **Arm A and Arm B repeat propose-then-seed-👍 verbatim**
  Both arms run the identical pair `await run([... "--propose", "--body", "the resolution" ...], X)` followed by `X.reactions.set("c1", [{ id: "r1", content: "+1", author: "jcfischer" }])`; only the store name differs. The `seeded()` helper already exists but stops short of the shared setup.
  Fix: Fold the propose + reaction seeding into an `async function proposed()` that returns the store, so each arm starts from the ratified state.

### HonestOracle

- **[important]** `docs/work-graph.md:374` — **Withdrawal rationale refuted by the hole the same section admits**
  The withdrawal is justified because prose ratification "would let a root-author reply of 'hold on, not this' produce `verified` on the one field the phase-2 auditor recomputes" — then the next paragraph concedes "passing the superseded id ratifies from the superseded proposal, and nothing in the runtime rejects it", which produces the same false `verified` on the same field through the retained 👍 path. The stated ground does not separate the rejected path from the kept one, so the doc records a decision its own text undercuts.

- **[important]** `test/cli-graph.test.ts:573` — **Headline negative assertion has no arm-A counterpart**
  The test asserts its own method — "The discrimination is between the two arms; neither carries it alone" — but `expect(receipt.evidence.some((entry) => entry.kind === "approved")).toBe(false)` has no arm-A assertion that an `approved` entry ever appears on a ratified close, so it passes identically if that evidence kind is never emitted anywhere. The PR description lists "asserts no `approved` evidence" as one of three proofs the invariant is pinned.

- **[suggestion]** `test/cli-graph.test.ts:570` — **Assertion checks the fixture the test itself wrote**
  `expect(amended.reactions.get("c1")).toHaveLength(1); // still there — not consumed, not retracted` reads back the map the test populated two lines earlier via `amended.reactions.set("c1", ...)`; the close path only reads reactions, so no production behaviour can make it fail. It is presented as establishing the 👍 stayed "live and unretracted".


---

jcfischer:
Round 2 addressed in `70a17b3` — plus the PR title and body, which were the real finding.

**[important] PR title claims a structural guarantee its own doc denies.** Correct, and the most valuable finding in the round: the title said "amendment inherits nothing" and the body "survives structurally", while the file being merged says *caller discipline, not a structural guarantee*. The squash commit is the permanent record, so the claim changed, not the doc. Title is now "…amendment is caller discipline"; the body states the limit and names the earlier revisions as having got it wrong.

**[suggestion] ratification shape triplicated.** Fixed properly — `Ratification` is a named exported type in `work-graph.ts`, referenced from `work-graph-attestation.ts` and `cli/graph.ts`. The observation that dropping one member cost a three-file edit is exactly why it earns a name.

**[suggestion] spec names a private symbol.** Fixed — §3.2 states the rule behaviourally ("matched on `+1`") without naming `selectRatification`, which a rename would have drifted silently.

**[suggestion] "only ratification" stated without the author condition.** Fixed in both files: any non-proposer's 👍 ratifies, and *whose* it is affects `attestation` rather than admissibility. That was already the first bullet of the same `closing.md` list and already has a test ("a 👍 from someone other than the root author closes, but never as verified") — the new bullet just needed to not read as a narrower rule than the one above it.

**[nit] "Two things" over three bullets** — fixed. **[nit] quoted back-reference that appears nowhere** — fixed; it now quotes wording that is actually in the section rather than a test name.

**[suggestion] §-number anchors rot.** Partially taken. The TSDoc and `closing.md` now refer to the HITL receipts *section* by name rather than by number, so a renumber doesn't strand them. Not converted wholesale to issue anchors: every other cross-reference in these files is a § number, and one file using a different convention is worse than the rot it avoids. Worth doing repo-wide or not at all.

**[nit] test arms repeat a short setup sequence.** Left as-is, per the finding's own guidance — a helper at a third arm, not before.

**[suggestion] "verified falsifiable" left no artifact.** Fair, and now moot: the two-arm test *is* the artifact. Deleting either seeded reaction fails the suite, which is reproducible from the repo — unlike the one-off manual mutation the earlier body cited. The body no longer claims a run a reader cannot repeat.

`bun test` 2264 pass / 0 fail **on the PR head** (`70a17b3`) · `bunx tsc --noEmit` clean.

jcfischer (commented):

4 finding(s): 4 suggestion.


### CodeQuality

- **[suggestion]** `src/cli/graph.ts:658` — **Producer of the ratification still spells the shape inline**
  The diff replaces the consumer site (`let ratification: Ratification | undefined;`) but `selectRatification` — the only thing that builds one — still returns `{ kind: "reaction"; id: string; author: string } | undefined` at line 658, so the next shape change is still a multi-site edit, which is the rationale the new doc comment gives for naming the type.
  Fix: Change `selectRatification`'s return type to `Ratification | undefined`.

### Maintainability

- **[suggestion]** `test/cli-graph.test.ts:576` — **Assertion couples test to joined reason prose**
  `expect(receipt.attestationFacts?.reasons?.join(" ")).toContain("no ratification found")` flattens a structured array into one string and substring-matches it, so any rewording of the reason line breaks the test while telling the reader nothing about which conjunct failed. The three preceding assertions (`ratification).toBeUndefined()`, `evidence.some(...approved)).toBe(false)`, `attestation).toBe("unverified")`) already pin the same outcome structurally.
  Fix: Assert membership on the array (`expect(receipt.attestationFacts?.reasons).toContain(<exact reason constant>)`) or drop this line as redundant.

### HonestOracle

- **[suggestion]** `docs/work-graph.md:362` — **New admissibility absolutes have no evidence in this diff**
  The added bullet states as fact "matched on `+1`, so no other emoji ratifies, and (per the rule above) any non-proposer's counts" — but the only evidence this PR supplies seeds exactly one reaction, `{ id: "r1", content: "+1", author: "jcfischer" }`, under the same identity as the seeded root author. Neither the other-emoji exclusion nor the non-proposer clause is exercised anywhere in the diff.

- **[suggestion]** `test/cli-graph.test.ts:537` — **Test comment keeps the "by construction" framing the PR retracted**
  The preamble says the amendment rule is "an invariant that holds by construction", while docs/work-graph.md in the same commit says "passing the superseded id ratifies from the superseded proposal, and nothing in the runtime rejects it". A future reader hitting the test first will take the invariant as enforced rather than as caller discipline.


---

jcfischer (commented):

9 finding(s): 7 suggestion, 2 nit.


### CodeQuality

- **[suggestion]** `docs/work-graph.md:362` — **New sentence names an admissibility condition, then denies one exists**
  "any non-proposer's counts, with *whose* it is affecting `attestation` rather than admissibility" states the non-proposer restriction (an admissibility rule keyed on whose 👍 it is) and in the same clause says whose-it-is doesn't affect admissibility. Read literally, a proposer's own 👍 ratifies.
  Fix: Scope the second clause: "...any non-proposer's counts; among those, *whose* it is affects `attestation`, not admissibility."

- **[suggestion]** `test/cli-graph.test.ts:574` — **"approved" evidence assertion has no positive counterpart**
  `expect(receipt.evidence.some((entry) => entry.kind === "approved")).toBe(false)` is a negative assertion over a bare string literal; a typo or a renamed evidence kind makes it pass forever, and Arm A never asserts the kind is present on the ratified close.
  Fix: Add the Arm A mirror: `expect(ratified.closed[0].receipt.evidence.some((e) => e.kind === "approved")).toBe(true)`.

- **[suggestion]** `test/cli-graph.test.ts:559` — **Arm A pins the ratification by id only, never kind or author**
  `expect(ratified.closed[0].receipt.attestationFacts?.ratification?.id).toBe("r1")` leaves the two fields that carry meaning unchecked — `kind`, the member this PR narrows, and `author`, which drives the attestation conjuncts.
  Fix: Assert the whole record: `.toEqual({ kind: "reaction", id: "r1", author: "jcfischer" })`.

- **[nit]** `src/work-graph.ts:241` — **Comment documents `kind`; `id` is the ambiguous field**
  The comment argues a bare shape "would leave a future reader to infer the mechanism from an id shape", then ships `id: string` undocumented — a reader still can't tell it holds the reaction id (`r1`) rather than the comment id, which lives separately on `proposal.commentId`.
  Fix: `/** The reaction id — not the proposal comment id. */` on `id`.

### Maintainability

- **[suggestion]** `src/work-graph.ts:224` — **Doc comment restates prose that lives in two other files**
  The 11-line block comment ("`kind` is deliberately a one-variant discriminant … dropping the withdrawn `"comment"` member (#525) took a three-file edit") repeats what `docs/work-graph.md` and `closing.md` already say in this same diff, and the field-level comment then points back at the doc anyway ("the HITL receipts section of `docs/work-graph.md` has why"). Four copies of one rationale means four places to update when it changes.
  Fix: Keep the one-line field comment pointing at `docs/work-graph.md`; drop the block comment's re-derivation down to a sentence.

- **[suggestion]** `test/cli-graph.test.ts:534` — **13-line test preamble outweighs the 20 lines it explains**
  The comment block from "§3.2's amendment rule" through "neither carries it alone" restates the PR description's two-arm argument inside the test body, on top of the same argument already appearing in `docs/work-graph.md` and `closing.md`.
  Fix: Trim to the two sentences that a future reader needs: why both arms exist and why deleting either weakens the test.

- **[nit]** `test/cli-graph.test.ts:553` — **Both arms repeat propose-then-seed-👍 verbatim**
  Arm A and Arm B each run `run(["graph", "close", "530", "--propose", "--body", "the resolution", "--repo", REPO], …)` followed by the identical `reactions.set("c1", [{ id: "r1", content: "+1", author: "jcfischer" }])`. The `seeded()` factory already extracts the store setup; the propose+react pair is the next obvious extraction.
  Fix: Extend `seeded()` into an async `proposedAndRatified()` helper that returns the store with c1 proposed and the 👍 seeded.

### HonestOracle

- **[suggestion]** `docs/work-graph.md:370` — **Withdrawal rests on an unverifiable auditor property**
  The spec now permanently justifies dropping the comment path with "produce `verified` on the one field the phase-2 auditor recomputes" — an exclusivity claim about the phase-2 close auditor with no citation, no code in this diff, and no way for a reader to check it. If the auditor recomputes more than that one field, the stated reason for the withdrawal is wrong as written.

- **[suggestion]** `src/work-graph.ts:245` — **`fix(...)` commit that fixes nothing, closing a feature request**
  The title is `fix(work-graph): ...` and the body says "Resolves #525", while the same body states "No behaviour change beyond the type narrowing" — the diff is docs, a type rename, and a test. A changelog or bisect reader sees a work-graph bug fix and a resolved feature issue where neither exists.


---

Issue #564, opened 2026-08-07
The frontier walks the subtree: transitive discovery over nested subIssues


jcfischer:
Implements the decision on #557: `soma graph frontier <root>` reports every
open, unassigned, unblocked node in the root's **membership subtree**, at any
depth. Depth records provenance; it no longer decides visibility.

**This is urgent in a way the decision node was not.** Executing #557's cleanup
— re-parenting #527 → #526 and #529/#530 → #499, so the tracker tells the truth
about provenance — left `soma graph frontier 495` returning **`none`** while six
open nodes sit in the subtree (#527, #529, #530, #556, #560, and #557 itself,
claimed). The map is correct and unwalkable at the same time. Nothing is
unsafe — §2.4 puts correctness on the claim and close gates, never on frontier
completeness — but an empty frontier reads exactly like a finished map, which is
the §2.4 false negative at full strength. Until this lands, the map's Notes carry
the warning.

## What to build

1. **`GraphStore.listCandidateFrontier(root)` changes meaning** — from "open
   direct children" to "every open node in the root's membership subtree". A §2.5
   seam change, same class as the `CreateNodeSpec.body`/`parent` addendum on #497,
   and the docblock at `src/work-graph.ts:312` needs to say so. ~~A backend with no
   nesting concept returns its flat set and is trivially conformant.~~
   **Corrected while implementing** (sage on PR #571): that was false, and it
   told an adapter author the opposite of the contract — a flat set carries
   neither subtree membership nor an order. What is true is narrower: a backend
   whose membership relation is only ever one level deep implements this
   honestly, because its subtree of `root` *is* `root`'s members and pre-order
   over one level is their order. A store that cannot express membership at all
   cannot implement the seam; a flat dump is a wrong answer, not a degenerate
   one.

2. **Descend into closed nodes; never report them.** The pre-filter at
   `src/work-graph-github.ts:333` conflates pruning a closed *node* with pruning
   its *subtree*. The doubly-invisible case is the normal one: scaffold outlives
   the node that spawned it by construction, so it goes dark at the exact moment
   the parent closes and it becomes takeable. #556 (below closed #501) and #560
   (below closed #510) are the live instances.

3. **Discover over GraphQL, not recursive REST.** Nested `subIssues` returns
   multiple levels in one round trip — probed on #557 against this map, two
   levels for the price of today's single `sub_issues` call. Recursive REST would
   scale with the map's *closed* history, which only grows, making a map slower
   as it succeeds.

[...]

Issue #560, opened 2026-08-07
Enrollment state is asserted in a generated file: confidentiality-gate.yml now lies


jcfischer:
## Problem

`.github/workflows/confidentiality-gate.yml` in this repo is **generated** (`compass/tools/rollout-gate.ts`, a different repo — `~/work/mf/compass`). Its header asserts:

> WARN-ONLY BURN-IN: this standalone workflow is **NOT enrolled in any required status check**, so a finding here never blocks a merge. … A principal flips enforcement later by adding this check context to the repo ruleset (`rollout-gate.ts --rulesets`) — deliberately NOT done here.

As of #510 that assertion is **false**: `gate / confidentiality-gate` is now a required status check on the `protect-main` ruleset. Two records disagree, and the wrong one is the copy a reader finds first.

## Why hand-editing the header is not the fix

The file is regenerated by `rollout-gate.ts`, so a hand edit is clobbered on the next bump. The generator is in another repo, which makes soma's tree the downstream copy.

Note also: `planRulesets()` only **adds contexts to an existing `required_status_checks` rule** and prints `NO ruleset / no required_status_checks rule — skipped (needs L5 ops)` otherwise. soma's ruleset had no such rule, so the tool would have skipped this repo — creating the rule by hand was the L5-ops step the tool defers, not a divergence from it. But the generated header does not model "enrolled" as a state a repo can be in.

## The decision

How does enrollment state stop being a lie in the generated header? Candidates: teach `rollout-gate.ts` to read the repo's ruleset and emit the accurate header; drop the enrollment claim from the generated text entirely (it is deployment state, not source state — arguably it never belonged in a source file); or accept the staleness and record it.

Scaffold of #510. Decision node — the fix lands in compass, not soma.

jcfischer:
## Resolution

**The generated header stops asserting enrollment at all.** It states only what the *file* controls and points at where enrollment actually lives, rather than claiming a value for it.

### The node found half of it

#560 caught the enrollment clause. Verifying it surfaced a second false clause in the same sentence, and the second one is the reason this was worth doing:

> WARN-ONLY BURN-IN: this standalone workflow is NOT enrolled in any required status check, **so a finding here never blocks a merge.**

**Enrolled** — `gate / confidentiality-gate` is a required status check on the `protect-main` ruleset (id `17446144`), added as part of #510. Read live.

[...]

Issue #556, opened 2026-08-06
The resolution comment has no verb on the AFK close path


jcfischer:
## The question

Doctrine requires a **resolution comment** on every node before it closes (`closing.md`, "Recording the resolution", step 1) and forbids reaching past the `soma graph` verbs to the tracker's own CLI (`SKILL.md`, "The verbs"). On the **AFK path those two rules contradict each other**: `--body`/`--body-file` feed `--propose` only, `postComment` is on `GraphStore` but no verb surfaces it, and so closing an `auto` node the doctrine's own way requires `gh issue comment`. Closing #501 did exactly that.

Not a cosmetic gap. The receipt is the machine-readable half of a close and the resolution comment is the human-readable half — the half a later reader actually reads. Leaving its posting outside the verbs means the one artefact humans consume is the one artefact with no contract, no repo resolution, and no backend abstraction: it is GitHub-specific by accident, which is exactly what the seam exists to prevent (and #533's GitLab map inherits the problem).

## What has to be decided

- Does the resolution comment become a verb, an option on `close` (`--resolution-file`, posted before the probes run), or is prose-to-tracker declared legitimately outside the verbs — and if the last, does `SKILL.md`'s "never reach past them" need the exception written down?
- If it rides `close`: posted before or after the probes? Before means a refused close leaves an orphan resolution comment on an open node (the `--propose` path already accepts that trade). After means a passing close that fails to comment has a receipt with no prose.
- Is it **required** or optional? A close that refuses for a missing resolution comment is a new hollow-close conjunct; one that merely warns leaves the doctrine unenforced, which is the state today.
- Does the AFK path want the same two-phase shape as HITL, or is the asymmetry correct — nobody is being asked for anything on an `auto` close.

Decision only; the implementation is a follow-up if a verb wins.

jcfischer:
## Resolution

**Every close carries prose, and `soma graph close` is the only thing that posts it.** The doctrine and the verbs stop contradicting each other by giving the verb the missing capability, not by writing an exception into the doctrine.

### The rule

| Path | What satisfies it |
| --- | --- |
| `auto` close | `--resolution-file <path>` |
| HITL **bare** close | `--resolution-file <path>` |
| HITL `--propose` → `close --proposal-comment <id>` | the proposal body, already posted |

One exemption, with a named reason: the proposal body **is** the resolution, so requiring a second prose comment on the ratification close would post the same thing twice.

[...]

Issue #530, opened 2026-08-04
What is the soma graph read path allowed to cost?


jcfischer:
## Question

The read path of `soma graph` is strictly serial over `gh` subprocesses, and the
verb orienteer runs *first, every session* is the worst of them. Sage's review of
#528 measured the shape; nothing here is a correctness bug, and none of it was
designed — it accumulated across #497 and #498.

Four findings, one question: **what is the read path allowed to cost, and which
of these do we fix?**

1. **`frontier` confirms candidates strictly serially** (`src/work-graph.ts`).
   `for (const candidate of candidates) { await this.store.readNode(candidate) }`
   — each `readNode` is three `Bun.spawn(["gh", …])` round-trips, so a 30-child
   map costs ~90 serial CLI invocations before a session can pick anything up.
   The candidates are independent reads with no shared state. Direct-fetch
   confirmation is required by §2.4 (it removes the false positives a lagging
   search index produces), so the fix is concurrency, not skipping it.

2. **`readNode` always pays a parent lookup that most callers discard**
   (`src/work-graph-github.ts`). `issue.parentNumber ?? await
   this.fetchParentNumber(...)` always fires, because REST carries no `parent`
   key at all — that is the #498 finding, and the GraphQL call was the fix. But
   `frontier` and `claim` throw `parent` away, so that spawn is pure waste on the
   hottest path, and the independent `blocked_by` fetch is serialised behind it.

3. **`close` re-reads the node it just read** (`src/cli/graph.ts`). `runClose`
   opens with `graph.readNode(ref)`, then `findGraphRoot(ref, …)` starts its walk
   at the same `ref` — three redundant spawns per close.

4. **Each `--blocked-by` edge re-walks the graph** (`src/cli/graph.ts`,
   `addBlockingEdge`). Every edge runs the cycle check's BFS with no cache
   carried across edges, so `soma graph add` with five blockers re-reads
   overlapping ancestors five times.

**To decide:**

- Is bounded concurrency acceptable on the confirmation loop, and what bound?
  (Ordering of `confirmed` must survive — the frontier is presented in order.)
- Should `readNode` take an opt-in flag for the parent walk, or should the root
  walk use a different call? A flag on a seam method is the kind of thing §2.5
  deliberately kept narrow.
- Is a per-invocation `NodeState` cache the right shape, or does caching inside a
  verb invite the staleness `frontier`'s direct-fetch confirmation exists to
  avoid?
[...]

PR #571, opened 2026-08-08
feat(work-graph): the frontier walks the subtree (#564)

### LINKED ISSUE #501: planSteps nodeId bridge (spec §2.7)
## Task

Implement the bridge contract from `docs/work-graph.md` §2.7 + `docs/algorithm-execution-modes.md`: optional `nodeId` on `AlgorithmPlanStep`; `updateAlgorithmPlanStep` refuses a direct status write on a bridged step; status re-derives via `GraphStore.readNode` / `soma graph node <id>`. One work item never has two authoritative homes.

Node: `kind: task` · `autonomy: auto` (branch work; merge rides normal PR review) · probes (close evidence): `bun test` exit 0 including a refusal test for direct status writes on bridged steps, `bunx tsc --noEmit` exit 0.

### LINKED ISSUE #510: Ivy-authored code: branch protection on main, then Write for ivy-agent

### LINKED ISSUE #557: Scaffold nodes are invisible to the frontier: the walk is one level deep
## The defect

Orienteer doctrine (#492, and this map's Notes) says **scaffold nodes attach below their spawning node, never to the map**, because hanging them off the map "makes the frontier report work the effort does not gate on". But `listCandidateFrontier` fetches only the root's **direct** `sub_issues` (`src/work-graph-github.ts:328`) and drops closed children as a pre-filter — so a correctly-placed scaffold node is **invisible to `soma graph frontier`**, and doubly invisible once its spawning node closes.

Demonstrated live this session: #556 was attached below #501 exactly as doctrine requires, and `soma graph frontier 495` does not list it. Meanwhile #527, #529 and #530 — all scaffold thrown off other nodes — were attached **directly to the map** and are visible. The doctrine and the practice already disagree, and the practice is what keeps the frontier honest.

This is the §2.4 false-negative hazard the #513 decision named as the one direction direct-fetch confirmation cannot recover: a short frontier is indistinguishable from a complete one. #513 closed that hazard for the map's own children and left this instance of it open.

## What has to be decided

- **Does the frontier walk transitively?** A recursive sub-issue walk makes doctrine and implementation agree, at the cost of N+1 fetches per level — and #530 ("What is the soma graph read path allowed to cost?") is the open node that owns that budget, so this decision is coupled to it.
- **Or does the doctrine change** — scaffold attaches to the map after all, and the "work the effort does not gate on" objection is answered some other way (a `kind`, a label, a rendering distinction) rather than by burying it?
- **Closed parents.** If the walk goes transitive, the closed-child pre-filter has to stop pruning *subtrees* while still pruning closed *nodes* — an open child of a closed parent is precisely the scaffold case, and today it is dropped twice.
- **What happens to #556** and to the three scaffold nodes already mis-attached to the map: re-parent to match whichever rule wins, or leave as historical placement.

Blocked by nothing, but resolve after or alongside #530 — the read-path budget decides whether the transitive walk is affordable.

jcfischer:
Implements #564, which resolves #557. Map: #495.

## The defect

`listCandidateFrontier` returned the root's **direct** open children. Orienteer doctrine puts scaffold below its spawning node, so a correctly-placed scaffold node was invisible to `soma graph frontier` — and dropped twice once the parent closed, since closed children were pruned as *subtrees*, not just as nodes.

**Scaffold outlives the node that spawned it by construction**, so it went dark at the exact moment the parent closed and it became takeable. Whether that is an edge case or the normal one is a testable claim, so here is the test — every open node in map #495's subtree, and its parent:

```bash
gh api graphql -f query='{repository(owner:"the-metafactory",name:"soma"){issue(number:495){
  subIssues(first:50){nodes{number state subIssues(first:20){nodes{number state}}}}}}}' \
  -q '.data.repository.issue.subIssues.nodes[]
      | . as $p | .subIssues.nodes[]? | select(.state=="OPEN")
      | "\(.number) below \($p.number) (\($p.state))"'
```

Today it prints six rows, and **every parent is closed**:

```
529 below 499 (CLOSED)
530 below 499 (CLOSED)
556 below 501 (CLOSED)
560 below 510 (CLOSED)
527 below 526 (CLOSED)
564 below 557 (CLOSED)
```

Two of those (#556, #560) were hidden when #557 was charted; the rest arrived from resolving it — re-parenting three scaffold nodes to where doctrine says they belong, plus this node. Which is the point rather than a confound: **the population of below-a-closed-parent nodes is every one of them, and it only grows**, because a map closes its steps as it walks them. Six for six is what "not an edge case" means here. §2.4 names this direction as the one direct-fetch confirmation cannot recover: a short frontier is indistinguishable from a complete one.

Resolving #557 made it maximal: re-parenting three scaffold nodes to where doctrine says they belong left `soma graph frontier 495` returning `none` while six open nodes sat in the subtree.

## What changed

**Seam contract** — `listCandidateFrontier` returns every open node in the root's membership subtree, any depth, depth-first pre-order. Descends into closed nodes; never reports them. Pre-order puts each node's scaffold immediately after it, so provenance reads off the listing.

**Discovery over nested GraphQL rather than recursive REST.** REST costs one request per visited node, and the walk must descend into *closed* nodes — so it would scale with the map's closed history, which only grows, making a map slower as it succeeded. The nested query carries three levels per round trip.

**Truncation is detected and completed, never silent.** `totalCount` at every level, plus a `totalCount`-only probe on the bottom row so a leaf and an unfetched subtree are distinguishable. Two modes — a level that hit its page size, and the bottom row — and both repair the same way: **re-root the query at the incomplete node**, where its children become the *top* connection and are cursor-paged to completion. Re-fetching in place returns the same truncated set, so re-rooting is the mechanism rather than an optimisation; every re-root strictly descends, so it terminates. `hasNextPage` with no cursor throws rather than reading short.

**Doctrine follows the code.** `references/fog.md` and the SKILL.md invariant both said burying scaffold keeps the frontier clear of work the effort does not gate on. That is now false, and it was the wrong carrier to begin with — gating is what a blocking edge means. Both now say provenance, and name the two real ways to keep work off the frontier: close it as out of scope, or block it.

## Verification

- `bun test` — 2300 pass, 0 fail. `bunx tsc --noEmit` clean. (Both are the node's declared probes, registered for this checkout under #526.)
- **Tests checked against three hand-applied mutations** — pruning closed subtrees fails 3; ignoring truncation fails 2; dropping the `seen` guard does not terminate. There is no committed harness, so this is a reproduction recipe rather than a standing gate; it is recorded as such at the tests it justifies (sage, cycle 1).
- **Live against map #495** — reports exactly the hand-checked set, in pre-order, with #564 withheld as claimed:

```
- 529 Ungated argv probes …        (below #499)
- 530 What is the read path …      (below #499)
- 556 The resolution comment …     (below closed #501)
- 560 Enrollment state …           (below closed #510)
- 527 Probe runner envelope …      (below #526)
```

## Not in scope

The serial `readNode` confirmation loop — 3 `gh` spawns per candidate, and the reason the live run above takes ~16s. Transitivity grew this map's candidate set from 4 to 6, so it costs +6 serial spawns: bounded by *open* work rather than closed history, and squarely inside #530.


jcfischer (commented):

### Performance

- **[important]** `src/work-graph-github.ts:428` — **Parallelize sibling completion fetches**
  `for (const child of await this.completeChildren(node)) await visit(child);` waits for each re-rooted GraphQL fetch before starting the next. A subtree with many bottom-row non-leaves incurs one sequential network round trip per sibling, producing avoidable cumulative latency.
  Fix: Complete sibling subtrees with `Promise.all`, then emit their results in original order.

### Maintainability

- **[suggestion]** `src/work-graph-github.ts:248` — **Centralize status normalization**
  Both `readSubtreeNode` and `fetchSubtree` repeat `record.state === "OPEN" ? "open" : "closed"`, including its non-OPEN policy. Keeping the conversion in two paths invites drift when GraphQL state handling changes.
  Fix: Extract a `readNodeStatus(record)` helper and use it in both locations.

### HonestOracle

- **[important]** `src/work-graph-github.ts:186` — **Query-cost headroom is overstated**
  “a future level can be added without recomputing it” is false for the established 20-wide shape: one added 20-wide level contributes at least 600,000 nested entries, exceeding the stated 500,000 ceiling. This can lead maintainers to ship a GitHub-rejected query.

- **[important]** `commit message:31` — **Mutation verification is unsubstantiated**
  “confirmed falsifiable by mutation: pruning closed subtrees fails 3, ignoring truncation fails 2” has no committed mutation harness, mutant cases, or reproducible command; the diff contains only ordinary fake-transport tests. This presents untraceable regression evidence as a verified merge signal.


---

jcfischer:
## Cycle 1 — three fixed, one deferred with reasons

`7bd24c6`.

**Query-cost headroom overstated** — fixed, and sage was *understating* it. The comment claimed a level could be added without recomputing. Each level multiplies everything below it, so a fourth 20-wide level scores **1 231 550** against the 500 000 ceiling, not 600 000; the widest fourth level that fits is **7**. The comment now says recompute before adding a level and names depth as the expensive axis — which is the real argument for staying shallow and letting re-rooting handle the tail. This is the class of finding this map keeps hitting: the code was fine, the claim about it was false.

**Status normalisation duplicated** — fixed. `readGraphQLStatus`, read at both levels. Agreed on the reasoning: a drift between the two sites silently changes which nodes get reported, which is the failure mode with no symptom.

**Mutation evidence unsubstantiated** — fixed, and the criticism is right about what mattered. The mutations were genuinely run and the numbers are real, but nothing in the repo let a reader re-run them, so it read as a verified merge signal backed by nothing checkable. The three mutations and their expected failures are now recorded at the tests they justify, explicitly as a reproduction recipe rather than a standing gate, and the PR body no longer says "confirmed falsifiable". Worth noting mutation 3 is why the cycle test asserts at all — a `seen` set over a shape GitHub guarantees to be a tree reads like dead code until you remove it.

**Parallelise sibling completion fetches** — deferred to #530, recorded there as finding 5 rather than dropped.

Two reasons, and the first is scoping rather than disagreement. Read-path cost is the open question #530 owns — findings 1–4 of that node are the same shape (serial confirmation loop, wasted parent lookups, redundant re-reads), and this belongs beside them, under whatever measurement target that node settles on. Fixing it here would decide #530's question by increment.

The second is that the cost is **conditional** in a way the other findings are not: `completeChildren` only re-roots where a subtree outran the query's depth or width, which on map #495 is never. So this is latency a real graph may never pay.

And the fix is not free. The walk emits in depth-first pre-order and dedupes through a shared `seen` set. Fetching siblings concurrently means hoisting the fetches out of `visit` and replaying results in order — otherwise the reported frontier order stops being deterministic, and order is part of what makes provenance readable in the listing. Worth doing against a measurement target; not worth doing blind.

Full suite green (2300/0), `tsc` clean.

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/work-graph.ts:334` — **Flat backend is not trivially conformant**
  The contract says results are in "depth-first pre-order" so that "provenance [is] legible," then claims "A backend with no nesting concept returns its flat set and is trivially conformant." A flat set cannot establish subtree membership or depth-first order, so this misstates conformance to downstream adapter authors.


---

jcfischer (commented):

2 finding(s): 2 important.


### CodeQuality

- **[important]** `src/work-graph-github.ts:496` — **Reject invalid pagination state**
  `if (pageInfo.hasNextPage !== true) break;` treats a missing or malformed `hasNextPage` as the final page, returning an incomplete candidate list despite the no-silent-truncation contract.
  Fix: Break only on `hasNextPage === false`; throw a backend error for any other value.

### HonestOracle

- **[important]** `src/work-graph-github.ts:426` — **Top-level count is not checked**
  The comment claims "`totalCount` at every level" means a subtree is "never silently dropped," but `fetchSubtree` reads only `nodes` and `pageInfo`. A top-level response with `totalCount: 2`, one node, and `hasNextPage: false` returns that one node without error, so the claimed fail-closed guarantee is false.


---

jcfischer (commented):

2 finding(s): 2 important.


### HonestOracle

- **[important]** `PR description:7` — **Live defect prevalence is unproven**
  It asserts "That is not an edge case" and "Two live instances on this map" without a command, captured issue state, or reproducible query establishing those relationships. This asks reviewers to accept the defect’s prevalence as fact rather than as a testable premise.

- **[important]** `docs/work-graph.md:237` — **Search-index caveat no longer matches discovery**
  The text says "The frontier verb confirms every search hit" and attributes false negatives to "a lagging index," while this diff replaces candidate discovery with a GraphQL `subIssues` subtree query. No evidence shows a search index remains in this path, so the #492 caveat misstates the current failure mode and conflicts with the new whole-subtree contract.


---

jcfischer:
## Cycle 4 — both fixed, and the first one caught a live staleness

`2342041`, plus a PR-body correction.

**Search-index caveat no longer matches discovery** — fixed, and this was the better of the two. §2.4 said the verb "confirms every search hit" and blamed false negatives on a lagging index. This backend has never searched; it walks native sub-issue edges. So the caveat described a discovery path that is not in the code, and after this PR it describes one that never was.

The confirmation rule is backend-agnostic and stays, restated over *candidates* rather than search hits. What can produce a false *negative* is backend-specific, and the distinction earns its words because the fixes differ: a search-based store loses what its index has not caught up with; this one loses nodes **nobody attached**, since a node in no subtree is in nobody's frontier. And a short read caused by the walk rather than by the graph is a defect, not a caveat — which is what the truncation handling is for.

**Live defect prevalence unproven** — fixed by publishing the query, and running it immediately proved the point in a way I did not expect. The body now carries the exact `gh api graphql` command and its output.

The output is **six rows, not the two I had asserted** — because resolving #557 re-parented three scaffold nodes to where doctrine says they belong, and created #564. My figure was accurate when written and stale by the time it was read, which is precisely the failure mode this lens exists to catch. Worth stating plainly: the finding was right for a reason better than the one given.

It also strengthens the claim rather than weakening it. Every open node one level down sits under a closed parent — six for six — because a map closes its steps as it walks them, so the below-a-closed-parent population only grows. That is what "not an edge case" was reaching for, now shown instead of asserted.

Full suite green (2302/0), `tsc` clean.

jcfischer (commented):

1 finding(s): 1 important.


### CodeQuality

- **[important]** `src/work-graph-github.ts:517` — **Reject repeated pagination cursors**
  `after = pageInfo.endCursor;` accepts the same cursor repeatedly. A response that keeps `hasNextPage: true` with a repeated cursor causes an unbounded loop rather than a backend error.
  Fix: Track seen cursors and throw when `endCursor` repeats.


---

jcfischer (commented):

2 finding(s): 2 important.


### CodeQuality

- **[important]** `docs/work-graph.md:237` — **Do not claim membership revalidation**
  The text says direct fetch removes false positives from "a membership edge that no longer means what it did," but `WorkGraph.frontier` only rechecks candidate status, assignees, and blockers; it never verifies the candidate remains in the requested root subtree.
  Fix: Limit the claim to candidate-state false positives, or add explicit ancestry validation.

### HonestOracle

- **[important]** `src/skills/orienteer/SKILL.md:86` — **Claims only two frontier exclusions**
  The text says "Work you don't want taken is closed as out of scope, or blocked; there is no third way." But the documented frontier predicate also requires a node to be unassigned, so claiming open work removes it from the frontier without closing or blocking it; the asserted doctrine is false.


---

jcfischer (commented):

1 finding(s): 1 important.


### ContextDrift

- **[important]** `src/work-graph.ts:312` — **Define membership in the glossary**
  The public seam now returns the root's "membership subtree", while CONTEXT.md:685-700 defines work-graph topology as nodes joined by "blocking edges" and says it records "what blocks what". This adds a foundational relation without defining how membership differs from blocking topology.
  Fix: Update CONTEXT.md's work-graph section to define membership edges/subtrees and their frontier-only role.


---

jcfischer (commented):

1 finding(s): 1 blocker.


### HonestOracle

- **[blocker]** `CONTEXT.md:693` — **Claims an unbuilt auditor is active**
  The text says raw tracker writes are handled because “the phase-2 auditor detects and reverts them after the fact,” while the same diff states “the §5 phase-2 tick and auditor are not” implemented. Downstream users may rely on enforcement that does not exist.


---

jcfischer (commented):

2 finding(s): 2 important.


### HonestOracle

- **[important]** `CONTEXT.md:687` — **Edge contract contradicts the spec**
  The new claim that nodes are "joined by two distinct edge kinds" conflicts with `docs/work-graph.md:217-218`, which still says "Blocking only" and "No other edge types in v1." This leaves consumers unable to tell whether membership is a supported graph relation or forbidden ceremony.

- **[important]** `CONTEXT.md:696` — **Auditor status remains contradictory**
  The new text says the phase-2 auditor is "not built" and raw writes are "never detected either," but `src/work-graph.ts:831` still claims the bypass is "detected by the phase-2 auditor." This misleads readers about whether a hollow tracker close is presently observable.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `PR description:1` — **Implementation claim omits #564 requirement**
  “Implements #564” is not supported: #564 says “A backend with no nesting concept returns its flat set and is trivially conformant,” while this diff says a store without membership “cannot implement this seam” in `src/work-graph.ts:337`. Readers will treat the issue as satisfied despite the PR rejecting one of its stated contract requirements.


---

jcfischer (commented):

No findings. Sage approves.



---

PR #462, opened 2026-07-13
the-algorithm: ship the executable reference closure (was: BPE safe cuts)

jcfischer:
## What

BitterPillEngineering audit of soma's Algorithm doctrine (`src/skills/the-algorithm/`).

**The PR reversed direction mid-review.** It began by *cutting* references to files believed missing; it now *commits those files*, because they were never missing — only unshipped. The original framing is preserved below for history, with what actually happened marked.

## Current state of the diff

**Ships the executable reference closure.** Eight skill-relative load targets, all resolving in source: `algorithm.md`, `capabilities.md`, `eval-guide.md`, `ideate-loop.md`, `mode-detection.md`, `optimize-loop.md`, `parameter-schema.md`, `target-types.md`.

**Normalizes paths.** The tree carried three spellings of the same file — `~/.claude/PAI/ALGORITHM/x.md`, `PAI/ALGORITHM/x.md`, and bare `x.md` — now all `references/x.md`, so the closure is machine-checkable rather than prose-shaped.

**Fixes an unrunnable mandate.** The doctrine required the ISA Skill at `~/.claude/skills/ISA/` at E2+. That path does not exist; soma bundles it as `VSA` (this repo's own #329 rename, commit `c5849d9`, in this branch's history). Codex and Pi projections could not have executed a protocol the doctrine made mandatory. The 20 `Skill("ISA", …)` invocations are now `Skill("VSA", …)`, referenced by name rather than path. The ISA *concept* vocabulary (71 occurrences) is untouched — that belongs to #329's migration.

**Two honesty fixes.**
- A phantom capability name was described as *"rejected by the audit gate"*, which reads as machine enforcement. Nothing enforces it; the gate is a self-check the executor runs, and it now says so.
- `capabilities.md` claimed ReReadCheck *"targets the 82% 'missed ask' complaint cluster"* — a precise figure with no source or method, backing a universal mandate. It now names the cluster without inventing a number.

**Does not ship `changelog.md`.** Committed in an earlier round and reverted: 279 lines of history naming `v5.0.0`–`v5.5.0`, `vX.Y.Z`, `canonical-isa`, `PAI_SYSTEM_PROMPT`. History is not doctrine, and `algorithm.md` itself calls it *"read on demand"*.

## Correcting the original claim

[...]

PR #577, opened 2026-08-08
fix(skills): de-PAI the Algorithm's shipped references

### LINKED ISSUE #574: the-algorithm's shipped references are PAI-shaped: 2169 lines carrying PAI paths, unsubstituted placeholders, and named external models as mandatory doctrine
#462 shipped the Algorithm's reference closure into the bundle, which closed the "a fresh install reads files that don't exist" gap. It was a lift-and-shift, not a de-PAI pass: the files went in essentially as authored under PAI, and an adopter now reads instructions that name PAI hooks, PAI tool paths, and specific external models as mandatory.

This is not the ISA/ISC vocabulary question. ADR 0001 (line 83) retires `ISA`/`ISC` as canonical Soma terms while **retaining them as an optional Algorithm pack** until the code/path rename lands, so the ~157 ISA/ISC mentions in these files are a deferred rename, not drift. What follows is everything *not* covered by that.

## Measurements

All counts from `src/skills/the-algorithm/references/` at `a0c3527`.

| File | lines | `~/.claude` | `PAI` | `{{TPL}}` | Forge/Anvil/Cato/Advisor |
|---|--:|--:|--:|--:|--:|
| `algorithm.md` | 674 | 7 | 11 | 6 | 19 |
| `capabilities.md` | 137 | 3 | 4 | 8 | 34 |
| `eval-guide.md` | 188 | 0 | 1 | 0 | 0 |
| `ideate-loop.md` | 143 | 0 | 0 | 0 | 0 |
| `mode-detection.md` | 93 | 0 | 0 | 2 | 0 |
| `optimize-loop.md` | 480 | 0 | 1 | 0 | 0 |
| `parameter-schema.md` | 247 | 0 | 0 | 0 | 0 |
| `target-types.md` | 207 | 1 | 1 | 0 | 0 |
| **total** | **2169** | **11** | **18** | **16** | **53** |

Concentrated: `algorithm.md` and `capabilities.md` carry 96% of it. Four files (`ideate-loop`, `parameter-schema`, `optimize-loop`, `eval-guide`) are already close to clean.

## What an adopter actually reads

**Unsubstituted template placeholders.** 16 `{{DA_NAME}}` / `{{PRINCIPAL_NAME}}` occurrences ship as literal text. Nothing in the projection path substitutes them — `rewriteSubstrateProjectionContent` does memory-root rewriting and Claude-only line stripping, no templating. The mandatory response format in `algorithm.md` ends with a literal `🗣️ {{DA_NAME}}: [8-16 word summary]`.

**PAI tool and file paths.** `bun ~/.claude/PAI/TOOLS/Inference.ts --mode advisor …` (both in `algorithm.md` and as a live `command` capability row), `~/.claude/checkpoint-repos.txt`, `Bash('rg -l "KEYWORDS" ~/.claude/projects/${HARNESS_USER_DIR}/memory/feedback_*.md')`, and a `doctrine` learning route pointing at `PAI/ALGORITHM/v<next>.md`.

**PAI hooks that do not exist in Soma.** `ISASync.hook.ts`, `CheckpointPerISC.hook.ts`, `hooks/lib/isa-utils.ts`, and a `PromptClassifier` hook. `algorithm.md` states "Mode and tier are decided by a Sonnet classifier at UserPromptSubmit, not by the executor" — which contradicts `soma algorithm classify` (`src/algorithm-classifier.ts`), the shipped classifier.

**Named external models as mandatory doctrine.** 53 mentions of Forge / Anvil / Cato / Advisor. `Cato` is "MANDATORY at E4/E5 in VERIFY" and is invoked via `subagent_type: "Cato"` — a Claude Code sub-agent, in a skill whose own `SKILL.md` says to "Treat Claude-specific hooks, voice curls, and Claude Code sub-agents as source history, not portable requirements." The skill contradicts itself.

**A Voice Announcements section**, which is substrate-specific by construction.

## `capabilities.md` is runtime input, and two-thirds of it is inert

`loadSomaHomeAlgorithmCapabilityRegistry` (`src/algorithm-capabilities.ts:395-466`) parses this file's table into the capability registry a run selects from. Loading the shipped bundle as a soma home yields **11 definitions and 33 unsupported rows**.

Registered among the 11:

| Capability | Kind | Target |
|---|---|---|
| `Forge` | agent | `Agent(subagent_type="Forge")` |
| `Anvil` | agent | `Agent(subagent_type="Anvil")` |
| `Claude Code Guide` | agent | `Agent(subagent_type="claude-code-guide")` |
| `Advisor` | command | `bun ~/.claude/PAI/TOOLS/Inference.ts --mode advisor …` |
| `FeedbackMemoryConsult` | command | `Bash('rg -l … ~/.claude/projects/…')` |

The 33 unsupported — `Skill("RedTeam")`, `Skill("Research")`, `Skill("SystemsThinking")`, `Skill("Council")`, Agent Teams, Background Agents, Mass Parallelism, Session Branching, Worktree Isolation, `/code-review`, `/codex:rescue`, and the rest — resolve only against the author's principal skill set. **`unsupported` is returned by the loader and surfaced by no CLI verb**, so a fresh adopter gets a two-thirds-inert registry with no signal that anything was dropped.

## Projection does not mitigate this

`rewriteSubstrateProjectionContent` (`src/substrate-projection-rewrites.ts:41-53`) returns claude-code's copy **verbatim** (`if (input.substrate === "claude-code") return input.content`). For codex it rewrites memory roots and then strips Claude-only lines — and strips **zero** lines from `algorithm.md`.

Why zero is itself a bug: `CLAUDE_ONLY_LINE` (`:9`) matches `VSASync.hook.ts` / `VSA-Tool`, but the shipped doctrine says `ISASync.hook.ts` and `CheckpointPerISC.hook.ts`. The regex was migrated for #329; the file it filters was not. That is the third instance of one pattern — **code renamed, the file the code reads left behind** — after the `ISA Skill` capability row and the VSA skill's `canonical-isa.md` pointers (both fixed on `fix/vsa-canonical-pointers`).

## Suggested shape

Not one pass. Roughly, in dependency order:

1. **Decide the doctrine question first**: is `references/algorithm.md` still authoritative, or has the harness superseded it? Much of what it mandates in prose is now enforced in code — the OBSERVE current-state floor (#331), the LEARN gate (#330), the record-time VerificationGate (#439), typed `specified`→`probed` evidence. A doctrine file that restates enforced gates in PAI vocabulary is a second, weaker source of truth. This decision sizes everything below it.
2. **`capabilities.md`**: decide whether the shipped table should be portable (bundled-skill and `inline doctrine` targets only, adopters add their own rows) or stay the author's table. Either way, surface `unsupported` somewhere — a registry that silently drops two-thirds of its rows cannot be debugged.
3. **Mechanical residue**, once (1) is settled: substitute or remove the 16 placeholders, remove or gate the PAI tool paths, and either fix `CLAUDE_ONLY_LINE` to match what the files actually say or delete the mechanism as ineffective.
4. **Guard it**: the pointer-closure guard in `test/bundled-skill-references.test.ts` catches dangling references. A sibling assertion — no `~/.claude`, no `{{TPL}}`, no unresolvable capability target in a bundled skill — would keep this from re-accumulating.

Steps 1 and 2 are decisions; 3 and 4 are work that follows them.

jcfischer:
Closes #574.

First, a correction to that issue's framing: **#462 was not a lift-and-shift.** Diffed against the PAI-era originals it changed 52 lines in `algorithm.md` alone — `Skill("ISA")`→`Skill("VSA")` throughout, `PAI/ALGORITHM/*` pointers rewritten to `references/*`, the changelog pointer dropped, plus BPE cuts. It de-PAI'd the parts that were mechanical. This PR finishes the parts that needed judgment.

## `algorithm.md` — 674 lines, PAI residue cleared

"Cleared" means what the guard measures and nothing more: zero `~/.claude/` paths, zero `PAI/{TOOLS,MEMORY,ALGORITHM,DOCUMENTATION}` paths, zero `{{DA_NAME}}`/`{{PRINCIPAL_NAME}}` placeholders, no named external models. It does **not** mean the file speaks current Soma vocabulary throughout: it still says ISA and ISC in many places, which ADR 0001 line 83 explicitly retains as "an optional Algorithm pack" until the code/path rename lands. The terminology migration is a separate piece of work and this PR does not attempt it.

**The section that was wrong, not just PAI-flavoured.** Mode Classification said "mode and tier are decided by a Sonnet classifier at UserPromptSubmit, not by the executor," described `hooks/PromptProcessing.hook.ts`, and promised "no regex fallback, no model judgment." Soma's classifier is the opposite: `ALGORITHM_CLASSIFIER_CONTRACT` is pattern *data*, held as sources rather than RegExp literals precisely so adapters can project it and classify identically. The file contradicted the code it governs. Rewritten from `src/algorithm-classifier.ts` — the three modes, the 180-character native ceiling and why it is load-bearing, the tier rules, the empty-prompt E3 fail-safe, and the fact that failing to match an algorithm pattern does not route work to `native`.

**Mechanisms that do not exist.** `ISASync.hook.ts`, `CheckpointPerISC.hook.ts`, `hooks/lib/isa-utils.ts`, `~/.claude/checkpoint-repos.txt`. Replaced by the ones that do: `algorithm advance` as the deterministic phase gate, `algorithm verify` per criterion. The seven `**FIRST ACTION:** Voice …` lines now call the gate rather than a notify endpoint.

**Voice Announcements** → substrate-neutral phase announcements. `SKILL.md` already says to treat voice curls as source history; the section contradicted it. A substrate with a notification surface still uses it — the phase header is the portable floor.

**The reflection JSONL echo** → `algorithm reflect`, with the point spelled out: `gatesFired` is computed from run state by the same predicates the live gates enforce, and hand-writing it is exactly the hollow claim the computed/proposed split exists to prevent.

**Rules 2/2a/3** keep the doctrine and lose the binding. Second opinions at the three commitment boundaries; cross-family audit at E4/E5; surface conflicts rather than silently switching, capped at two re-asks. Gone: `bun ~/.claude/PAI/TOOLS/Inference.ts`, `subagent_type: "Cato"`, the Forge/Anvil/Cato auto-include table. Rule 2a is now explicit that it is the one rule that cannot be satisfied portably — where no second family is reachable, **record the gap in `## Decisions`; do not fabricate the audit, and do not let a same-family reviewer stand in for it, because that is the blind spot rather than the check.**

Also: PAI memory paths → `<soma-home>/memory/…`, `♻︎ Entering the PAI ALGORITHM…` → `Entering the Algorithm`, all 6 identity placeholders removed, and the knowledge check switched from `rg` over the tree to `memory recall` — recall is note-aware and records an event the retrieval-quality probe reads, whereas a grep leaves no trace, so a corpus that never answers anything looks identical to one that always does.

## `capabilities.md` — portable core, plus an overlay install cannot clobber

The constraint that decides the shape: `installBundledSkillsIntoHome` rewrites every bundled file byte-for-byte on each run **but leaves principal-added files under a skill dir alone**. So the bundled table cannot hold an adopter's rows, and simply making it portable would delete their capability set at the next `--apply`.

`loadSomaHomeAlgorithmCapabilityRegistry` now also reads a sibling **`capabilities.local.md`**, parsed **before** the bundled table, so a local row of the same name wins. An unresolvable local row lands in `unsupported` and *blocks* the bundled row of that name rather than falling back — an adopter who retargets a capability at a skill they do not have has disabled it, and should see that.

The shipped table is 8 rows. Five resolve anywhere — bundled-skill and `inline doctrine` targets. Three are `Contract(...)` rows (`Advisor`, `CrossFamilyCoder`, `CrossFamilyAudit`) that declare a contract rather than an invocation: they resolve, they are selectable, and they are **not invocable until bound locally** — `algorithm invoke` refuses them while unbound, and `capabilities --list` flags each one `← needs a local binding`. Against the bundle the registry goes from **11 definitions / 33 unsupported → 14 / 0**.

The file also documents what was undocumented and load-bearing: **every** table with a `Capability` header cell is parsed, not just the first (that is where 44 rows came from); skills auto-register under their own names, so a row exists to *narrow* a capability rather than to enable it; and which invoke shapes map to which kind, including that an unrecognised one is dropped silently.

## `substrate-projection-rewrites.ts`

`CLAUDE_ONLY_LINE` matched `VSASync.hook.ts` while the documents it filters said `ISASync.hook.ts` — the #329 rename migrated the regex, not the content. It stripped **zero** lines from `algorithm.md`. Now matches both spellings, and says why in a comment so the next rename does not silently re-break it. Third instance of that pattern this week, after the `ISA Skill` capability row and the `canonical-isa.md` pointers (#575).

## Guard

A third assertion in `test/bundled-skill-references.test.ts`: no bundled skill may carry a Claude-home path, a PAI tree path, or an identity placeholder. `migrate-pai-purpose` is exempt with a reason — naming `~/.claude/` is its subject matter, not residue — and the placeholder pattern is scoped to identity names so `{{SHA}}`/`{{VERSION}}` in example VSAs do not flag and train everyone to ignore the test.

Plus two tests for the overlay: a local row overriding a bundled one by name, and an unresolvable local row disabling rather than falling back.

## Verification

- `bun run typecheck` clean; no lint errors in any touched file.
- **212 tests pass, 0 fail** across `algorithm`, `algorithm-importer`, `algorithm-compat`, `install`, `home-projection`, `projection`, `skill-projection`, `pai-pack-normalizer`, `content-compare-doctor`, `bundled-skill-references`.
- Registry probe against the bundle: **14 definitions, 0 unsupported** — 12 resolved from the home plus the two compiled-ins, with the three contract rows flagged `← needs a local binding`.
- Registry probe against a live soma home carrying a `capabilities.local.md`: 122 definitions, with `Forge`, `Anvil`, `Advisor`, `RedTeam`, `IterativeDepth` and `VSA` all still registering — the overlay preserves an existing capability set intact.

## Note for adopters with an existing table

Anyone whose capability rows lived in the bundled `capabilities.md` should copy them to `capabilities.local.md` in the same directory before the next `soma install --apply`. Nothing generates that file; absent it, only the shipped table applies.


jcfischer (commented):

5 finding(s): 5 important.


### CodeQuality

- **[important]** `src/skills/the-algorithm/references/algorithm.md:633` — **Use the supported reflection flags**
  The new command repeats `--smarter-run`, but `algorithm reflect` only accepts `--missed-early-step`, `--missed-verify-or-parallel`, and `--highest-value-move`; this invocation fails with an unknown-option error.
  Fix: Map the three prompts to the three supported flags.

- **[important]** `src/skills/the-algorithm/references/algorithm.md:306` — **Preserve the classifier source**
  The instruction says `set effort_source: classifier`, while the preceding classifier contract says normal classifier output is `source: auto`; a direct `algorithm classify` result is therefore recorded with different provenance.
  Fix: Use the returned source, reserving `classifier` for an injected host classification if intended.

### Architecture

- **[important]** `src/skills/the-algorithm/references/capabilities.md:66` — **Do not define a generic agent invocation**
  The diff defines "`Agent(…)` | `agent`" and says it binds to a substrate with "sub-agents." `CONTEXT.md` §"agent (always qualified, never bare)" bans bare `agent`, and `docs/architecture.md` §Skills assigns loading and execution to adapters; this core table creates an unqualified cross-substrate execution primitive.
  Fix: Use adapter-declared, substrate-qualified invocation targets; reject rows lacking a matching adapter capability.

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/algorithm.md:632` — **Reflection command is nonexistent**
  The new command uses `"--smarter-run"`, but `algorithm reflect` accepts only `--missed-early-step`, `--missed-verify-or-parallel`, and `--highest-value-move`. This replacement for the JSONL echo fails as an unknown option, so the promised durable reflection is never recorded.

- **[important]** `src/skills/the-algorithm/references/algorithm.md:236` — **Canonical VSA path is wrong**
  `"task ISA at <soma-home>/memory/WORK/{slug}/ISA.md"` conflicts with the VSA Scaffold workflow's task output at `<soma-home>/memory/WORK/{slug}/VSA.md` and the core VSA store at `<soma-home>/vsa/<slug>.md`. The new portable instruction directs agents to a file the documented VSA workflow does not create.


---

jcfischer (commented):

4 finding(s): 4 important.


### CodeQuality

- **[important]** `src/cli/algorithm.ts:919` — **List compiled-in capabilities**
  The added claim that “the capability vocabulary is closed but PER-MACHINE” is false for `--list`: it returns only home-loaded definitions, while selectable built-ins `ReReadCheck` and `sequential-analysis` remain absent.
  Fix: Merge built-in definitions into `--list` output, with resolved home definitions overriding by name.

### Architecture

- **[important]** `src/algorithm-capabilities.ts:300` — **Do not use Adapter for invocation**
  The new `Adapter("<contract>")` form makes “adapter” an Algorithm capability invoked at runtime, while CONTEXT.md §adapter says an adapter “projects” only and that optional invocation belongs to `SubstrateExecutor`, “never to an adapter.” This turns a locked projection-boundary term into a public runtime capability syntax.
  Fix: Name this capability form `Contract(...)` or `Executor(...)`, reserving `adapter` for SomaAdapter projection.

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/algorithm.md:108` — **Identical classification is unproven**
  The text claims adapters "classify identically" because they project `ALGORITHM_CLASSIFIER_CONTRACT`, but the shown verification lists no cross-adapter classification fixtures or parity results. Shared pattern data alone does not establish identical adapter behavior.

- **[important]** `src/skills/the-algorithm/references/algorithm.md:54` — **Closed vocabulary claim overstates enforcement**
  The text says "nothing may be added to it at run time," then states that "every remaining installed skill" feeds the registry. The diff provides no evidence that installed skills or manifests cannot change after startup, so downstream readers may treat a dynamically discovered vocabulary as immutable.


---

jcfischer (commented):

6 finding(s): 1 blocker, 3 important, 2 suggestion.


### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:464` — **Reject unbound contract invocations**
  `definitions.set(name, buildCapabilityDefinition(name, "adapter", ...))` makes an unbound contract selectable, and `algorithm invoke` can mark it invoked using arbitrary evidence. That lets `CrossFamilyAudit` satisfy COMPLETE without the local binding the new doctrine requires.
  Fix: Track whether an adapter row is locally bound and reject invocation while it is unbound.

### Architecture

- **[important]** `src/skills/the-algorithm/references/algorithm.md:54` — **Do not call capabilities adapters**
  The new text introduces “**Adapter capabilities**” and later calls `CrossFamilyAudit` “**an adapter row**”. `CONTEXT.md` §adapter defines an adapter as the actor that performs projection and states optional invocation belongs to `SubstrateExecutor`, “never to an adapter”; these rows describe invocation contracts, not projection actors.
  Fix: Call these contract capabilities/rows; reserve adapter for `SomaAdapter` projection actors.

### Maintainability

- **[suggestion]** `src/cli/algorithm.ts:623` — **Use the registry contract type**
  The formatter restates the model as `definitions: { name: string; kind: string; phases: string[]; invoke: { target: string } }[];` instead of using the capability registry type. Registry contract changes now require updates in both modules.
  Fix: Type this parameter with `SomaHomeAlgorithmCapabilityRegistry` or `AlgorithmCapabilityDefinition[]`.

- **[suggestion]** `test/algorithm.test.ts:370` — **Centralize capability-table fixtures**
  `"| Capability | Phases | Trigger Signal | Invoke | Typical Cost |"` and its separator are repeated across each new table fixture. A table-schema change requires editing every setup block.
  Fix: Add a `writeCapabilityTable` test helper that owns the header and separator.

### HonestOracle

- **[blocker]** `src/skills/the-algorithm/references/capabilities.md:119` — **Unbound contracts remain self-attestable**
  “It cannot be invoked” and “fails the run at the gate” are false as enforcement claims: `algorithm invoke` accepts any nonempty `--evidence` and marks the selected adapter capability invoked without a local binding or tool execution. A run can therefore record a fabricated second opinion or cross-family audit and complete.

- **[important]** `src/skills/the-algorithm/references/algorithm.md:296` — **Injected classifications are still auto**
  “`classifier` is for a host that classified ahead of you” is unsupported: the Claude Code and Codex hooks forward `algorithm classify`’s returned `source`, which is `auto`, unchanged. Recording `classifier` for injected context invents provenance rather than preserving it.


---

jcfischer (commented):

6 finding(s): 5 important, 1 suggestion.


### CodeQuality

- **[important]** `src/skills/the-algorithm/references/algorithm.md:647` — **Remove the stale capability enumeration**
  The new registry rule says names must come from the installed registry, but this rule still says “Names MUST come verbatim from the v6.3.0 closed enumeration” and includes `ISA`. `ISA` is not a registered capability in the shipped table, so following this mandatory rule makes selection fail.
  Fix: Replace the enumeration with a reference to `algorithm capabilities --list`.

- **[important]** `src/algorithm-capabilities.ts:543` — **Preserve local disablement for built-ins**
  `const { definitions } = await loadSomaHomeAlgorithmCapabilityRegistry(options);` drops `unsupported`; an unresolvable local override of `ReReadCheck` or `sequential-analysis` therefore falls back to the compiled-in definition. This violates the new promise that an unresolved local row disables a capability rather than falling back.
  Fix: Carry unsupported names into run resolution and suppress matching compiled-in defaults.

### Architecture

- **[important]** `src/skills/the-algorithm/references/algorithm.md:57` — **Qualify agent terminology**
  The new text says “sub-agents, agent teams …” without naming a substrate. `CONTEXT.md` §“agent (always qualified, never bare)” bans bare `agent`; this portable reference must distinguish substrate-native primitives from Soma concepts.
  Fix: Use substrate-qualified terms, or replace with “delegated workers”.

### Maintainability

- **[suggestion]** `src/cli/algorithm.ts:929` — **Centralize registry merging**
  `"const byName = new Map(listAlgorithmCapabilityDefinitions()...)"` followed by `"for (const definition of resolved.definitions) byName.set(...)"` duplicates the default-then-override merge that the comment identifies as `definitionsForRun` precedence. Separate copies can drift when registry precedence changes.
  Fix: Export a merge helper from algorithm-capabilities and use it for both run definitions and --list.

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/algorithm.md:47` — **Registry doctrine contradicts fixed rules**
  "Four sources feed it" makes the machine registry authoritative, but the still-shipped Rules section requires a different fixed enumeration. Valid local or automatically registered capabilities are therefore simultaneously allowed and a "CRITICAL FAILURE," leaving downstream agents without one completion contract.

- **[important]** `src/skills/the-algorithm/references/capabilities.md:38` — **Local bindings do not always win**
  "a local row of the same name wins" promises an override, yet manifest-declared capabilities are registered before local rows and existing names are skipped. An adopter with a manifest-backed matching name cannot retarget or bind it locally; the manifest definition silently remains active.


---

jcfischer (commented):

5 finding(s): 4 important, 1 suggestion.


### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:524` — **Prune disabled definitions from existing runs**
  The diff promises that an unresolvable local row “also blocks the bundled row,” but registration only adds resolved definitions to a persisted run. A run created before the local override retains its old definition, so it remains selectable and invocable despite appearing in `unsupported`.
  Fix: When refreshing a run registry, remove non-default persisted definitions named in `unsupported` before registering resolved definitions.

### Architecture

- **[important]** `src/algorithm-capabilities.ts:592` — **Keep registry states mutually exclusive**
  `return { definitions: [...byName.values()], unsupported: [...resolved.unsupported] }` can return a capability in both arrays after the preceding rule says compiled-ins are “always present and are NOT suppressed.” This exported registry no longer gives consumers one coherent availability state per capability.
  Fix: Make `unsupported` disjoint from `definitions`, and expose invalid overridden rows as separate diagnostics if they must remain visible.

### Maintainability

- **[suggestion]** `src/skills/the-algorithm/references/algorithm.md:337` — **Centralize phase-transition protocol**
  The six phase sections repeat "`<prefix> algorithm advance --id <run-id>` — the deterministic gate" followed by a phase-frontmatter mirror. This policy is already defined centrally, so future changes require synchronized edits across every phase.
  Fix: Define the advance-and-mirror protocol once; keep only phase-specific follow-up in each section.

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/algorithm.md:53` — **Invocation remains self-attested**
  The claim that a selected name "you cannot back with an invocation fails the run" treats COMPLETE as proof of a real invocation. For every non-Contract capability, `algorithm invoke` records arbitrary non-empty evidence without executing or verifying its declared target, so a written assertion still clears this gate.

- **[important]** `src/skills/the-algorithm/references/capabilities.md:20` — **Not every skill is registered**
  The statement "Every remaining skill under `<soma-home>/skills`, automatically" omits substrate filtering. With `--substrate`, the loader marks skills whose manifest excludes that substrate unsupported, so this guidance falsely assures an adopter that every catalogued skill is selectable.


---

jcfischer (commented):

6 finding(s): 5 important, 1 suggestion.


### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:560` — **Remove deleted local capabilities**
  `"Registration only ever ADDS or overrides by name"` means a local-only row removed from `capabilities.local.md` is neither resolved nor `unsupported`, so its persisted definition remains selectable even though `--list` no longer shows it.
  Fix: Prune persisted definitions absent from the current resolved registry as well as unsupported ones.

- **[important]** `src/skills/the-algorithm/references/algorithm.md:300` — **List the actual fail-safe source**
  The template offers `"source: explicit /eN | classifier | context-override | auto"`, but the preceding contract says classifier output is `auto`, `explicit`, or `fail-safe` and explicitly says `classifier` is never emitted. Empty-prompt runs will be recorded with invented provenance.
  Fix: Replace `classifier` with `fail-safe` in the template.

### Security

- **[important]** `src/algorithm-capabilities.ts:794` — **Bound contracts remain self-attestable**
  `if (definition.kind === "adapter")` rejects only unbound contracts; a local `Bash(...)` or `Agent(...)` binding bypasses it, while the diff states `algorithm invoke` "records the evidence you hand it without executing the declared target." A run can therefore claim `CrossFamilyAudit` without obtaining one.
  Fix: Preserve a contract-required marker across local bindings and accept invocation only with substrate-verified execution provenance.

### Architecture

- **[suggestion]** `src/skills/the-algorithm/references/algorithm.md:47` — **Use principal instead of owner**
  “the machine's owner has the last word” violates `CONTEXT.md §principal`, which reserves `principal` for this formal documentation context.
  Fix: Replace `owner` with `principal` here and in `capabilities.md`.

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/capabilities.md:54` — **Overlay absence does not limit registry**
  "If it does not exist, only the shipped table applies" is false: manifest-declared capabilities, every installed skill, and compiled-ins still resolve. An adopter relying on this will misread what is selectable.

- **[important]** `src/algorithm-capabilities.ts:584` — **Listed registry omits selectable stale capabilities**
  "Everything a run may select" does not hold: registration prunes only names marked unsupported, retaining persisted definitions when their rows are simply removed. Such a capability remains selectable in that run but is absent from `capabilities --list`, making the claimed closed vocabulary incomplete.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:576` — **Prune unsupported-only refreshes**
  `if (definitions.length === 0) { return run; }` retains persisted capabilities when a previously valid local row is retargeted to a missing skill and no other row resolves. That contradicts the new disablement behavior: the stale capability remains selectable.
  Fix: When `unsupported.length > 0`, replace persisted home definitions even if `definitions` is empty.

### Maintainability

- **[suggestion]** `src/skills/the-algorithm/references/algorithm.md:47` — **Centralize registry-resolution doctrine**
  "Four sources feed it, resolved in this order" duplicates the canonical "Four sources, merged in this order" section in `capabilities.md`. Changes to precedence or compiled-ins now require synchronized edits in both references.
  Fix: Keep the command and closed-vocabulary rule here; link to `capabilities.md` for resolution order and built-in exceptions.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:584` — **Refresh drops non-home definitions**
  `"const withoutHomeDefinitions = { ...run, capabilityDefinitions: [] };"` clears every persisted definition before home rows are registered. Public callers can register run-specific definitions, which a later capability select or invoke silently removes when a home registry exists.
  Fix: Track home-derived definitions separately, or preserve definitions not owned by the home registry.

### Maintainability

- **[suggestion]** `src/algorithm-capabilities.ts:580` — **Remove duplicate empty-registry guard**
  `if (definitions.length === 0 && unsupported.length === 0) { return run; }` appears at both lines 556 and 580, so the second branch is unreachable. Keeping both obscures the replacement policy and invites future edits to diverge.
  Fix: Keep the first guard and delete the second copy.

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/algorithm.md:95` — **Fail-safe claim excludes nonempty prompts**
  "Unclassifiable input fails toward more rigour, never less" is false: any nonempty prompt that reaches Algorithm mode but matches no tier rule receives the default E1 with source `auto`. Readers can therefore rely on an E3-style fail-safe that only exists for an empty prompt.

- **[important]** `src/skills/the-algorithm/references/capabilities.md:140` — **Unbound contracts can be removed**
  "selecting a contract you cannot satisfy fails the run" and "It cannot be talked past" omit `remove-capability`: it accepts any nonempty reason, and COMPLETE accepts selections marked `removed`. An unbound audit can therefore be selected, removed, and bypassed without the documented decision record.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:575` — **Migrate legacy home definitions**
  `"definition.origin !== \"soma-home\""` retains every capability persisted before this PR, because those rows have no `origin`. A later refresh therefore cannot remove a formerly home-derived row that was deleted or became unsupported.
  Fix: Add a legacy migration marker or explicitly replace prior home registry rows on the first refresh.

### Architecture

- **[important]** `src/skills/the-algorithm/references/algorithm.md:54` — **Keep agent terminology qualified**
  “For a skill, agent, or command capability” uses bare `agent` in shipped Soma doctrine. `CONTEXT.md` §agent bans bare `agent` in docs and code because it collapses distinct substrate primitives.
  Fix: Use “qualified substrate-worker capability” or name the relevant substrate primitive.

### Maintainability

- **[suggestion]** `src/algorithm-capabilities.ts:605` — **Share capability precedence merge**
  `"const byName = new Map(listAlgorithmCapabilityDefinitions()..."` recreates the precedence that the new comment says is "the same defaults-then-override order `definitionsForRun` applies". `definitionsForRun` keeps a second implementation, so precedence changes require synchronized edits.
  Fix: Extract `mergeCapabilityDefinitions(overrides)` and use it from both functions.

### HonestOracle

- **[blocker]** `src/skills/the-algorithm/references/algorithm.md:108` — **Capability floor is not gated**
  It claims `algorithm advance` "refuses the transition when required capabilities ... are missing," but `assertAlgorithmCapabilitiesSatisfied` only rejects already-selected capabilities left selected/failed. A run with zero thinking capabilities, no ReReadCheck, or an under-floor E2–E5 count can COMPLETE, so calling these requirements a deterministic harness gate misleads users into trusting an unenforced quality floor.

- **[important]** `PR description:1` — **Shipped-table claim is stale**
  The description says "The shipped table is 6 rows that resolve anywhere — bundled-skill and `inline doctrine` targets only," while the diff ships eight rows, including three `Contract(...)` rows that require local binding before invocation. The portability/result claim hides the non-invocable contract rows from adopters relying on it.


---

jcfischer (commented):

3 finding(s): 3 important.


### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:811` — **Do not reject every adapter capability**
  `if (definition.kind === "adapter")` treats every adapter-kind definition as “a contract with no binding.” Skill manifests may already declare `algorithmCapability.kind: "adapter"`; those resolve through the manifest path and are now rejected even when their target is available.
  Fix: Mark Contract(...) rows explicitly as unbound contracts, and refuse only that marker.

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/algorithm.md:164` — **Existing VSA sync hook denied**
  The new text says, "There is no hook that parses or rewrites the file on your behalf." Claude Code's shipped hook runner detects edited `VSA.md` files and invokes `algorithm sync-from-isa`, which parses them into runs; this misleads users about what records run state.

- **[important]** `test/bundled-skill-references.test.ts:143` — **Whole skill bypasses advertised guard**
  The guard claims "no bundled skill" carries forbidden residue, but `const EXEMPT = new Set(["migrate-pai-purpose"])` skips that skill before checking any pattern. The stated reason covers its Claude-home references, not its PAI-tree paths or identity placeholders, so the test cannot support the universal claim.


---

jcfischer (commented):

1 finding(s): 1 important.


### CodeQuality

- **[important]** `src/cli/algorithm.ts:632` — **Only flag unbound contracts**
  `"definition.kind === \"adapter\""` marks every adapter capability as needing a local binding, but manifest-declared adapter capabilities are invocable because only `unbound === true` denotes a contract row. The list therefore contradicts actual invocation behavior.
  Fix: Use `definition.unbound === true` for the suffix.


---

jcfischer (commented):

2 finding(s): 2 important.


### Architecture

- **[important]** `src/algorithm-capabilities.ts:487` — **Scope local bindings by substrate**
  The docs say "`Agent(…)` rows bind to a substrate", but every local `Agent(...)` row becomes `buildCapabilityDefinition(name, "agent", ...)` without consulting `substrate`; one same-named local binding therefore shadows every substrate. This prevents one Soma home from representing distinct Claude Code and Codex executor bindings, contrary to `docs/architecture.md` §Adapters, where executors own substrate-local capability probing and invocation.
  Fix: Add a substrate selector to local bindings and resolve it per requested substrate, or delegate binding resolution to SubstrateExecutor.

### HonestOracle

- **[important]** `PR description:1` — **“Zero residue” is unsupported**
  The claim “`algorithm.md` — 674 lines, now zero residue” is not supported by a defined residue check and conflicts with retained text such as “`The ISA has twelve sections`” and “`Determine ISA home`” in the changed file. Downstream adopters can reasonably read this as a complete terminology cleanup when the artifact still deliberately mixes the old and new vocabulary.


---

jcfischer (commented):

7 finding(s): 6 important, 1 nit.


### CodeQuality

- **[important]** `src/skills/the-algorithm/references/capabilities.md:129` — **Describe the actual unbound-contract check**
  The text says "`algorithm invoke` **refuses** a capability whose resolved kind is still `adapter`", but the implementation rejects only `definition.unbound === true`; a manifest-declared `adapter` capability remains invocable.
  Fix: Say it refuses an unbound `Contract(...)` capability, not every `adapter` kind.

### Security

- **[important]** `src/skills/the-algorithm/references/algorithm.md:66` — **Quote classifier prompt safely**
  The new instruction ` <prefix> algorithm classify --prompt "..." [--json]` directs an executor to place the current user prompt inside double quotes. A prompt containing `$(...)`, backticks, or `"; ...` can execute shell syntax when followed literally.
  Fix: Pass the prompt as a structured argv value, or document shell-safe single-quote escaping.

### Architecture

- **[important]** `src/algorithm-capabilities.ts:459` — **Do not model contracts as adapters**
  The new contract path stores `buildCapabilityDefinition(name, "adapter", ...)`, while CONTEXT.md §adapter says optional invocation belongs to `SubstrateExecutor`, “never to an adapter.” Using the adapter kind for a declared contract expands that boundary violation into persisted/public capability state.
  Fix: Add a distinct `contract` capability kind and migrate these rows to it.

- **[important]** `src/algorithm-capabilities.ts:824` — **Avoid bare Agent invocation syntax**
  The new public guidance tells adopters to use `Agent(…)`; CONTEXT.md §agent says bare `agent` is banned in Soma docs and code and requires a substrate qualifier. This makes a prohibited ambiguous term part of the capability-table surface.
  Fix: Use a substrate-qualified invocation form, or a neutral executor-oriented term.

### Maintainability

- **[nit]** `src/algorithm-capabilities.ts:432` — **Correct local-table precedence comment**
  The comment says local rows are "Read after the bundled table", but the new flow calls `registerCapabilityTableRows(localMarkdown, ...)` before the bundled table. Conflicting precedence explanations make this resolution order harder to maintain.
  Fix: Change it to say the local table is read before the bundled table and wins on collisions.

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/capabilities.md:140` — **Adapter kind is not the refusal condition**
  The document says `algorithm invoke` "refuses a capability whose resolved kind is still `adapter`", but the implementation refuses only `definition.unbound === true`; manifest-declared `adapter` capabilities remain invocable. This misstates which registered capabilities are usable and makes the documented gate broader than the shipped behavior.

- **[important]** `PR description:1` — **Registry verification count is contradictory**
  The description first claims the bundle registry becomes "14 / 0", then reports "Registry probe against the bundle: 10 definitions, 0 unsupported." The current bundle resolves 12 home definitions and 14 after compiled-ins, so the verification statistic cannot support the stated registry result.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:582` — **Legacy caller definitions are dropped**
  `"An ABSENT origin is treated as home-derived"` and `filter((definition) => definition.origin === "caller")` remove every pre-existing direct registration, because those runs predate `origin`. Adapters were already instructed to register capabilities directly, so their persisted definitions disappear on the first refresh.
  Fix: Migrate or preserve origin-less definitions until their producer can be identified; do not use missing origin as a deletion signal.

- **[important]** `src/skills/the-algorithm/references/algorithm.md:340` — **Audit references deleted enumeration**
  The mandatory audit says each name must appear in `"the closed enumeration"` and rejects `"anything outside the list"`, but this change removed that list in favor of the per-machine registry. Following this gate leaves no enumeration to check and can reject valid installed capabilities.
  Fix: Require names to appear in `algorithm capabilities --list`.

### Maintainability

- **[suggestion]** `src/types.ts:350` — **Centralize capability vocabulary**
  `AlgorithmCapabilityKind = "skill" | ... | "contract"` is immediately duplicated by `AlgorithmCapabilityContract = "skill" | ... | "contract"`. New kinds now require synchronized edits to parallel declarations and runtime validation.
  Fix: Export one capability-kind tuple; derive both aliases and runtime membership checks from it.

- **[suggestion]** `src/algorithm-capabilities.ts:314` — **Remove stale adapter terminology**
  The new comment says `The stored kind is still "adapter"`, but this diff creates `buildCapabilityDefinition(name, "contract", ...)`. This leaves the parser's `adapterInvocationTarget` name and its documentation describing different concepts.
  Fix: Rename it to `contractInvocationTarget` and update the comment to describe `contract`.

### HonestOracle

- **[important]** `PR description:1` — **Guard does not check named models**
  The description says “Cleared means what the guard measures and nothing more” and includes “no named external models,” but `test/bundled-skill-references.test.ts:132-139` only checks Claude-home paths, specified PAI paths, and identity placeholders. A future named-model reference passes the advertised guard, misleading adopters about residue protection.


---

jcfischer (commented):

3 finding(s): 3 important; 1 lens(es) failed to run: HonestOracle.


### CodeQuality

- **[important]** `src/skills/the-algorithm/references/algorithm.md:170` — **Do not claim a nonexistent VSA sync hook**
  The new text says “Claude Code's hook runner watches for an edited `VSA.md` and calls `algorithm sync-from-isa`,” but the installed PostToolUse runner is configured only for `writeback-tool`; edits will not reach the Algorithm run as documented.
  Fix: Document manual `algorithm sync-from-isa` use, or register a hook handler that invokes it.

- **[important]** `src/algorithm-capabilities.ts:823` — **Do not reject every contract-kind definition**
  `ALGORITHM_CAPABILITY_KINDS` now accepts `"contract"`, but `if (definition.kind === "contract")` rejects it unconditionally. A manifest- or caller-registered contract with a real target is therefore selectable yet permanently uninvokable.
  Fix: Restrict the unbound marker to parsed Contract rows, or reject `kind: "contract"` in manifest/caller registration.

  - raw: undefined
  - balanced-largest: undefined
  - trailing: undefined
--- output ---
{"summary":"The registry presents agent rows as substrate-capable without checking that the selected substrate can execute them.","findings":[{"path":"src/skills/the-algorithm/references/capabilities.md","line":104,"severity":"important","title":"Agent availability is not enforced","rationale":"“`Agent(…)` rows need a substrate that can spawn a short-lived worker” is not enforced: local Agent rows resolve for every substrate, and invocation records supplied evidence without executing or checking a substrate worker. The registry can therefore present an unsupported Agent binding as usable.","}]}


---

jcfischer (commented):

4 finding(s): 4 important.


### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:330` — **Reject malformed Contract rows**
  `const contract = /Contract\("([^"]+)"/.exec(value)` accepts `Contract("audit"` without the closing `)`, then registers it as a `contract`. A malformed row becomes selectable and uninvokable instead of being dropped.
  Fix: Require the closing `)` in the Contract regex after stripping Markdown emphasis.

### Architecture

- **[important]** `src/algorithm-capabilities.ts:691` — **Keep home provenance kernel-owned**
  `const origin = clone.origin ?? "caller"` trusts the new public `origin?: "soma-home" | "caller"` field, and the later contract guard treats `"soma-home"` as privileged. A direct caller can impersonate the loader and alter refresh ownership, crossing the core-owned canonical-state boundary in `docs/architecture.md` §Adapters.
  Fix: Reject caller-supplied home origin; attach it only through an internal home-registration path.

- **[important]** `src/skills/the-algorithm/references/capabilities.md:43` — **Provide an overlay migration**
  The new contract says "`soma install` rewrites every bundled file" and a row in this file "will be overwritten", while `capabilities.local.md` becomes the durable owner. Existing principal capability rows have no migration or backup before an install removes them, making this an irreversible configuration-surface change.
  Fix: Add an install-time migration or explicit pre-overwrite backup for existing custom rows.

### HonestOracle

- **[important]** `src/algorithm-capabilities.ts:698` — **Contract minting is forgeable**
  The comment claims “Only the `Contract("…")` parser may mint one,” but `AlgorithmCapabilityDefinition.origin` is public and a caller can pass `origin: "soma-home"` with `kind: "contract"`, bypassing this guard. Downstream code is misled into treating an untrusted provenance label as proof that the parser produced the definition.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/bundled-skills.ts:72` — **Do not promote old bundled table wholesale**
  `if (existing === undefined || incoming === undefined || existing === incoming) return undefined;` treats every version difference as user customization, then writes `+ existing` into the higher-precedence overlay. A normal upgrade from an older unmodified bundle therefore freezes all old rows locally and masks the new bundled capability table.
  Fix: Only migrate explicitly identifiable user-added rows, or require an opt-in migration instead of copying the full prior file.

### HonestOracle

- **[blocker]** `src/bundled-skills.ts:60` — **Upgrade migration reactivates PAI rows**
  The claim "the existing content becomes the overlay" and "a row that duplicates a shipped one simply wins" treats every prior bundled table as adopter configuration. A normal upgrade copies the old PAI-era table into the local-first overlay, so rows such as the former PAI-backed Advisor win over this PR's portable replacements; existing installations retain the residue this PR claims to clear.

- **[important]** `src/algorithm-capabilities.ts:315` — **Contract representation comment is false**
  The comment says "The stored `kind` is still `\"adapter\"`", but this diff adds `contract` to the exported kind union and constructs contract rows with `buildCapabilityDefinition(name, "contract", ...)`. This misstates the public capability state and contradicts the shipped reference.


---

jcfischer (commented):

4 finding(s): 4 important.


### CodeQuality

- **[important]** `src/skills/the-algorithm/references/capabilities.md:12` — **Correct documented registry precedence**
  The reference lists "Skills declaring `algorithmCapability` in their manifest" before "`capabilities.local.md`, the adopter's own table", while the new resolver registers local rows before manifest definitions. Adopters will expect manifest metadata to win when their local binding actually takes effect.
  Fix: List `capabilities.local.md` first, then manifest-declared capabilities.

### Architecture

- **[important]** `src/types.ts:355` — **Separate declaration from invocation kinds**
  `ALGORITHM_CAPABILITY_KINDS` adds `"contract"` while `AlgorithmCapabilityContract = AlgorithmCapabilityKind` publishes a declaration-only state as an execution-kind vocabulary. Consumers and future executors must now accept an uninvocable kind or require a breaking narrowing change, contrary to the projection/execution split in `docs/architecture.md` §Adapters.
  Fix: Use a definition/declaration-kind union that includes `contract`, and a separate invocation-kind union that excludes it.

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/algorithm.md:324` — **Auto-inclusion is only prose**
  “auto-include at E3/E4/E5” has no backing: selection occurs only through an explicit `algorithm capabilities --capability`, while the gate requires merely some selected capability. E3–E5 runs can therefore complete without the promised cross-family capability.

- **[important]** `src/skills/the-algorithm/references/capabilities.md:53` — **Three-column rows misstate phase behavior**
  The advertised `| Capability | Trigger Signal | Invoke |` shape has no Phase cell, yet the parser substring-matches its Trigger Signal cell as a phase. A trigger containing “think” silently registers at `think` instead of the documented three-column default of `plan`.


---

jcfischer (commented):

5 finding(s): 5 important.


### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:687` — **Preserve caller capability bindings**
  `"nextDefinitions.set(definition.name, { ...cloneCapabilityDefinition(definition), origin: \"soma-home\" })"` overwrites a retained caller definition with the same name. A caller-provided `Advisor` binding is therefore replaced by the bundled unbound contract on each CLI refresh.
  Fix: Do not overwrite an existing `origin: "caller"` definition during home refresh; add a collision test.

- **[important]** `src/algorithm-capabilities.ts:727` — **Reject contract-marked caller definitions**
  The public guard only checks `"if (clone.kind === \"contract\")"`. A caller can register `kind: "skill"` with `invoke.contract: "contract"`; it selects successfully but invocation always refuses it.
  Fix: Reject public definitions with `invoke.contract === "contract"` as well.

### Architecture

- **[important]** `src/skills/the-algorithm/references/algorithm.md:249` — **Use principal in shipped doctrine**
  The diff adds “restate the user's request” and “re-read the user's message.” CONTEXT.md §principal says `user` is never used in `docs/` or `src/`; this is shipped source doctrine.
  Fix: Use “principal's request” and “principal's message.”

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/algorithm.md:114` — **Gate inventory omits enforced conditions**
  The claim “What it actually refuses, per transition” says OBSERVE→THINK requires a probe, but the live gate also refuses when there are no criteria; it says EXECUTE requires “recorded build changes,” while the code accepts any nonempty changelog regardless of phase. This presents an incomplete and stronger enforcement contract than the harness has.

- **[important]** `src/skills/the-algorithm/references/algorithm.md:642` — **Reflection parity claim is false**
  “using the same predicates the live gates enforce” is not true: `learnGateClean` adds `criteria.length > 0`, while the LEARN gate does not, and `completeness` is not a COMPLETE-gate predicate. Reflection consumers can therefore read a false gate miss as an enforced-gate result.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/adapters/shared/algorithm-rendering-contract.ts:58` — **PAI banner still ships**
  The description claims "`♻︎ Entering the PAI ALGORITHM…` → `Entering the Algorithm`", but this projected `the-algorithm/SKILL.md` still emits `♻︎ Entering the PAI ALGORITHM… (Soma)`. Downstream Codex/Grok users will still receive the residue the PR says was removed.


---

jcfischer (commented):

1 finding(s): 1 suggestion.


### HonestOracle

- **[suggestion]** `src/algorithm-capabilities.ts:440` — **Local-table order is misstated**
  The comment says `capabilities.local.md` is "Read after the bundled table," while the implementation immediately below reads and registers the local table before the bundled table. That conflicts with the PR’s own claim that a local row wins by name and misstates the governing precedence rule.


---

jcfischer (commented):

3 finding(s): 2 important, 1 suggestion; 1 lens(es) failed to run: HonestOracle.


### Architecture

- **[important]** `src/skills/the-algorithm/references/capabilities.md:121` — **Keep capability probing out of adapters**
  The text says “the rest either cannot or have no adapter that says so,” but `docs/architecture.md` §Adapters assigns substrate-local capability probing and invocation to `SubstrateExecutor`, never an adapter. This directs future bindings to the projection layer instead of the execution boundary.
  Fix: Say the substrate has no `SubstrateExecutor` that exposes the worker capability.

### Maintainability

- **[suggestion]** `src/algorithm-capabilities.ts:738` — **Centralize unbound-contract predicate**
  `clone.kind === "contract" || clone.invoke.contract === "contract"` is repeated later as `definition.kind === "contract" || definition.invoke.contract === "contract"`. A future contract representation change must update both guards.
  Fix: Extract `isUnboundContractCapability(definition)` and retain the distinct error messages.

  - raw: undefined
  - balanced-largest: undefined
  - trailing: undefined
--- output ---
{"summary":"The claimed safe upgrade path has a confirmed data-loss hole after its first use.","findings":[{"path":"src/bundled-skills.ts","line":67,"severity":"blocker","title":"Backup claim fails on later upgrades","rationale":"“A backup answers the real risk — irreversible loss” is false after the first migration: line 86 skips preservation when the backup exists, then installation overwrites later edits to `capabilities.md`. A second upgrade silently deletes those rows.","}]}


---

jcfischer (commented):

### Performance

- **[suggestion]** `src/bundled-skills.ts:97` — **Backup scan grows with every upgrade**
  `"for (let attempt = 2; ; attempt += 1)"` performs an awaited `"readFile(backup, \"utf8\")"` for every prior backup. Because numbered backups are never bounded, a repeatedly edited installation accumulates serial filesystem round trips on each later install.
  Fix: Cap/prune retained backups or keep an index of stored content hashes.

### HonestOracle

- **[important]** `src/cli/algorithm.ts:90` — **List hides unresolved local overrides**
  The help promises `--list` shows the registry “plus any rows that resolved to nothing.” Yet `mergedAlgorithmCapabilityRegistry` filters unsupported names that collide with compiled-ins, so an unresolvable local `ReReadCheck` or `sequential-analysis` row is omitted rather than reported. This misleads adopters into believing their overlay resolved cleanly.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/bundled-skills.ts:100` — **Avoid truncated-hash backup collisions**
  `digest("hex").slice(0, 8)` makes two distinct customised tables share one backup path; `if (alreadySaved) return undefined` then overwrites the newer table without saving it.
  Fix: Use the full digest, or compare the existing backup content before treating it as deduplicated.

### Maintainability

- **[suggestion]** `src/bundled-skills.ts:53` — **Keep installer JSDoc adjacent**
  The new JSDoc follows the existing "Copy every bundled skill" JSDoc but precedes `backupCustomisedCapabilityTable`, so documentation tooling associates the nearer block with the helper and leaves `installBundledSkillsIntoHome` undocumented.
  Fix: Move the installer JSDoc to immediately before `installBundledSkillsIntoHome`.

### HonestOracle

- **[important]** `src/skills/the-algorithm/references/capabilities.md:54` — **Upgrade backup filename is false**
  It says install “saves the previous `capabilities.md` alongside as `capabilities.pre-upgrade.md`,” but the implementation writes `capabilities.pre-upgrade.<hash>.md`. An adopter following this migration guidance will look for a file that does not exist and may conclude their preserved rows were lost.


---

jcfischer (commented):

1 finding(s): 1 important.


### CodeQuality

- **[important]** `src/bundled-skills.ts:94` — **Backup collision cap overwrites saved content**
  The loop `for (let attempt = 2; held !== undefined && attempt < 100; attempt += 1)` leaves `target` as `backup` when all candidates through `.99` exist; `await writeFile(target, ...)` then overwrites the original colliding backup and loses its saved table.
  Fix: Keep probing until a free suffix, or throw before writing when none is found.


---

jcfischer (commented):

1 finding(s): 1 important.


### CodeQuality

- **[important]** `src/bundled-skills.ts:91` — **Compare backup payloads exactly**
  `"if (held?.endsWith(existing) === true) return undefined;"` (and the same check for numbered backups) treats a different colliding backup whose content merely ends with the current table as already saved, then overwrites the current table without a backup.
  Fix: Compare the stored payload to the exact generated backup content, not `endsWith(existing)`.


---

jcfischer (commented):

3 finding(s): 3 important.


### CodeQuality

- **[important]** `src/algorithm-capabilities.ts:527` — **Do not hide local-table read errors**
  `readFile(join(referenceDir, CAPABILITY_LOCAL_REFERENCE_FILE), "utf8").catch(() => "")` treats permission and I/O failures as an absent overlay. That can silently re-enable a bundled capability the local table was meant to override or disable.
  Fix: Ignore only ENOENT; rethrow other read errors.

### Security

- **[important]** `src/bundled-skills.ts:124` — **Refuse dangling backup symlinks**
  `readFile(candidate, "utf8").catch(() => undefined)` treats a dangling symlink as vacant, then `await writeFile(target, payload, "utf8")` follows it. A writable skill directory can pre-place a predictable dangling backup symlink and make install create or overwrite an arbitrary path.
  Fix: Use lstat and reject symlink candidates before selecting or writing them.

### Architecture

- **[important]** `src/bundled-skills.ts:156` — **Upgrade migration runs during reproject**
  `if (name === "the-algorithm")` now calls `backupCustomisedCapabilityTable` inside the shared installer. Since `reproject` and `upgrade` both call that installer, this adds migration to reproject, contradicting CONTEXT.md §Lifecycle verbs: reproject re-emits a projection while upgrade “Migrates and reprojects.”
  Fix: Pass an explicit upgrade/migration flag and create the backup only for `soma upgrade`.


---

Issue #576, opened 2026-08-08
One query is the frontier: fold confirmation into discovery


jcfischer:
Implements the decision on #530: `soma graph frontier` becomes **one GraphQL
round trip** for a map that fits the query, discovery *is* confirmation, and the
read path stops doing work nothing reads.

Measured baseline on map #495, live: **16 spawns / 9 606ms**, 100% of it in `gh`.
The one-query form was measured at **1 spawn / 1 028ms** returning the identical
hand-checked set. That is the target shape, not an aspiration.

## What to build

1. **One query carries the whole frontier.** Extend the #564 subtree walk's
   selection with `assignees`, `author`, `blockedBy`, `body` and `url` per node
   — everything `NodeState` holds. `WorkGraph.frontier` then filters in memory
   instead of re-reading each candidate.

2. **Amend §2.4's confirmation clause.** Today: "confirms every candidate by
   direct fetch". After: **discovery must be a live read of the authoritative
   store, and when it is, it confirms.** A backend that discovers through a
   lagging index still owes the second read — the obligation attaches to the
   staleness, not to the ceremony. Say plainly that this supersedes #492
   correction 3, and why: the fetch was written against a search index this
   backend never had, and fifteen reads spread over 9.6s are *less* coherent
   than one snapshot.

3. **Recompute the query cost, do not eyeball it.** With `assignees` and
   `blockedBy` per node, 50/30/20 scores **534 800** against the 500 000 ceiling
   — over. Level 3 narrowed to 10-wide clears it. Put the arithmetic in the
   comment next to the page sizes, as #571 had to, and state that adding any
   field with a connection re-opens the sum.

4. **`close` stops re-reading what it just read.** `runClose` calls
   `graph.readNode(ref)` and then `findGraphRoot(ref, …)` starts its walk at the
   same ref. Pass the state already in hand. **Not a cache** — #530 asked
   whether a per-invocation `NodeState` cache is the right shape, and the answer
   is no: a cache opens the staleness window the question feared, an argument
   cannot.

5. **`addBlockingEdge` shares one ancestor walk** across the `--blocked-by`
   edges of a single `add`, instead of re-running the cycle-check BFS per edge.
   Same invocation, so no staleness window opens.

## Boundaries

- **No measurement target.** The principal chose the standing rule "do no
  obviously-wasted work" over a number, so do not add a latency assertion to the
  suite — it would measure the connection, not the code. Spawn *counts* are
[...]

Issue #495, opened 2026-08-02
Wayfinder map: implement the work graph (DD-16)


jcfischer:
## Destination

The **work graph primitive live on main**, per `docs/work-graph.md` and DD-16: typed `GraphStore` seam with the GitHub backend, `soma graph` verbs (frontier / node / claim / add / close-with-probes), the **orienteer** skill re-based onto them (merged *together* — ships-with-consumer, #484 clause 4), autonomy policy attached, planSteps `nodeId` bridge enforced, and the GitHub machine account provisioned so HITL receipts flip from degraded to verified.

## Notes

- **Execution override:** unlike the default plan-don't-do, this map carries execution — its tickets deliver code and provisioned infrastructure, not just decisions. Spec questions are settled (DD-16); a ticket that surfaces a genuine spec change goes back through `design/design-decisions.md`, not this map.
- **Dogfood contract:** this map is the second real work-graph walk (after #477/#492). Until the `soma graph` verbs exist we walk it wayfinder-style with raw `gh` — legitimate, since the tracker is the sole authoritative store (#491). Each ticket body carries the node vocabulary from the spec (`autonomy`, `kind`, probes where AFK-closable) so the walk exercises the contract before the code exists. All receipts run `attestation: unverified` until the machine-account ticket closes — per spec §3.2, visible state, never silent.
- Ground every session in: `docs/work-graph.md` (the spec), DD-16 in `design/design-decisions.md`, `CONTEXT.md` (work graph term entry), `docs/governance-event-runtime-policy.md` (policy surface for §4).
- Orienteer doctrine (#492): scaffold nodes attach below their spawning ticket, never to the map — for **provenance**, not to hide them. #557 removed the old rationale (that burying scaffold keeps the frontier from reporting ungated work): gating is what a blocking edge means, not what depth means.
- Repo conventions: TypeScript/Bun. PRs are **not** gated on an approving review — #598 dropped `required_approving_review_count` to 0 on `protect-main` after finding no soma PR had ever been approved: every PR is authored by `jcfischer` and every review posted as `jcfischer`, and an author cannot approve their own PR, so reviews land `COMMENTED` by construction. Review is advisory and still worth requesting (sage's tail: ContextDrift lens is broken, sage#104 — verdict pins at changes-requested; good-enough call is the principal's). The status checks — `gate / confidentiality-gate` and `Deterministic portability` — still gate, and direct pushes to main are still refused.
[...]

PR #578, opened 2026-08-08
feat(work-graph): one query is the frontier (#576)

### LINKED ISSUE #530: What is the soma graph read path allowed to cost?

jcfischer:
Implements #576, which resolves #530. Map: #495.

> **Description rewritten at cycle 5** — it had drifted from the diff across four review rounds, still describing two changes that were reverted and quoting a query budget the source now corrects. Both were sage findings. What follows describes the branch as it stands.

## Measured, not estimated

`soma graph frontier 495` against the live map, instrumented to count and time every backend call:

```
before   wall 9606ms · 16 spawns · 9602ms in gh (100% of wall)
            6× graphql            avg 670ms
            5× GET issues/N       avg 605ms
            5× GET .../blocked_by avg 512ms

after    wall  ~950ms ·  1 spawn  ·  ~99% of wall in gh
```

Same map, same four nodes. Reproduce either with `bun run measure-graph-read-path -- --root 495 --node 530`, added here.

Two facts shaped the fix: **there is no compute to optimise** — all of it is subprocess and network — and the per-spawn floor is ~600ms, so the only lever is how many spawns there are. Five of those six GraphQL calls were the parent lookup `frontier` discards: **~3.4s of 9.6s producing output nothing reads.**

**Spawn count is the claim.** The wall-clock figures are network-dependent, which is why the property is pinned by a test that counts transport calls against a fake, not by a latency assertion.

## What changed

GraphQL exposes `assignees`, `author`, `blockedBy` and `body` on `Issue`, so the #564 subtree walk now selects everything `NodeState` holds and the seam returns confirmed states instead of candidate refs (`listCandidateFrontier` → `readSubtree`, a **breaking** seam change, recorded in `CHANGELOG.md`). `parent` comes free — the walk knows the edge it arrived on. `frontier` became a pure filter over that one read.

## The §2.4 amendment strengthens the guarantee

Confirm-by-direct-fetch (#492 correction 3) was written for a discovery step assumed to be a **lagging search index**. This backend has never searched. So the old shape took one live read and then fifteen more **spread across 9.6 seconds** — reporting a subtree blended from observations up to ten seconds apart. The second read wasn't making the answer more trustworthy; it was making it less coherent.

New clause: **discovery must be a live read of the authoritative store, and when it is, it confirms.** A backend that discovers through a stale index still owes the second read, internally, because the obligation attaches to the staleness and not to the ceremony.

Recorded as a **spec** amendment, not a DD one — DD-16 delegates mechanics to the spec and no normative clause moves. Precedent: #549 dropped §3.2's principal-comment clause the same way.

## What removing the second read made dangerous

Two short-read risks go from harmless to fatal, because nothing re-checks downstream: a short `assignees` page makes a claimed node look **unclaimed**, and a short `blockedBy` page makes a blocked node look **takeable** — false positives on the frontier. Both connections carry `totalCount`, and a node that comes back short is repaired by a direct `readNode`, whose REST paths paginate. That is the one place the two-phase read is still the right tool, and the repair keeps the walk's `parent` rather than letting a second lookup answer.

## #530's findings 3 and 4 were built and then reverted

Both are in the diff only as comments explaining why not, which is the honest outcome and the most interesting thing this PR found: **two of the four "wasted reads" were buying freshness.**

- **`close` re-reading the node `runClose` just read.** Seeding it meant deriving the root, its author and the whole attestation from a state read *before* the probes — and probes are the long part of a close, declared with timeouts up to 900s. ~600ms buys a reading taken after the work.
- **Sharing the ancestor reading across an `add`'s edges.** The cycle check is already best-effort (no compare-and-swap on GitHub), and sharing widens the window from one check to the whole batch — which is exactly how a path to `blocked` goes unseen and the cycle gets written.

Deferred and recorded on #530 rather than dropped: the two remaining conditional round-trip costs, re-rooting a truncated subtree and repairing a short-read node. Neither fires on any graph that exists — the repair needs a node with more than 10 assignees or 20 blockers.

## Query budget — measured, after four wrong models

A node costs `1 + ASSIGNEE_PAGE + BLOCKER_PAGE` = **31**; `author{login}` is an object, not a connection, and is free — which is what every earlier version of this arithmetic got wrong.

```
  L1   r=50            50 × 31       =    1 550
  L2   r=1 250      1 250 × 31       =   38 750
  L3   r=12 500    12 500 × 31       =  387 500
  bottom-row totalCount probe        =   12 500
  the root's own two connections     =       30
                                       ---------
                               total    440 330   (88% of the 500 000 ceiling)
```

`40 330 + 40 000w` for level-3 width `w`, confirmed against GitHub in both directions: **w=12 refused at 520 330, w=15 refused at 640 330**, predicted exactly. So 11 is the widest that fits and 10 is what we run.

The comment in the source leads with the reproduction rather than the table: `MAX_NODE_LIMIT_EXCEEDED` states the score, so send the query.

## Verification

- `bun test` 2306 pass / 0 fail; `bunx tsc --noEmit` clean.
- **One round trip is pinned by counting transport calls** in `test/work-graph-github.test.ts`.
- Falsifiable by mutation: restoring the second read fails 1, trusting a short `assignees` page fails 1, dropping `parent` on the repair path fails 1, plus the three #564 mutations (prune closed subtrees fails 3, ignore truncation fails 2, drop the `seen` guard does not terminate).
- Live numbers come from a script that now ships; the installed `soma` binary runs from the main checkout, so the CLI cannot exercise this branch until merge.


jcfischer (commented):

5 finding(s): 1 blocker, 4 important.


### CodeQuality

- **[important]** `src/cli/graph.ts:768` — **Re-read the closing node for root derivation**
  `"findGraphRoot(ref, async (nodeRef) => await graph.readNode(nodeRef), state)"` reuses state fetched before probes run. If the issue is re-parented during those probes, root author, veto handling, and attestation are derived from the old graph.
  Fix: Do not pass `state`; re-read the start node when deriving the root after probes.

### Architecture

- **[important]** `src/work-graph.ts:355` — **Preserve GraphStore migration path**
  The diff replaces `listCandidateFrontier(root: NodeRef): Promise<NodeRef[]>` with `readSubtree(root: NodeRef): Promise<NodeState[]>`. `GraphStore` is the exported backend seam, so existing backend implementations break and must adopt a materially stronger contract without a compatibility path.
  Fix: Keep a deprecated adapter during migration, or mark this as an explicit breaking API release.

### Performance

- **[important]** `src/work-graph-github.ts:225` — **Hydrates every historical node**
  `"number title state body url databaseId author{login} assignees(first:${ASSIGNEE_PAGE})... blockedBy(first:${BLOCKER_PAGE})"` is selected for every traversed node. A 50×25×10 closed subtree returns no frontier nodes but still pays the documented 454,100-node query budget and transfers/parses every body, a large regression from the prior status-only walk with no candidate hydrations.
  Fix: Use a narrow traversal, then batch-hydrate only nodes that pass the frontier predicate.

### HonestOracle

- **[important]** `docs/work-graph.md:253` — **Traversal is not one observation**
  “where one traversal is a single observation” is contradicted by the implementation’s required pagination and re-rooted follow-up queries. A returned subtree can still blend states from those sequential calls, so calling it confirmed/current misleads callers about its coherence.

- **[blocker]** `src/work-graph.ts:811` — **Batch cache admits stale cycle checks**
  “no staleness window opens” only considers this invocation’s own writes. Another actor can change a cached ancestor while `await`ing an earlier edge; `reaches` then reuses that stale state and can miss a path to `blocked`, allowing the cycle the method claims to reject.


---

jcfischer:
## Cycle 1 — three fixed (two by reverting my own work), two pushed back

`6e1a9c4`.

### The blocker and finding 1 were the same mistake, and it's worth naming

Both are #530 findings I implemented as pure waste **because they measure as pure waste**, without asking what the redundant read was buying. Reverted both.

**`close` re-reading the node it just read.** You're right and the window is worse than the diff suggests: the seed was taken before the probes run, and probes are the long part of a close — `command` probes are declared with timeouts up to 900s, and #527 exists because nothing caps their sum. A re-parent in that window derives the root, its author, and the whole attestation from a graph that no longer exists. ~600ms buys a reading taken *after* the work, which is the only one that describes what is being closed.

**The shared ancestor reading.** My comment claimed "no staleness window opens" and reasoned only about our own writes — the exact shape of overclaim this PR has already produced twice. A concurrent writer can change a cached ancestor while we await an earlier edge, and a stale ancestor is how a path to `blocked` goes unseen and the cycle gets written. The check was *already* best-effort (no compare-and-swap on GitHub, so read-then-write is never atomic), but that is a reason not to widen the window, not a licence to. And the saving only exists for an `add` with 3+ blockers with overlapping ancestries. `addBlockingEdges` is gone; the CLI is back to its per-edge loop.

So two of #530's four findings turn out **not to be waste at all**, which is a more useful result than the fix would have been.

**"One traversal is a single observation"** — also right, also overstated. Pagination and re-rooting are extra calls, so a subtree needing them is still blended. Narrowed to the true version: one observation for a graph that fits a single request, strictly fewer and closer-spaced otherwise. Still decisive against fifteen reads over 9.6 seconds; just not absolute.

### Pushing back: the migration adapter

Declining. There are no other `GraphStore` implementations — #491 decided a second backend "lands only with its first real consumer," and this map's **Out of scope** says so explicitly. The only implementors are `GitHubGraphStore` and three in-repo test fakes, and the compiler found all four.

A deprecated adapter would be a compatibility shim for zero callers, which DD-16 clause 3 rules out by name: *"every gate names its consumer; zero-consumer structure gets deleted."* This repo also has direct evidence that the alternative is worse — #525 narrowed away a `"comment"` ratification kind that nothing ever produced, on the grounds that a union member with no producer reads as a path the system has.

On "explicit breaking API release": fair, and it is recorded — the seam change is in the spec §2.5 diff, the commit, and #576. If you want a changelog entry as well, say so and I'll add one.

### Pushing back: hydrating every traversed node

Declining the fix, accepting the observation — which is now recorded next to the field selection rather than only here.

The measurement contradicts "a large regression": same map, same answer, **9 606ms → 929ms**. And the 454 100 figure is a *ceiling* GitHub checks the query against, not work performed — the real map's response is 57KB.

The proposed shape — narrow traversal, then batch-hydrate the survivors — is **two round trips, always**, and the narrow pass cannot be very narrow: the frontier predicate needs `state`, `assignees` and `blockedBy`, so all three connections stay. What one-pass over-fetches is `title`, `url` and `body` for nodes that get discarded. At ~600ms per round trip, the second call has to beat the bytes it saves, and at every size measured so far it does not.

Where you are right is that this scales with the *closed* history, which only grows — a map ten times this one pays ten times the payload for the same handful of frontier nodes. That is a real crossover and I do not know where it is, so the docblock now states the trade, the 57KB measurement, and the condition to revisit under: when payload rather than round trips is what a real map is paying.

Full suite green (2306/0), `tsc` clean, frontier re-measured after the reverts at 1 spawn / 929ms.

jcfischer (commented):

### CodeQuality

- **[important]** `src/work-graph-github.ts:557` — **Preserve parent during state repair**
  `"states.push(node.stateTruncated ? await this.readNode(node.state.ref) : { ...node.state, parent });"` returns the repaired state without `parent`, despite `readSubtree` promising membership provenance for every child. Any node with more than the selected assignees or blockers loses its subtree parent in frontier output.
  Fix: Use `{ ...(await this.readNode(node.state.ref)), parent }` in the truncated branch.

### HonestOracle

- **[important]** `src/work-graph.ts:322` — **Confirmation is an unchecked backend promise**
  "**This read confirms**" and "a backend ... still owes the second read" replace WorkGraph's enforced re-fetch with prose on `readSubtree`. The interface exposes only `Promise<NodeState[]>`, so a stale-index adapter can satisfy the type and make `frontier` report stale false positives as confirmed.

- **[suggestion]** `docs/work-graph.md:263` — **Observation reduction is overstated**
  "strictly fewer and more tightly spaced observations otherwise" is false for a paginated or re-rooted subtree with no open candidates: the old walk made the same traversal calls and then made zero confirmation reads. The change can be equal, not strictly fewer.


---

jcfischer:
## Cycle 2 — the bug is fixed, the suggestion taken, the architecture finding answered

`bd3c1f2`.

**The repair path dropped the membership edge** — real bug, and my own repair test walked past it. When `assignees` or `blockedBy` comes back short the node is repaired with a direct `readNode`, and I returned that state as-is, so the `parent` the walk had just established was replaced by whatever `readNode`'s independent parent lookup resolved. Applied on both branches now, and the test asserts it — confirmed falsifiable by putting the bug back.

Worth saying why this matters beyond the field being absent: the seam promises *the edge the walk arrived by*. A second lookup answering instead means the two can disagree about the edge we are standing on, and #557 was caused by exactly that class of thing — a membership relation nobody had pinned down.

**"Strictly fewer" was wrong at the edge** — taken. For a subtree that pages or re-roots and yields no candidates, the old shape made the same traversal calls and then zero confirmation reads, so the two are equal. Now: never more, equal in that case, far fewer whenever the subtree yields any candidate.

### On "confirmation is an unchecked backend promise"

The observation is accurate; the framing that it *replaced* enforcement is what I'd push back on.

The old second read was `readNode` — another method on the same seam, implemented by the same backend. A stale-index adapter could return stale data from `readNode` exactly as easily as from `readSubtree`. So the contract layer was never verifying liveness; it was calling a different method on the same untrusted implementation and inheriting whatever freshness that method had. What changed is the number of calls, not who is trusted.

Liveness now sits alongside the seam's other unverifiable obligations, which is the honest place for it: `claim` must re-read after writing, `createNode` must assign the id and refuse `auto` with zero probes, `readSubtree` must not truncate in silence. None of those is checkable from the contract layer either — a seam is a set of promises, and the type only carries their shape.

I did consider making it checkable — a declared `discovery: "live" | "indexed"` capability, with `WorkGraph` re-confirming when a backend says `indexed`. I left it out because it would be a branch with no implementation to exercise it: there is one backend, and #491 put a second one behind "its first real consumer". DD-16 clause 3 rules that out by name, and #525 is the local precedent — a `"comment"` ratification kind that nothing produced was narrowed away for reading as a path the system has.

If a stale-index backend is ever written, that flag is the right shape and this is where it goes.

Full suite green (2306/0), `tsc` clean.

jcfischer (commented):

### Architecture

- **[important]** `src/work-graph.ts:352` — **Preserve GraphStore migration path**
  The exported seam removes `listCandidateFrontier(root): Promise<NodeRef[]>` and replaces it with `readSubtree(root): Promise<NodeState[]>`; existing external `GraphStore` implementations cannot upgrade compatibly. `CONTEXT.md` defines the work graph as exposed through a typed `GraphStore` seam, so this is a public contract break.
  Fix: Keep a deprecated compatibility method or ship this as an explicitly versioned breaking API change.

### Performance

- **[important]** `src/work-graph-github.ts:566` — **Parallelize truncated-state repairs**
  `"node.stateTruncated ? { ...(await this.readNode(node.state.ref)), parent }"` runs inside the depth-first loop, so every node exceeding 10 assignees or 20 blockers waits for a full multi-request `readNode` before the next repair starts. A broad subtree with truncated connections turns this into minutes of sequential `gh` calls.
  Fix: Collect repairs and execute them with bounded concurrency, then preserve preorder when assembling states.

### Maintainability

- **[suggestion]** `src/work-graph-github.ts:329` — **Centralize NodeState assembly**
  The new `state: { ref: { id: String(number) }, node, status: issue.status, assignees: issue.assignees, ... }` repeats the NodeState assembly in `readNode`. A NodeState change now has two production sites to update despite the nearby “Two parsers would be two answers” intent.
  Fix: Extract `toNodeState(issue, blockedBy, parent?)`; leave this function to decode GraphQL-specific fields and truncation.

### HonestOracle

- **[important]** `src/work-graph-github.ts:233` — **Benchmark claim assumes an extra round trip**
  The comment calls the alternative "two round trips always," but a traversal with no surviving frontier nodes performs no second hydration call. It then says "there is no map yet where it does not," extending the #495 measurement into an unsupported general performance claim for hydrating every traversed historical node.


---

jcfischer:
## Cycle 3 — three taken, one declined

`d4f1a2c`.

**Centralize NodeState assembly** — taken, and it was arguing against a comment I had written three lines above it. `toNodeState` is now the single production site; both read paths decode their own wire shape into a `GitHubIssue` plus blockers and hand off.

**"Two round trips always"** — wrong, taken. No survivors, no hydrating call, so the comparison is now stated per case. And **"there is no map yet where it does not"** turned one measurement into a performance law; the note now says what was actually measured and names the limit — this scales with *closed* history, which only grows, and nothing here establishes where the crossover is.

**Migration path** — you raised this twice, and the second half of your suggestion was simply right. A deprecated shim for zero implementors is what DD-16 clause 3 forbids, so I still decline that. But "ship it as an explicitly versioned breaking change" costs nothing and the repo already keeps a Keep-a-Changelog file claiming SemVer — not using it was an omission, not a decision. `CHANGELOG.md` now carries the break under Unreleased/Changed, marked BREAKING, with the obligations an external implementation takes on: live state, report-and-descend through closed nodes, no silent truncation, every node whole.

### Declining: parallelize the truncated-state repairs

The repair fires when a node has **more than 10 assignees or more than 20 blockers**. On a work-graph node that is not a rare case, it is a case that has never occurred — the busiest node on this map has one assignee and one blocker. "A broad subtree with truncated connections" is not a shape the graph produces.

It is also the same class as the sibling-fetch concurrency raised on #571, which was deferred to #530 and settled there: conditional cost on a path that does not fire. #530's decision was "do no obviously-wasted work", not "maximise concurrency" — and work that does not happen is not slow.

I did look at widening the pages so it fires even less, and the budget forbids it: per-node connections already put the query at **454 100 of 500 000**, so assignees cannot go much past 10 without narrowing a traversal level, which trades a never-taken path for a more-often-taken one.

What would change my mind is evidence it fires: any real graph where a node exceeds those widths. The two conditional paths — re-rooting and repair — sit next to each other and would parallelize together, so this stays cheap to do later and is recorded on #530 rather than dropped.

Full suite green (2306/0), `tsc` clean.

jcfischer (commented):

1 finding(s): 1 suggestion.


### Maintainability

- **[suggestion]** `test/cli-graph.test.ts:96` — **Deduplicate fake subtree walks**
  This fixture duplicates the recursive walk also added in `test/work-graph.test.ts`: `const seen = new Set<string>([root.id]);` and `states.push({ ...(await this.readNode({ id })), parent });`. Two copies of the new `readSubtree` semantics will drift when the seam changes.
  Fix: Extract a shared test helper for the depth-first confirmed-subtree walk, parameterized by child lookup.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### Architecture

- **[suggestion]** `docs/work-graph.md:345` — **Update frontier CLI contract**
  The amended contract says discovery “confirms,” but §2.6 still says `frontier` is “confirmed by direct fetch.” That leaves the public CLI contract inconsistent with the new GraphStore boundary.
  Fix: Replace “confirmed by direct fetch” with “confirmed by GraphStore.readSubtree”.

### Maintainability

- **[suggestion]** `src/work-graph.ts:322` — **Consolidate frontier semantics**
  `**This read confirms**` and `second read buys nothing and costs coherence` restate the same policy explained again by `**A pure filter over the subtree read**` and in the spec. Separate copies can drift.
  Fix: Keep the normative explanation in docs §2.4; shorten both JSDoc blocks to links.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `CHANGELOG.md:23` — **Benchmark is not reproducible**
  "Measured on the reference backend" reports exact 21-node timing and spawn-count results, but the diff contains no benchmark script, captured transport trace, or fixture for that live map. Release readers are asked to rely on a performance result they cannot independently verify.


---

jcfischer (commented):

3 finding(s): 2 important, 1 nit.


### CodeQuality

- **[important]** `src/work-graph-github.ts:584` — **Paged blocker repairs cannot parse**
  `node.stateTruncated ? { ...(await this.readNode(node.state.ref)), parent }` invokes `readNode` for a short `blockedBy` connection. That read uses `gh api --paginate`, which emits pages as separate JSON values, but the transport parses stdout as one JSON value; nodes with more than one REST page of blockers fail instead of being repaired.
  Fix: Use `--slurp` and flatten paginated arrays, or paginate this endpoint in the store.

### Maintainability

- **[nit]** `scripts/measure-graph-read-path.ts:42` — **Remove unused repo parameter**
  `function instrument(repo: string)` is immediately followed by `void repo;`, so callers pass a value the function deliberately discards.
  Fix: Remove `repo` from `instrument` and its call sites.

### HonestOracle

- **[important]** `src/work-graph-github.ts:200` — **Query budget reports the wrong cost**
  The comment says "at 15 the query scores 660 350" and bases the retained shape on "454 100 (91% of the ceiling)." GitHub rejects that exact 15-wide query as requesting 640,330 nodes, so the documented arithmetic is not the query budget maintainers are told to reopen when adding connections.


---

jcfischer:
## Cycle 4 — budget corrected (you were right), nit fixed, pagination finding refuted

`f2b8c31`.

### The budget was wrong, and worse than you said

You reported 640 330 against my 660 350. Sending the query settles it: **640 330**. The model was wrong for a reason worth keeping — `author{login}` is an object, not a connection, so it costs nothing. A node costs `1 + ASSIGNEE_PAGE + BLOCKER_PAGE` = **31**, not 32.

Which means the *current* shape was misreported too: **440 330**, not 454 100 — 88% of the ceiling, not 91%.

The corrected formula is `40 330 + 40 000w` for level-3 width `w`, and it now predicts GitHub exactly at both probed points:

```
w=12  →  refused, 520 330
w=15  →  refused, 640 330
```

So **11** is the widest that fits (480 330) and 10 is what we run, which the comment now says instead of implying 10 was forced.

Fourth correction to this budget. The comment now leads with the reproduction rather than the arithmetic — `MAX_NODE_LIMIT_EXCEEDED` states the score, so send the query. A model that has been wrong four times is a guess with a table around it.

**The unused `repo` parameter** — fixed, that was mine and it was silly.

### Refuting: paged blocker repairs cannot parse

Checked this one empirically because it would be a real pre-existing bug, and it does not reproduce. `gh api --paginate` **merges array pages into a single JSON array**:

```
$ gh api --paginate "repos/the-metafactory/soma/issues?per_page=2&state=all" | ...
type: list | items: 577
```

`per_page=2` and 577 items back, parsed by a single `JSON.parse`. If `gh` emitted one JSON value per page that call would have failed at the first page boundary — 288 boundaries went by.

The behaviour you are describing is real but belongs to *object* responses — `search/*`, or anything shaped `{total_count, items}` — where `--paginate` does emit a document per page and `--slurp` is the fix. `dependencies/blocked_by` returns a bare array, so it takes the merging path.

Worth noting the transport has used `--paginate` on this endpoint since #497 and the map has exercised it throughout, so a break here would not have been subtle. Happy to be shown a case where it does emit multiple documents for an array endpoint — that would be a genuine bug in the transport and not only in this path.

Full suite green (2306/0), `tsc` clean, frontier re-measured at 1 spawn.

jcfischer (commented):

3 finding(s): 3 important.


### CodeQuality

- **[important]** `docs/work-graph.md:250` — **Spec contradicts closed-node contract**
  The diff documents `readSubtree(root): Promise<NodeState[]>` as the confirmed subtree seam, but §2.4 still says the walk descends into closed nodes "while never reporting them." The implementation and `GraphStore` contract now require closed nodes to be returned, so external stores cannot follow both specifications.
  Fix: Update §2.4 to say the store reports closed nodes and `frontier` filters them.

### HonestOracle

- **[important]** `PR description:1` — **Description claims reverted work shipped**
  It says "findGraphRoot now takes it as a seed" and "addBlockingEdges shares one reading," but the diff says the close path is "Deliberately NOT seeded" and the shared ancestor walk was "reverted on review of #578." This misstates both the delivered savings and the resolution of #576's cited work items.

- **[important]** `PR description:1` — **Query budget remains wrong**
  The description claims "total 454 100 (91% of the 500 000 ceiling)," while the changed source explicitly corrects the same budget to "440 330 (88% of the ceiling)." This leaves reviewers relying on a calculation the PR itself says was disproved.


---

jcfischer:
## Cycle 5 — all three taken; the description was the problem

`8c4e0d9`, plus a rewritten PR description.

**§2.4 still told stores to hide closed nodes.** Real contradiction: "descends into closed nodes while never reporting them" was true of the candidate-ref seam and is the opposite of what `readSubtree` requires. An external store following both sentences could not exist. Fixed with the reason attached, since this is the second time the two drifted — filtering is the contract layer's job, and a store that pre-filters is deciding §2.4 rather than serving it.

**The description claimed reverted work shipped**, and **quoted the budget its own diff disproves.** Both correct, and both mine: I fixed the code across four rounds and never went back to the description, so it kept advertising the `findGraphRoot` seed and `addBlockingEdges` — the two things this PR now exists partly to explain *not* doing — and kept the 454 100 figure that the source corrects to 440 330.

Rewritten against the actual diff, with a note at the top saying it was rewritten and why. The reverts now have their own section, because "two of #530's four wasted reads were buying freshness" is the most interesting thing this PR found and it was invisible in a description that said they shipped.

Worth naming the pattern, since it is the third artifact-drift finding on this branch after §2.3 and §2.6: **the PR description is not a record of intent, it is a claim about the diff**, and it goes stale exactly like a comment does. I will re-read it against the diff before each review from here.

Full suite green (2306/0), `tsc` clean.

jcfischer (commented):

No findings. Sage approves.



---

Issue #580, opened 2026-08-08
The probe directory is stated, not inherited


jcfischer:
Implement #579: the probe directory becomes a **stated** value rather than an
ambient one, and the receipt names the tree it tested.

## What to build

1. **Resolve once, in the CLI.** `soma graph close` computes the probe directory
   and passes it to `defaultRunProbes(probes, { registry, cwd })`, instead of
   letting `runProbe` fall through to `process.cwd()`
   (`src/work-graph-probes.ts:205`, `src/cli/graph.ts:452`). The value is
   unchanged — the session's cwd — so no registry entry's meaning changes.

2. **Derive the evidence pointer from that same value.** `defaultEvidencePointer`
   (`src/cli/graph.ts:439`) runs a bare `git rev-parse HEAD`; it must run in the
   resolved probe directory. Today the pointer describes whatever tree the
   process happened to start in.

3. **The receipt names the probe tree.** Resolved absolute directory, its HEAD,
   and its dirty state (`git status --porcelain` non-empty), on both the probe
   lines and the derived `probed` evidence entry. `renderCloseReceipt`
   (`src/work-graph.ts:734`) is the render site.

4. **Dirty is recorded, never refused** — #579 decided this explicitly. Nothing
   in `assertClosable` gains a dirty-tree branch.

5. **Spec §2.2** gains the clause: a `command` probe runs in the directory the
   close was invoked from, resolved once and recorded in the receipt alongside
   that tree's HEAD and dirty state. Spec amendment, not a DD one — DD-16
   delegates mechanics to the spec and no normative clause moves (#530, #549).

## Watch for

- **Not a behaviour change on the happy path.** With `~/bin/soma`'s `cd` removed
  (done under #579), the resolved value already equals the session cwd. The point
  is that it becomes *stated and printed*, so a wrong tree is visible rather than
  silent. A test that only asserts "probes run in the session cwd" passes today
  and pins nothing.
- **The test that matters** is the one that would have caught this: run a close
  with a process cwd deliberately different from the invocation directory, and
  assert the receipt names the directory that was actually used. Absent that,
  this fix is unfalsifiable.
- **`git` calls need a `cwd`,** not a bare spawn — `runGit` already threads
  `-C` (`src/work-graph-probes.ts:186`); the evidence pointer does not.
- **The dirty check is a third git call per close.** #530 established the cost
  model is per-spawn (~600ms), so fold it into the HEAD read rather than adding
[...]

Issue #579, opened 2026-08-08
Probes run in the install tree, not the tree being closed


jcfischer:
`soma graph close` runs its `command` probes, and resolves its evidence pointer,
in **the tree the soma binary is installed from** — never the tree the work is
in. On any worktree-based session the receipt attests to different code than the
one being closed, and says nothing to indicate it.

## How it surfaced

Closing #576 from a git worktree at `dfea720` (the merge commit the node
delivered). The dry-run receipt reported:

```
- `probed` — 2/2 declared probes ran and passed (HEAD 6887d5030957840feb063aeffa2636b9d9658a82)
```

`6887d503` is not this worktree's HEAD, and **does not contain #576's change** —
it is an ancestor. So the receipt would have recorded a passing `bun test`
against a tree without the work, as evidence for closing the node that delivered
it. The gate accepted; nothing warned.

## Mechanism

`~/bin/soma` is:

```bash
cd "/Users/fischer/.claude/bin/soma" && exec bun run src/cli.ts "$@"
```

and that path is a symlink to the main checkout. So the CLI's process cwd is
always the install tree, whatever directory the session is in. Two consequences
follow:

- **`defaultEvidencePointer`** (`src/cli/graph.ts`) runs `git rev-parse HEAD`
  with no `cwd`, so it describes the install tree.
- **`command` probes** resolve their `cwd` against the same process cwd, so they
  *execute* in the install tree. The registry match then passes for the wrong
  reason: the install tree is declared, so the gate is satisfied by a directory
  the session never chose.

The second one is the serious half. This is not a mislabelled receipt — the
probe genuinely ran somewhere else. `bun test` passing in the install tree says
nothing about the branch under review, and if another session is editing that
tree (as one was here), the probe result is about *their* uncommitted work.

## Why this is the hollow-close class

DD-16 clause 1 wants gates that read facts the agent cannot author. A probe that
runs in an arbitrary other checkout still reads a fact — just not a fact about
the thing being closed. §2.2's "a probe that cannot run is a failed probe" was
written so an unrun probe can never read as passed; this is the same failure
with an extra step, since the probe *did* run and its result is still not
evidence.

It also quietly widened when the probe registry landed (#526). Registry entries
are keyed on resolved absolute `cwd` precisely so a declared command cannot be
run in an attacker-chosen directory — but the directory here is chosen by the
installer's symlink, not by the caller, and it is declared, so the gate sees
nothing wrong.

## What has to be decided

[...]

Issue #497, opened 2026-08-02
GraphStore seam + GitHub backend (typed contracts, spec §2)


jcfischer:
## Task

Implement the typed contracts from `docs/work-graph.md` §2 in soma core, on the work-graph integration branch (NOT merged to main alone — ships-with-consumer, see the merge ticket): `WorkGraphNode` (discriminated on autonomy, non-empty probes for `auto`, store-boundary validation), `Probe`/`ProbeResult` (typed, outcome pass|fail), blocking edges with DAG/cycle rejection, and the `GraphStore` interface with the GitHub backend (attestation capability, claim re-read + lexicographic tie-break, frontier direct-fetch confirmation, close receipt with evidence pointers).

Node: `kind: task` · `autonomy: auto` (branch work, below the irreversibility line) · probes (close evidence): `bun test` exit 0, `bunx tsc --noEmit` exit 0, branch pushed with the new module present.

jcfischer:
### Finding from #496 (2026-08-02): bind `addBlockingEdge` to GitHub's native relationship

The map's Notes claimed this tracker has no native blocking. It does — verified while resolving #496:

- `GET /repos/{owner}/{repo}/issues/{n}/sub_issues` and `.../dependencies/blocked_by` both respond (empty arrays, **not** 404).
- `POST .../dependencies/blocked_by` succeeded under the **Triage**-role agent token against an issue authored by the *principal*, and read back as `blocked_by #<n>` (probe 6).
- The same agent is **refused** when editing that issue's body (probe 7, negative control).

So under the provisioned identity, native edges are writable and body-text edges are not. §2.3's edge contract should bind to the native relationship rather than to a `Blocked by: #N` body line, and `listCandidateFrontier` can derive blocked-ness from `issue_dependencies_summary.blocked_by` instead of parsing prose.

Two things that follow:

- **Derived over declared** (§1 corollary) gets stronger here, not weaker — a native edge is structural tracker state; a body line is text any writer can desync from the truth.
- The `Child of #<map>` / `Blocked by: #N` lines on the current tickets are legacy. Decide during implementation whether to migrate them to native edges or leave them as readable residue; the map's Notes now say they stay readable until this ticket lands.

jcfischer:
### Acceptance criteria inherited from #502 (attestation derivation)

#502 filled spec §3.2 (PR #512). Four constraints land on this ticket:

[...]

PR #584, opened 2026-08-08
fix(work-graph): the probe directory is stated, not inherited (#580)

jcfischer:
Closes the implementation half of #579, as scoped by #580.

## The defect

`soma graph close` resolved the probe directory wherever it happened to need one — the runner fell through to `process.cwd()`, and the evidence pointer ran a bare `git rev-parse HEAD`. Under a launcher that `cd`s to the install tree, `bun test` executed in an unrelated checkout, passed against a commit that did not contain the work, and the receipt recorded only `HEAD <sha>`: a true fact about the wrong tree, with nothing to indicate it. The registry gate was satisfied for the same reason — the install tree *was* declared.

## The change

- **Resolved once, in the CLI.** A `probeCwd` dep produces the value; it is passed to the runner, and through it to the registry match. `ProbeRunnerOptions.cwd` is now required, so an ambient probe directory is no longer expressible.
- **The receipt names every tree it read.** `CloseReceipt.probeTrees` describes each distinct directory the declared probes actually resolve to — absolute dir, HEAD, dirty state, all as of *before* the run, since a probe may write to the tree it tests. Not the base directory for its own sake: an absolute `cwd` on a probe leaves the base entirely, and describing a tree nothing ran in is the same mislabel one indirection along. Each `ProbeResult` carries its own directory, and a probe line names it whenever more than one tree is in play.
- **Anchoring is all-or-nothing.** The `probed` evidence pointer names every tree, and is withheld if *any* of them has no readable HEAD — `n/n passed` beside a pointer that accounts for only some of the trees is the overstatement this PR is about. A `url`-only close records no tree at all and anchors on its targets instead, which is what a reader can re-check.
- **The pre-flight read is hardened.** HEAD and dirt come from one `git status --porcelain=v2 --branch` per tree (#530's per-spawn cost model), run concurrently. The directory is **tracker-supplied** and the read happens before the registry refuses anything, so it overrides `core.fsmonitor` and `core.pager` — `git status` would otherwise execute a program named by the target repo's own config. Published paths have the home prefix collapsed to `~`: a receipt goes to a tracker whose visibility soma cannot know, and the account name is the one part that does not help a reader tell one checkout from another.
- **Dirty is recorded, never refused** — #579 decided that explicitly; `assertClosable` gains no branch.
- Spec §2.2 gains the clause. DD-16 is untouched: it delegates mechanics to the spec and no normative clause moves (#530, #549).

**This is detection, not prevention** — worth stating plainly, since the title could be read as more. The stated directory still originates as the process cwd, so a launcher that `cd`s still moves it; what changes is that the substituted tree is named in the receipt instead of being silent. Refusing a probe tree that does not contain the work is the preventive version, and it needs to know which commit a node claims — #579 named it and scoped it out.

### What this changes about closing an `auto` node

Narrower than before, in one case: a probe that resolves to a directory with no readable HEAD now withholds `probed` evidence, where the old CLI-cwd `git rev-parse HEAD` could anchor that close with a sha from a tree the probe never used. No node in this repo is affected — every declared probe is a `command` in the checkout — and the alternative is anchoring evidence on an unrelated commit, which is the bug.

## Tests

The one that matters is the one that would have caught #579: a close whose stated probe directory deliberately differs from `process.cwd()`, asserting on the directory the runner actually spawned in *and* the directory the receipt names. A test that only asserts "probes run in the session cwd" passes on the broken code too — on the developer's machine the two were the same directory.

Also: a registry declaring another checkout no longer authorises; every tree described and each line naming its own; a base tree no probe used is never described or anchored on; one HEAD-less tree unanchors the set; a dirty tree closes and says so; the tree is read before the probes run; a relative `probeCwd` is resolved before it is recorded or compared; the status call carries its config overrides; home-collapsing keeps what distinguishes checkouts.

`bun test` 2322 pass / 0 fail, `bunx tsc --noEmit` clean — both in this worktree.


jcfischer (commented):

4 finding(s): 1 blocker, 3 important.


### CodeQuality

- **[important]** `src/cli/graph.ts:794` — **Receipt omits effective probe cwd**
  `const probeTree = probes.length > 0 ? await deps.describeProbeTree(probeDir) : undefined;` always records the base directory, while command probes execute with `cwd: resolvedCwd`. A declared command with its own absolute `cwd` can run in another checkout yet produce a receipt naming the session tree.
  Fix: Record probe-tree metadata per effective probe cwd, or explicitly restrict recorded probes to the base cwd.

- **[important]** `src/cli/graph.ts:794` — **Snapshot tree before probes run**
  `const probeResults = await deps.runProbes(probes, registry, probeDir);` precedes `await deps.describeProbeTree(probeDir)`. A declared probe can change HEAD or the worktree, causing the receipt to identify post-probe state rather than the revision the probe tested.
  Fix: Capture the tree status before running probes, or record before-and-after state explicitly.

### Performance

- **[important]** `src/cli/graph.ts:792` — **Avoid status spawn after failed probes**
  `"const probeTree = probes.length > 0 ? await deps.describeProbeTree(probeDir) : undefined;"` runs even when probes failed, while the diff states the per-spawn cost is `"~600ms"`. Failed closes cannot use this tree to create `probed` evidence, so they now incur a sequential ~600ms delay unnecessarily.
  Fix: Only describe the tree when rendering a dry-run receipt or when all probes passed.

### HonestOracle

- **[blocker]** `src/cli/graph.ts:501` — **Probe directory still inherits launcher CWD**
  The docs claim “The probe directory is stated, not inherited” and name a launcher that `cd`s as the failure being closed. Yet `probeCwd: () => resolve(process.cwd())` still takes the post-launcher ambient directory, so that launcher can run and authorise probes in the install tree while merely recording that wrong tree.


---

jcfischer:
## Round 1 addressed (`b7fd921`)

**Per-probe directory** *(CodeQuality, important)* — accepted, and it was the sharper version of the bug this PR exists to fix. A `command` probe carrying its own `cwd` (absolute especially) runs outside the base tree, and the receipt's single `probeTree` was speaking for it. `ProbeResult` now records the resolved directory each probe ran in — absent for `url`, which runs against a host, not a tree. The rendered probe line appends `[in <dir>]` when it differs from the base, and the `probed` evidence summary counts how many ran elsewhere so the pointer cannot imply the whole run happened where it points. Spec §2.2's `ProbeResult` shape gains the field.

**Snapshot before probes run** *(CodeQuality, important)* — accepted. The tree is now described before `runProbes`, so a probe that writes to the tree it tests (a fixture left behind, a stray build artefact) cannot make the receipt report dirt the probes themselves caused. New ordering test.

**Probe directory still inherits launcher CWD** *(HonestOracle, blocker)* — accepted as a **prose** defect, which is what it is: the code does what #580 scoped, and the docs claimed more than that. The value does still originate as `process.cwd()`, so a launcher that `cd`s still moves it — soma cannot know which tree you meant. What changed is that it is read once and *recorded*, so a substituted tree shows up in the receipt as a directory and HEAD that are not the ones under review, where before it was silent. §2.2 and the `probeCwd` dep doc now say "detection, not prevention" in those words, and name the preventive version — refuse a probe tree that does not contain the work, which needs to know which commit a node claims — as the thing #579 explicitly scoped out.

**Avoid status spawn after failed probes** *(Performance, important)* — **declined.** The fix above moves the read *in front of* the probes, so there is no post-failure branch left to skip. Keeping it unconditional is also what lets `--dry-run` render a refused close with its tree named, which is exactly when a reader needs to know which tree produced the failure. One ~600ms git call ahead of a run whose probes are declared with timeouts up to 900s is not a cost worth a blind receipt.

`bun test` 2316 pass / 0 fail, `bunx tsc --noEmit` clean.

jcfischer (commented):

### CodeQuality

- **[important]** `src/cli/graph.ts:776` — **Normalize injected probe directory**
  `const probeDir = deps.probeCwd();` preserves a relative dependency value, while the runner uses `const baseCwd = resolve(options.cwd);`. The command runs and authorizes against the absolute directory, but the receipt retains the relative one and falsely labels it as another tree.
  Fix: Use `const probeDir = resolve(deps.probeCwd());`.

### Maintainability

- **[suggestion]** `src/work-graph.ts:812` — **Share outside-tree predicate**
  The predicate "result.state === "probed" && result.cwd !== undefined" is duplicated in the summary and renderer, so their definitions of “elsewhere” can drift.
  Fix: Export `isOutsideProbeTree(result, tree)` and use it in both sites.

- **[suggestion]** `src/work-graph-probes.ts:244` — **Bind probe result context**
  Each branch repeats shared context such as "return probed(probe, "fail", authorization.reason, at, ranIn);" across the switch.
  Fix: Use a local `finish(outcome, observed)` closure that binds `probe`, `at`, and `ranIn`.

### HonestOracle

- **[important]** `src/work-graph.ts:805` — **URL-only receipt claims a tested tree**
  The renderer states "Ran in ${describeProbeTree(receipt.probeTree)}." whenever results exist, while the runner states "A `url` probe runs against a host, not a tree." `runClose` creates `probeTree` for any nonempty probe list, so a URL-only close records and displays a checkout as though it produced the host check.


---

jcfischer:
## Round 2 addressed (`f27f282`) — all four taken

**Normalize injected probe directory** *(important)* — accepted, and it was worse than a cosmetic mismatch: the runner resolves its cwd, so a relative value recorded on the receipt is a *different string* from the one every probe reports running in, and the new "did this run outside the tree?" comparison would then mark every probe as elsewhere. `resolve(deps.probeCwd())`, with a test that closes with `probeCwd: () => "."` and asserts the receipt records the absolute directory and flags nothing as elsewhere.

**URL-only receipt claims a tested tree** *(HonestOracle, important)* — accepted; the same mislabel pointing the other way. A close whose probes are all `url` now describes no tree at all — the git call is skipped, `probeTree` is absent, and nothing renders "Ran in …". It still has to anchor its `probed` evidence, and the honest anchor is the targets, since a reader re-runs the request; that also replaces the process HEAD it used to borrow, which was never evidence about a host check. §2.2 says so.

**Share outside-tree predicate** *(suggestion)* — accepted. `probeRanOutsideTree(result, tree)` is exported from `src/work-graph.ts` and used by both the evidence summary and the renderer.

**Bind probe result context** *(suggestion)* — accepted. `finish(outcome, observed)` in `runProbe` binds probe, timestamp, and directory once. Worth having beyond tidiness: the failure mode it removes is a branch silently omitting the new `cwd`.

`bun test` 2318 pass / 0 fail, `bunx tsc --noEmit` clean.

jcfischer (commented):

2 finding(s): 1 blocker, 1 nit.


### CodeQuality

- **[blocker]** `src/cli/graph.ts:817` — **Anchor actual probe trees**
  `"const probeTree = treeBound ? await deps.describeProbeTree(probeDir) : undefined;"` snapshots the base even when every non-URL probe has an absolute `cwd`/`repo`; `"const anchor = probeEvidencePointer(probes, probeTree);"` then permits an auto close using that unrelated HEAD. An actual probe tree with no HEAD is therefore bypassed rather than withholding `probed` evidence.
  Fix: Derive evidence only from trees actually used by probe results, or withhold `probed` when none ran in `probeTree`.

### Maintainability

- **[nit]** `src/work-graph.ts:334` — **Attach tree-rendering JSDoc**
  "How a probe tree reads in a receipt" is immediately followed by a second "/**" for `probeRanOutsideTree`, leaving the first block attached to no declaration.
  Fix: Move it immediately above `describeProbeTree`.


---

jcfischer:
## Round 3 addressed (`c904a09`) — both taken

**Anchor actual probe trees** *(blocker)* — accepted, and it was the same mislabel one indirection along: with every probe naming an absolute `cwd`, the base directory was described and used as the anchor while nothing had run there. Fixed by describing **every distinct directory the declared probes resolve to**, never the base for its own sake — `CloseReceipt.probeTree` becomes `probeTrees`. The directory set comes from `probeDirectory(probe, base)`, now exported from the runner, so the close path and the runner cannot hold two opinions about where a probe goes. Cost is one git spawn per distinct directory: one in practice, zero for a `url`-only close.

On the second half of your fix note — "an actual probe tree with no HEAD is bypassed": the pointer now names every tree and **withholds when any of them lacks a readable HEAD**, since `n/n passed` beside a pointer accounting for only some of the trees is the overstatement this PR exists to remove. That is a deliberate **tightening**: an `auto` node whose probe runs in a non-git directory no longer closes. No such node exists in this repo (every declared probe is a `command` in the checkout), and the alternative — anchoring on the trees that happen to have a HEAD — is the behaviour you flagged.

Three tests: every tree described and each probe line naming its own; a base tree no probe used is never described or anchored on; one HEAD-less tree unanchors the set.

**Attach tree-rendering JSDoc** *(nit)* — fixed; the block had been orphaned above `probeRanOutsideTree`. That predicate also changed meaning with the tree list — it now answers "does this line need to name its own directory?", which is what both callers were really asking.

`bun test` 2320 pass / 0 fail, `bunx tsc --noEmit` clean.

jcfischer (commented):

### Security

- **[blocker]** `src/cli/graph.ts:517` — **Preflight executes untrusted repo hooks**
  `"argv: [\"git\", \"-C\", dir, \"status\", \"--porcelain=v2\", \"--branch\"]"` runs before probe authorization. `git status` honors `core.fsmonitor`; tracker-supplied `cwd`/`repo` can point to a local attacker-controlled repository and execute its hook even when the registry later refuses the probe.
  Fix: Disable repository-controlled fsmonitor hooks for this status invocation and add a regression test.

- **[important]** `src/work-graph.ts:364` — **Absolute paths leak into tracker receipts**
  `"return `${head} in ${tree.dir} (${state})`;"` publishes the raw local directory in a receipt posted to the tracker. This exposes usernames and potentially sensitive customer or project names to every tracker reader.
  Fix: Use a redacted or opaque tree identifier in persisted receipts; reserve raw paths for local output.

### Performance

- **[suggestion]** `src/cli/graph.ts:831` — **Parallelize tree status probes**
  `for (const dir of probeDirs) { probeTrees.push(await deps.describeProbeTree(dir)); }` serializes independent status spawns before probes begin. The diff states each spawn costs “~600ms”, so three trees add about 1.8s of avoidable latency.
  Fix: Use `await Promise.all(probeDirs.map((dir) => deps.describeProbeTree(dir)))`.

### HonestOracle

- **[important]** `PR description:13` — **Auto-close semantics are not unchanged**
  “what closes an `auto` node is unchanged” is false when a probe resolves to another tree without a readable HEAD. The removed CLI-CWD `git rev-parse HEAD` could anchor that close; the new per-tree anchor withholds evidence and refuses it.

- **[suggestion]** `PR description:12` — **Spawn-cost claim is too broad**
  “One extra spawn, not two” omits that the new loop describes every distinct probe directory. A multi-tree close adds one `git status` spawn per tree, as the diff’s two-tree test exercises.

- **[suggestion]** `src/work-graph.ts:69` — **Rejected probes are described as run**
  “directory this probe ran in” is assigned before authorization in `src/work-graph-probes.ts:262`; a rejected command returns that cwd without spawning. A dry-run receipt can therefore name a tree a probe never ran in.


---

jcfischer:
## Round 4 addressed (`f2565f9`) — all six taken

**Preflight executes untrusted repo hooks** *(blocker)* — accepted, and it is a hazard this PR introduced: before it, no git command ran in a tracker-named directory. The status call now carries `-c core.fsmonitor=false -c core.pager=cat`, and `probeTreeStatusArgv(dir)` is exported so the overrides are a checked property rather than a comment. Regression test asserts the exact argv.

One adjacent thing I did **not** change, so it is not silently ignored: the `git-ref-exists` / `git-merged-into` / `artifact-exists` probes also run git in a `repo`-named directory and are *ungated* by the registry (§2.2's table says so deliberately). That is pre-existing and lower risk — `rev-parse`, `merge-base`, and `cat-file` do not consult `core.fsmonitor` — but it is the same shape, and belongs in its own node rather than smuggled into this one.

**Absolute paths leak into tracker receipts** *(important)* — accepted, with the narrower fix. An opaque identifier would defeat the point, since the receipt exists so a reader can tell one checkout from another — and below `~` the path is exactly what does that. So the **home prefix** is collapsed to `~` in everything published; `ProbeTree.dir` stays absolute internally, because it is what the runner compares against and a display convention must not quietly become a comparison one. §2.2 says so, and the test covers home-relative, outside-home, home-itself, and the `/Users/someone-else` near-miss.

**Parallelize tree status probes** *(suggestion)* — accepted; `Promise.all`. These are read-only reads of distinct directories, so the reason `runProbes` is sequential (probes share a tree and see each other's side effects) does not apply here.

**Auto-close semantics are not unchanged** *(important, HonestOracle)* — correct, and the claim was stale rather than wrong-at-the-time: round 3's all-or-nothing anchoring made it false. The PR description now has a section stating the narrowing plainly, including that no node in this repo is affected and why the alternative is worse.

**Spawn-cost claim is too broad** *(suggestion)* — same edit: one `git status` per distinct probe directory, run concurrently.

**Rejected probes are described as run** *(suggestion)* — accepted as a naming defect. The value is fixed before the gate, so a refused probe reports the directory it was refused *for* — which is the directory the adopter has to declare, so it is worth keeping. Field doc and spec now say "dispatched to — where it ran, or was refused for".

`bun test` 2322 pass / 0 fail, `bunx tsc --noEmit` clean.

jcfischer (commented):

4 finding(s): 4 important.


### CodeQuality

- **[important]** `src/work-graph.ts:389` — **Collapse Windows home paths**
  `return dir.startsWith(`${root}/`) ? ...` only recognizes slash-separated descendants. Windows resolves probe paths with backslashes, so published receipts retain the full home path instead of collapsing it.
  Fix: Compare paths using the platform separator, or normalize both paths before the boundary check.

### Architecture

- **[important]** `src/work-graph.ts:312` — **Keep probe-tree evidence out of work graph**
  The diff adds `export interface ProbeTree` with `HEAD` and ``git status`` state, then adds it to `CloseReceipt`; CONTEXT.md:696 says “The graph contributes topology + claims and no verification machinery — checkpoint keeps the verification monopoly.” Exporting this through `src/index.ts` makes Git-tree verification a permanent work-graph API contract for every backend.
  Fix: Move probe-tree snapshot and receipt-rendering types into the checkpoint/verification close-evidence boundary; keep GraphStore’s graph contract topology-focused.

### HonestOracle

- **[important]** `PR description:1` — **Unrelated frontier rewrite hidden**
  “Closes the implementation half of #579, as scoped by #580.” omits commit dfea720’s #576 frontier rewrite, including the `GraphStore.readSubtree` contract change. A reviewer relying on the stated scope can merge an independent behavioral change without reviewing it as such.

- **[important]** `docs/work-graph.md:132` — **Command cwd claim contradicts itself**
  “A `command` probe runs in the directory the close was invoked from” is false for command probes with their own `cwd`; the same clause immediately says a probe “may name its own `cwd`/`repo`.” This misstates where configured commands execute and what tree the receipt describes.


---

jcfischer:
## Round 5 (`5ed82de`) — two taken, two declined

**Collapse Windows home paths** *(important)* — accepted. `runCommand` has a `win32` branch, so a backslash-separated directory can reach the renderer, and a `/`-only boundary check would publish the full path on exactly the platform the collapsing protects. Compared on a normalised copy, returned with the caller's own separators; test covers `C:\Users\someone\work\x`.

**Command cwd claim contradicts itself** *(important)* — accepted, the clause did say two things. It now states the rule once: the close resolves one base directory, a probe naming no `cwd`/`repo` runs there, and one that does resolves against it — relative lands under it, absolute replaces it outright.

---

**Unrelated frontier rewrite hidden** *(important)* — **declined: the premise is wrong.** `dfea720` is the branch *point*, not part of this PR — it landed on `main` as #578 before the branch existed. The PR is five commits (`7b439a5`, `b7fd921`, `f27f282`, `c904a09`, `5ed82de`), and `git diff origin/main...HEAD` is 7 files / ~733 lines with **zero** occurrences of `readSubtree`. Nothing about the frontier or the `GraphStore` contract is in this diff. Worth flagging for the next round: reviewing the branch's history rather than its merge-base diff will keep re-reporting whatever `main` most recently merged.

**Keep probe-tree evidence out of work graph** *(important)* — **declined, as out of scope rather than wrong.** `CloseReceipt` already lives in `src/work-graph.ts` and already carries `probeResults`, `evidence`, `attestation`, and `attestationFacts`; `Probe` and `ProbeResult` are declared there too, and all of it is already exported through `src/index.ts`. That boundary was set by #497/#498 and shipped. `ProbeTree` adds one field to that existing shape — it introduces no new coupling, and relocating the close-evidence types to a checkpoint module would be a re-architecture of merged work.

The CONTEXT.md line you cite ("The graph contributes topology + claims and no verification machinery — checkpoint keeps the verification monopoly") is about **who decides**: `assertClosable` refuses a hollow close on the checkpoint's terms, and the graph does not invent a second gate. A receipt recording *which tree the probes ran in* changes nothing about that — it makes the existing evidence honest. If the boundary should move, that is a DD/spec decision, and this map's Notes are explicit that a ticket surfacing a genuine spec change goes back through `design/design-decisions.md` rather than being smuggled into an implementation node. Filing it as its own node is the right route; I am not doing it inside #580.

`bun test` 2322 pass / 0 fail, `bunx tsc --noEmit` clean.

jcfischer (commented):

5 finding(s): 2 important, 3 suggestion.


### CodeQuality

- **[important]** `src/cli/graph.ts:537` — **Handle status spawn failures**
  `"const status = await runCommand(...)"` only handles `"if (status.exitCode !== 0)"`; a missing or unspawnable `git` throws instead, aborting the close despite the adjacent `"git is unreachable"` fallback.
  Fix: Wrap `runCommand` in try/catch and return `{ dir }` on failure.

### Architecture

- **[suggestion]** `src/index.ts:752` — **Keep probeDirectory internal**
  The diff adds the public export `probeDirectory,` although its stated consumer is the close path’s internal preflight logic. This expands Soma’s public API and creates a compatibility commitment for a helper whose semantics are coupled to probe execution.
  Fix: Remove `probeDirectory` from `src/index.ts`; import it directly where internal code needs it.

### Maintainability

- **[suggestion]** `test/cli-graph.test.ts:792` — **Reuse the real-runner fixture**
  The new tests repeat `runProbes: async (probes, registry, cwd) => await runProbes(probes, { cwd, registry, deps: ... })` at lines 792, 829, 866, and 966 despite the existing `realRunner` helper.
  Fix: Extend `realRunner` for the shared command outcome and use it in these cases.

- **[suggestion]** `src/cli/graph.ts:833` — **Extract probe-tree preparation**
  `const probeDir = resolve(deps.probeCwd())` through `const probeTrees = await Promise.all(...)` combines resolution, deduplication, and preflight I/O inside already multi-phase `runClose`.
  Fix: Extract a `prepareProbeTrees` helper returning `{ probeDir, probeTrees }`.

### HonestOracle

- **[important]** `docs/work-graph.md:158` — **Pager threat claim is false**
  The claim that both overrides are needed because "git status would otherwise execute a program named by the target repo's own config" overstates the pager risk: the implementation explicitly states `core.pager` is "not reachable from --porcelain" in `src/cli/graph.ts:526`. This misdescribes the security boundary to readers relying on the documented hardening.


---

jcfischer:
## Round 6 addressed (`d01698b`) — all five taken

**Handle status spawn failures** *(important)* — accepted, and the fallback comment right beside it already claimed to cover this: `Bun.spawn` raises rather than exiting non-zero when the binary is missing, so `git` absent from PATH aborted the close *before its probes had run*. Now caught; the directory alone is the honest answer.

**Keep `probeDirectory` internal** *(suggestion)* — accepted. `src/cli/graph.ts` imports it from the runner directly, so the `src/index.ts` re-export bought nothing but a compatibility promise.

**Reuse the real-runner fixture** *(suggestion)* — accepted. `realRunner` gained an optional spawn collector, and all five new close tests go through it.

**Extract probe-tree preparation** *(suggestion)* — accepted: `prepareProbeTrees(probes, deps) → { probeDir, probeTrees }`, with the why-before-the-probes and why-every-directory reasoning moved onto it. `runClose` had enough phases already.

**Pager threat claim is false** *(important, HonestOracle)* — accepted, and it was the doc overstating what the code says. Both now say the same thing: `core.fsmonitor` is the reachable vector; `core.pager` is belt-and-braces because porcelain output is not paged, and it is pinned only because it is the other knob in `git status`'s reach that names a program.

`bun test` 2322 pass / 0 fail, `bunx tsc --noEmit` clean.

jcfischer (commented):

### Security

- **[important]** `src/cli/graph.ts:546` — **Bound preflight status concurrency**
  `const probeTrees = await Promise.all(probeDirs.map(async (dir) => await deps.describeProbeTree(dir)));` starts one `git status` per distinct tracker-supplied probe directory before the registry can refuse command probes. A large probe list can concurrently scan the same checkout through many subdirectories, exhausting processes, CPU, and file descriptors on the closing machine.
  Fix: Use bounded concurrency and reject or cap excessive distinct probe directories before spawning status reads.

### Architecture

- **[suggestion]** `src/index.ts:702` — **Keep receipt helpers internal**
  The public barrel now exports `describeProbeTree` and `probeRanOutsideTree`, although these are close-receipt implementation helpers. CONTEXT.md defines the work graph’s public seam as typed `GraphStore` and `soma graph` verbs; exporting helper semantics commits future receipt internals to compatibility.
  Fix: Do not export these helpers from `src/index.ts`; export only types required by public receipt contracts.

### HonestOracle

- **[important]** `src/cli/graph.ts:491` — **Mixed probes omit URL evidence**
  The comment claims, "the pointer covers the whole run," but when any tree exists lines 508-511 emit only tree descriptions; URL targets are added only when `trees.length === 0` (line 513). A mixed command-plus-URL close can report every probe passed while its evidence pointer omits the URL portion, misleading a reader about what is anchored or re-checkable.


---

jcfischer:
## Round 7 addressed (`HEAD`) — all three taken

**Bound preflight status concurrency** *(important)* — accepted, and it is the right correction to round 6's `Promise.all`: the directory list is tracker content, so the fan-out was node-body-sized rather than machine-sized. `mapBounded(dirs, 4, …)` caps in-flight reads, preserves order, and stays local — this is the only fan-out in the close path. Test declares twelve distinct probe directories and asserts the observed peak never exceeds the cap.

I did not add a hard cap on distinct directories. Refusing a close for declaring too many trees is a new refusal rule, and with the fan-out bounded the remaining cost is just time — the probes themselves already run sequentially, so a hundred-probe node is slow long before the pre-flight matters.

**Mixed probes omit URL evidence** *(important, HonestOracle)* — accepted; the comment claimed the pointer covers the whole run and the code only covered the trees. Both halves are always included now, so a command-plus-URL close cannot report `n/n passed` while quietly dropping the host checks from what a reader can re-check.

**Keep receipt helpers internal** *(suggestion)* — accepted. `describeProbeTree` and `probeRanOutsideTree` are off `src/index.ts`; they are internals of `renderCloseReceipt`, and the test imports them from `src/work-graph` directly.

`bun test` 2324 pass / 0 fail, `bunx tsc --noEmit` clean.

jcfischer (commented):

No findings. Sage approves.



---

Issue #582, opened 2026-08-08
Contain the ungated probes to the stated tree


jcfischer:
## What

Enforce the containment decided on
[#529](https://github.com/the-metafactory/soma/issues/529): the **resolved**
path of every ungated probe must stay under the runner's base cwd — the tree
[#580](https://github.com/the-metafactory/soma/issues/580) makes explicit.
Escape is a **failed probe**, never an exception and never a skip, so
`assertClosable` refuses through the path it already owns (#526's shape).

## The premise correction that shapes the fix

#529 asked whether **`repo`** is bounded. It is not the only escape.
`artifact-exists.path` resolves against the cwd too
(`src/work-graph-probes.ts:268`), so an **absolute `path` escapes with no `repo`
at all** — verified live at `dfea720`:

```
artifact-exists path:"/etc/passwd"                → pass | /etc/passwd exists
artifact-exists path:"/Users/…/.ssh/id_rsa"       → pass | /Users/…/.ssh/id_rsa exists
git-ref-exists  ref:HEAD repo:"../../.."          → pass | HEAD → 6887d503…
git-ref-exists  ref:main repo:"/Users/fischer/.soma" → pass | main → a4ca03e4…
```

So the check is on the **resolved directory or file the runner is about to
touch**, computed once and shared with the spawn — not a validation of the
`repo` field. Two resolutions per probe today (`probeCwd`, and
`artifact-exists`'s `resolve(cwd, probe.path)` in the no-`atRef` branch); both
must pass through it, and a containment check that only sees one of them is the
defect being fixed, not a fix.

## Shape

- One containment predicate, applied **before dispatch**, next to
  `authorizeProbe` — for the reason #526 put the gate there: a probe type added
  later must not inherit "contained" by omission.
- Contained means the resolved absolute path is the base tree or a descendant of
  it. Compare resolved absolute paths with a separator-aware prefix test, so
  `/base-evil` is not a descendant of `/base`.
- Refusal message names the resolved path and the base tree — the adopter has to
  be able to see *which* of the two it was, since the node's literal field and
  the resolved path differ (#526's `cwd` lesson).
- Nothing reads a node's autonomy class: uniformity stays structural, not
  asserted.

## Ordering against #580

#580 makes the base cwd stated rather than ambient. Containment against an
**ambient** base is containment against whatever directory the close happened to
run from — which is the hazard #579 found, not a bound. Land #580 first, or land
both together; this node inherits #580's resolved value rather than resolving
[...]

Issue #561, opened 2026-08-07
A required check that never reports: why did PR #555's head produce zero runs?


jcfischer:
## Problem

Making `gate / confidentiality-gate` and `Deterministic portability` required (#510) rests on an assumption that did not hold on this repo's most recent merge.

**PR #555** (`feat/plansteps-node-bridge`, merged `337e2b9`) has **zero check runs on its head commit** `1322b23`. `gh pr checks 555` reports *"no checks reported on the branch"*. The only Actions runs that ever fired on that branch were on an earlier sha `59aadaf3` — and **both failed**:

```
failure   confidentiality-gate   59aadaf3
failure   Portability            59aadaf3
```

So the last signal that branch produced was red, later pushes produced no signal at all, and it merged. It was admin-merged (recorded on #501), which is why nothing noticed.

## Why this matters now

A required check that never reports is not a gate — it is an indefinite block. Under the ruleset as of #510, any PR whose head produces no check runs cannot merge except by admin bypass. Admin bypass is retained at `always`, so the principal is never locked out; **`ivy-agent` at Write would be**, and it is the identity #510 exists to enable.

## The question

Why did pushes to `feat/plansteps-node-bridge` after `59aadaf3` produce no workflow runs? Candidates not yet distinguished: Actions auto-disabled after repeated failures, a billing/spending limit on the org, a `paths`/branch filter (neither workflow has one — `portability.yml` is bare `pull_request`), or runs pruned. Until this is diagnosed the required-checks rule has an unbounded false-block rate.

Scaffold of #510.

jcfischer:
## Resolution — the cause was exogenous: the GitHub Actions incident of 2026-08-06

Every repo-side candidate this node listed is ruled out by direct probe. What produced zero runs was a GitHub-wide Actions incident whose recovery mode **throttled webhook delivery**, and throttled webhooks are not backfilled.

### The repo-side candidates, each killed

| Candidate | Probe | Verdict |
|---|---|---|
| Actions auto-disabled after failures | `GET /repos/…/actions/permissions` → `{"enabled":true,"allowed_actions":"all"}`; `GET /actions/workflows` → all three `state: active` | ✗ |
| Billing / org spending limit | repo is **public** → free GitHub-hosted runners; org on `team`, `runs-on: ubuntu-latest` | ✗ |
[...]

Issue #529, opened 2026-08-04
Ungated argv probes: is `repo` really bounded to a local tree?


jcfischer:
## Question

DD-16 Amendment A ungates `git-ref-exists`, `git-merged-into` and
`artifact-exists` on the grounds that they are "argv, no shell, no egress,
bounded to existence checks in a local tree" (`docs/work-graph.md` §2.2). Sage's
review of #528 showed the last clause is **false as written**, and the amendment
never engaged the counter-case.

`repo` on those three probes is **tracker content**. The runner resolves it with
`probeCwd(probe.repo, baseCwd)` = `resolve(fallbackCwd, probeRepo)`
(`src/work-graph-probes.ts`), with no containment check — so `repo: "../../.."`,
or an absolute path, probes any directory on the closing machine. Two
consequences:

1. **The observed string is published.** `` `${probe.ref} does not resolve in
   ${cwd}` `` and `` `${probe.path} present at ${probe.atRef}` `` land in the
   close receipt on a possibly world-readable tracker. Directory existence and
   local paths leak — which is the *same disclosure risk* the amendment cites as
   the reason `url` probes are gated ("the receipt publishes the observed status
   to a possibly world-readable tracker"). The argument that gates one type
   applies to three it exempts.
2. **`git` is still an executable with a config surface.** `git -C <attacker
   path>` reads that tree's `.git/config`. Whether that reaches anything
   dangerous under `rev-parse` / `merge-base` / `cat-file` is exactly the kind of
   thing to establish rather than assume.

The amendment's own residual limit says a probe type that changes these
properties "would reopen this decision rather than inherit it". This is the
adjacent case: an *existing* type whose stated justification does not hold.

**To decide:**

- Does `repo` need containment (resolved path must stay under the runner's base
  cwd), or is it a declared value like `run`/`cwd` — i.e. does it belong in the
  probe registry too?
- Is directory-existence disclosure in a receipt acceptable, or does the
  `observed` string need bounding for these types?
- If containment lands, what happens to legitimate cross-tree probes — is
  `repo: "../sibling-checkout"` a real use case, and does it survive?
- What does §2.2's justification say afterwards, given the current wording is
  not true?

**Not in scope:** the `command`/`url` gate itself (settled, #524 → #526) and the
receipt's output *size* bounds (#527 — related but a different question; that one
is about volume, this one is about what may be observed at all).

Charted 2026-08-04 from sage's HonestOracle finding on PR #528.

[...]

PR #586, opened 2026-08-08
fix(work-graph): contain the ungated probes to the stated tree (#582)

jcfischer:
Implements [#582](https://github.com/the-metafactory/soma/issues/582), the containment half of [#529](https://github.com/the-metafactory/soma/issues/529), on the tree [#580](https://github.com/the-metafactory/soma/issues/580) made explicit.

## What was false

DD-16 Amendment A ungates `git-ref-exists`, `git-merged-into` and `artifact-exists` as *"argv, no shell, no egress, bounded to existence checks in a local tree"*. The last clause was false as written. `repo` and `path` are **tracker content** resolved against the runner's base cwd with no bound, so a node body could name any directory on the closing machine — and the receipt publishes the answer.

Verified live at `dfea720`, before this change:

```
artifact-exists path:"/etc/passwd"                   → pass | /etc/passwd exists
artifact-exists path:"~/.ssh/id_rsa"                 → pass | …/.ssh/id_rsa exists
git-ref-exists  ref:HEAD repo:"../../.."             → pass | HEAD → 6887d503…
git-ref-exists  ref:main repo:"/Users/fischer/.soma" → pass | main → a4ca03e4…
```

And after, at this branch's HEAD:

```
artifact-exists path:"/etc/passwd"                   → fail | probe refused — outside the probe tree …
artifact-exists path:"~/.ssh/id_rsa"                 → fail | probe refused — outside the probe tree …
git-ref-exists  ref:HEAD repo:"../../.."             → fail | probe refused — outside the probe tree …
git-ref-exists  ref:main repo:"/Users/fischer/.soma" → fail | probe refused — outside the probe tree …
artifact-exists path:"src/work-graph-probes.ts"      → pass | …/src/work-graph-probes.ts exists
git-ref-exists  ref:HEAD                             → pass | HEAD → dc9bcdf9…
```

**The premise correction:** #529 asked whether **`repo`** is bounded. It is not the only escape — `artifact-exists.path` resolves against the cwd too, so an absolute `path` escapes with **no `repo` at all**. A containment check that saw only the probe directory would have been the defect being fixed, not the fix.

## Shape

- **One predicate, before dispatch, beside `authorizeProbe`**, exhaustive with no `default` — for the reason #526 put the gate there: a probe type added later must be classified, not inherit "contained" by omission.
- **Contained** = the resolved absolute path is the stated probe tree or a descendant, by a separator-aware prefix test, so `/base-evil` is not a descendant of `/base`.
- **Both resolutions** pass through it: the probe directory, and `artifact-exists`'s `path` in the no-`atRef` branch. With `atRef` the path is a repository-relative object name for `git cat-file` and never touches the filesystem, so only the directory is contained there.
- **Escape is a failed probe** — never an exception, never a skip — so `assertClosable` refuses through the path it already owns. Same shape as a registry refusal, with its own prefix, because the two have different fixes.
- The refusal **names the resolved path and the tree**, home-collapsed: the node's literal field and the path the runner would touch are different strings (#526's `cwd` lesson).
- The **pre-flight tree read** skips uncontained directories before spawning `git status` in them — that read runs in a directory a node body names, ahead of any gate.
- **`command` and `url` stay uncontained**, deliberately: a `command`'s resolved `cwd` must already match an absolute directory the adopter declared in soma-home, and a `url` names a host and no tree. Containment on top would forbid the adopter's own deliberate act.
- Nothing reads a node's autonomy class — uniformity stays structural.

**Lexical, not `realpath`**: a symlink inside the tree pointing out of it still escapes. Resolving links would mean filesystem I/O in a predicate that runs before the runner has decided to touch anything, and would still race the probe. Recorded in the spec rather than implied away.

## Spec

`docs/work-graph.md` §2.2's gate row is rewritten to what is now true, plus a *Containment of the ungated probes* subsection. A **spec** amendment, not a DD one — DD-16 delegates mechanics to the spec and no normative clause moves (precedent: #549, #530, #579).

It carries #529's tested-negative honestly: the exemption rests on **containment**, not on a proof that `git`'s config surface is inert. Five exec-capable knobs were pointed at a marker script and none fired under the three verbs' exact argv on git 2.40.0 — a negative result on a tested set, which on this map has four times looked identical to a positive one (#557, #510, #561, #579).

## Tests

Each of the three types refused on a relative escape, on an absolute `repo`, and — for `artifact-exists` — on an absolute `path` with no `repo`; the sibling-prefix case refused; contained probes still pass, including a `repo` naming a subdirectory; the refusal is a `probed`/`fail` result and `assertClosable` refuses on it; the pre-flight never reads an uncontained tree.

`bun test` 2351 pass / 0 fail · `bunx tsc --noEmit` clean.

## Not in scope

A registry `repoRoots` list for cross-tree probes (#529 ruled it fog — no consumer, and #579 found the registry's existing worktree entries had never once been used), bounding the `observed` string (useless against a chosen-path oracle; containment removes the oracle instead), and output volume bounds (#527).

### DIFF
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -219,7 +219,10 @@ The rule, now enforced:
 - **Escape is a failed probe**, never an exception and never a skip — same shape
   as a registry refusal, refusing through the path `assertClosable` already owns.
   The message names the resolved path *and* the tree, since the node's literal
-  field and the resolved path are different strings.
+  field and the resolved path are different strings. `soma graph close` surfaces
+  it ahead of that generic failure in its **own** section, separate from the
+  registry's: the fix is the node, and pointing an adopter at their registry over
+  an escape would send them to widen a gate that was never the one refusing.
 - The pre-flight tree read (§2.2) skips uncontained directories **before**
   spawning in them: it runs in a directory a node body names and before any gate
   has refused anything.
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -183,7 +183,69 @@ depend on knowing who is trusted there.
 | --- | --- |
 | `command` | Refused unless the exact `run` **and** `cwd` pair is declared in the local probe registry for this repo. `cwd` is part of the match: a declared command run in an attacker-chosen directory is a different command. |
 | `url` | Refused unless the target host is in the declared host set. Ungated, this is a blind SSRF oracle — the request issues from the closing machine and the receipt publishes the observed status to a possibly world-readable tracker. |
-| `git-ref-exists`, `git-merged-into`, `artifact-exists` | Ungated — argv, no shell, no egress, bounded to existence checks in a local tree. |
+| `git-ref-exists`, `git-merged-into`, `artifact-exists` | Ungated by the registry — argv, no shell, no egress — and **contained to the stated probe tree** (below). |
+
+##### Containment of the ungated probes
+
+**Status: enforced**
+([#582](https://github.com/the-metafactory/soma/issues/582)), in
+`src/work-graph-probes.ts` beside the registry gate, before dispatch and with the
+same exhaustive switch — a probe type added later must not inherit "contained" by
+omission either.
+
+The row above previously read *"bounded to existence checks in a local tree"*.
+That clause was **false as written**, which is what
+[#529](https://github.com/the-metafactory/soma/issues/529) refuted: `repo` and
+`path` are tracker content resolved against the runner's base cwd with no bound,
+so a node body could name any directory on the closing machine. Verified live at
+`dfea720`: `artifact-exists path:"/etc/passwd"` passed, as did
+`path:"~/.ssh/id_rsa"` and `git-ref-exists repo:"../../.."`. The disclosure this
+creates is the *same* one cited as the reason `url` is gated — the receipt
+publishes the observed string to a possibly world-readable tracker — and it
+applies to a probe whose path the reader chose.
+
+The rule, now enforced:
+
+- **Contained** means the *resolved absolute* path is the probe tree (§2.2's
+  stated base cwd, [#580](https://github.com/the-metafactory/soma/issues/580)) or
+  a descendant of it. The comparison is a separator-aware prefix test, so
+  `/base-evil` is not a descendant of `/base`.
+- **Every path the probe touches**, not just its directory. Two resolutions
+  exist: `repo` → the probe directory, and `artifact-exists`'s `path` when no
+  `atRef` is given, which resolves against that directory. A check seeing only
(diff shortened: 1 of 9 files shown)


jcfischer:
## Review round 1 — local multi-lens (CodeQuality / Security / Architecture / Performance)

Sage's review consumer is **dormant** on this machine — `cortex-default.log` at the running daemon's last boot:

```
cortex: review consumer DORMANT for agent=sage … cortex MyelinRuntime subscriptions
disabled (G-1111 pending; tasks.code-review.* envelopes will not be claimed)
```

`pilot request-review` published fine (correlation `201d225f`) but nothing was subscribed to claim it, so this round was run locally instead. Recorded here rather than left implicit: it is a **self-review**, which is weaker evidence than an independent one, and the reason it was self-reviewed is infrastructure, not judgement.

### Findings, all fixed in `852465f`

**1 — major, correctness of the operator surface.** The close path already had a block whose whole purpose is that a gate refusal must not reach `assertClosable` as the generic *"ran and failed"*. It keyed on `isProbeRefusal`, which matches the **registry** prefix only — so a containment escape fell straight through it, and the refusal text naming the resolved path and the tree reached nobody. `isProbeEscape` is now its sibling, reported in its own section.

Two sections rather than one list, deliberately: the trailer *is* the fix, and telling an adopter to edit their probe registry over a path that escaped the tree sends them to widen a gate that was never the one refusing. An escape is fixed on the node; no declaration widens it.

**2 — major, duplication of a rule this module warns about.** The containment check and the runner's `artifact-exists` branch were each resolving `path` against the directory. Two expressions of one rule is the #579 shape (`src/work-graph-probes.ts` says so in its own comments); now one, `artifactFilePath`.

**3 — nit.** `ResolvedProbePath.field` was `string`; now `"cwd" | "repo" | "path"`.

**4 — nit.** The doc comment describing `resolvedProbePaths` had attached itself to the interface declared above it. Moved onto the function.

### Checked and found clean

- No path reaches the filesystem before containment: the runner gates before dispatch, and the pre-flight tree read filters uncontained directories before spawning. `probeKey` builds strings only.
- Case-insensitive filesystems fail **closed** — `/USERS/…` against base `/Users/…` refuses rather than admits.
- Windows: `sep` is platform-derived, and `resolve` handles drive-absolute `repo` values.
- Lint baseline unchanged (27 files with pre-existing errors on `main`; none of them are files this PR touches — `test/cli-graph.test.ts:80` is untouched by the diff).

`bun test` 2351 pass / 0 fail · `bunx tsc --noEmit` clean.

### Not fixed, recorded instead

Containment is **lexical**, not `realpath` — a symlink inside the tree pointing out of it still escapes. Documented in `docs/work-graph.md` §2.2 and in the code, rather than implied away.

PR #590, opened 2026-08-09
feat(work-graph): every close carries prose (#588)

### LINKED ISSUE #588: Every close carries prose: --resolution-file folded into the receipt
## What

Implement the resolution-comment decision from
[#556](https://github.com/the-metafactory/soma/issues/556): **every close carries
prose**, and `soma graph close` is the only thing that posts it.

## The rule

| Path | What satisfies it |
| --- | --- |
| `auto` close | `--resolution-file <path>` |
| HITL bare close | `--resolution-file <path>` |
| HITL `--propose` → `close --proposal-comment <id>` | the proposal body, already posted |

One exemption, with a named reason: the proposal body *is* the resolution, and
requiring a second prose comment on the ratification close would post the same
thing twice.

## Shape

- **Folded into the receipt comment**, not posted separately: one comment,
  resolution above, receipt below, one write. No orphan comment is possible on a
  refused close, and there is no window in which one half landed and the other
  did not — which is the property #556 says the two halves have.
- The gate is a **precondition check**, not a tracker scan, and the spec should
  say so plainly. With folding there is nothing on the tracker to scan at close
  time: what is checked is that prose was supplied (or that a ratified proposal
  carries it). It is a forcing function, not evidence — unlike every other
  hollow-close conjunct, which checks a fact the agent could not fake. #556
  decided this knowingly; the honesty is the point, and a spec that implies
  otherwise would be the hollow-verification shape DD-16 exists to refuse.
- `--dry-run` renders the folded comment, prose included, and writes nothing.
- `--propose` already takes `--body`/`--body-file`; combining it with
  `--resolution-file` is refused as redundant rather than silently preferring one.
- Empty or whitespace-only prose is refused the way `--propose` already refuses
  it (`src/cli/graph.ts:846`).

## Why this closes a real hole

`--body`/`--body-file` are read **only** on the `--propose` branch
(`src/cli/graph.ts:844-873`), and `parseClose` rejects them without it
(`:354`). `postComment` exists on `GraphStore` (`src/work-graph.ts:462`) and the
GitHub backend already uses it to post the receipt
(`src/work-graph-github.ts:769`) — the capability is there, unexposed. So closing
an `auto` node the doctrine's own way required `gh issue comment`, which
`SKILL.md` forbids. #582 and #580 both closed that way.

Once this lands, `SKILL.md`'s "never reach past the verbs" needs **no exception
written down** — the gap it would have excused is gone. [#533](https://github.com/the-metafactory/soma/issues/533)'s GitLab map
inherits the fix for free, since the prose rides the same `postComment` seam as
the receipt rather than a GitHub-specific step.

## Spec

`docs/work-graph.md` §5 (close receipt) gains the folded shape and the new
conjunct; `closing.md`'s "Recording the resolution" collapses from three steps to
two, since posting and closing become one act.

## Tests

- An `auto` close with no `--resolution-file` refuses, and the message names the
  flag.
- A HITL bare close with no `--resolution-file` refuses.
- A HITL close naming a ratified `--proposal-comment` closes without one.
- The posted comment carries prose above the receipt, in one write.
- `--dry-run` renders the prose and writes nothing.
- `--propose` with `--resolution-file` refuses.
- Empty prose refuses.

Node: `kind: implementation` · `autonomy: approve`.

jcfischer:
Implements [#588](https://github.com/the-metafactory/soma/issues/588), the decision from [#556](https://github.com/the-metafactory/soma/issues/556).

## The contradiction

Doctrine required a **resolution comment** on every node before it closes, and forbade reaching past the `soma graph` verbs to the tracker's own CLI. On the AFK path those rules contradicted each other: `--body`/`--body-file` fed `--propose` only (`src/cli/graph.ts:844-873`), and `parseClose` rejected them without it. So closing an `auto` node the documented way required `gh issue comment` — leaving the one artefact humans actually read as the one with no contract, no repo resolution and no backend abstraction. #582 and #580 both closed that way, and #556's own close was the third.

## The rule

| Path | What satisfies it |
| --- | --- |
| `auto` close | `--resolution-file <path>` |
| HITL **bare** close | `--resolution-file <path>` |
| HITL `--propose` → `close --proposal-comment <id>` | the proposal body, already posted |

One exemption, only where prose demonstrably exists. Exempting HITL wholesale would have let a `grilling` node — whose entire output *is* a decision — close with no human-readable half, while an `auto` node that merely ran `bun test` was refused for the same omission. The bare close is the normal single-operator route (§3.2), not an edge case.

## Folded, not posted beside

`CloseReceipt.resolution` renders above the receipt and the backend posts **one** comment from it. Posted separately, both orderings lose something — before the probes leaves an orphan resolution on a node whose close then refuses; after them leaves a receipt whose prose failed to post. Folded, neither state is reachable. Cost accepted: prose cannot react to probe output, because it is written first.

## Refused in two places, neither redundant

- **`assertClosable`** checks the receipt, so the rule holds for every consumer of the seam — which was #556's actual complaint, that the behaviour lived only in one CLI branch.
- **`soma graph close`** checks before the probes run, so a missing paragraph costs nothing instead of a 900-second `bun test`, and it can name the flag. The contract cannot: it knows nothing about a CLI.

## The honest part

The new conjunct is **a forcing function, not evidence**, and the spec and the code both say so in those words. Conjuncts 1–3 check facts a session could not fake — a checkpoint is attached, a probe ran, a pointer resolves. This one checks that *something was written*, and no machine can check that it says anything. A conjunct described as verification when it is a prompt to write something would be the self-declared verification DD-16 exists to refuse.

## Verified live

Against the node this PR implements, using the worktree CLI:

```
$ bun src/cli.ts graph close 588
Close refused: node 588 has no resolution.
Every close carries prose — the human-readable half, above the receipt in one comment (#556).
  soma graph close 588 --resolution-file <path>
Exempt only when a proposal already carries it: a close naming --proposal-comment reuses that body.
Nothing was written, and no probe ran.

$ bun src/cli.ts graph close 588 --resolution-file res.md --dry-run
close would be ACCEPTED
## Resolution
…prose…
## Close receipt
```

**A finding from doing that**: `soma` on PATH runs an *installed snapshot* (`~/.claude/bin/soma/src/cli.ts`), not the tree under development — so my first attempt closed #588 with no prose at all, through a binary that predates this change. Reopened, and worth naming rather than hiding: it is §1 clause 5 working as designed (enforcement lives in the installed binary), and it means this gate does not bind any session until the install is refreshed. Same shape as #580's receipt being written by the pre-#580 binary.

## Also

`--dry-run` renders the prose and writes nothing · `--propose` with `--resolution-file` is refused rather than silently preferring one · an empty resolution file is refused the way `--propose` already refuses an empty body.

Spec §3.0. The orienteer skill's `closing.md`, `WalkTheMap.md` and `SKILL.md` collapse "post the resolution, then close" into one act.

`bun test` 2363 pass / 0 fail · `bunx tsc --noEmit` clean.

### DIFF
--- a/docs/work-graph.md
+++ b/docs/work-graph.md
@@ -565,6 +565,55 @@ its effect.
 
 ## 3. Receipts by autonomy class
 
+### 3.0 Every close carries prose
+
+**Status: enforced**
+([#588](https://github.com/the-metafactory/soma/issues/588)), in
+`assertClosable` (contract) and `soma graph close` (surface).
+
+A close has two halves: the receipt, written for machines, and the **resolution**
+— why this node resolved the way it did, in prose. The second is the half a later
+reader actually reads. It was doctrine with no implementation until
+[#556](https://github.com/the-metafactory/soma/issues/556): `--body`/`--body-file`
+fed `--propose` only, so recording a resolution the documented way meant reaching
+past the verbs to `gh issue comment`, which made the one artefact humans consume
+the one artefact with no contract, no repo resolution and no backend abstraction.
+
+The rule:
+
+| Path | What satisfies it |
+| --- | --- |
+| `auto` close | `--resolution-file <path>` |
+| HITL **bare** close | `--resolution-file <path>` |
+| HITL `--propose` → `close --proposal-comment <id>` | the proposal body, already posted |
+
+One exemption, and only where prose demonstrably exists: the proposal body **is**
+the resolution, and requiring a second would post the same thing twice. A bare
+HITL close has no proposal — it is the normal single-operator route (§3.2) — so
+it carries prose like any other. Exempting HITL wholesale would let a `grilling`
+node whose entire output is a decision close with no human-readable half, while
+an `auto` node that merely ran `bun test` was refused for the same omission.
+
+**Folded into the receipt comment**, not posted beside it: `CloseReceipt.resolution`
+is rendered above the receipt, and the backend's `close` posts one comment from
+it. Posted separately, both orderings lose something — before the probes leaves
+an orphan resolution on a node whose close then refuses; after them leaves a
+receipt whose prose failed to post. Folded, neither state is reachable. The cost,
+accepted: prose cannot react to probe output, because it is written first.
+
+**This conjunct is a forcing function, not evidence.** The other three check
+facts a session could not fake — a checkpoint is attached, a probe ran, a pointer
+resolves. This one checks that *something was written*, and no machine can check
+that it says anything. Adopted knowingly. Stating it plainly is load-bearing: a
+conjunct described as verification when it is a prompt to write something is the
+self-declared verification DD-16 exists to refuse.
+
+The CLI refuses **before the probes run**, so a missing paragraph costs nothing
+rather than a 900-second `bun test`; the contract refuses on the receipt, so the
+rule holds for every consumer of the seam and not merely for the CLI. `--dry-run`
+renders the prose and writes nothing. `--propose` with `--resolution-file` is
+refused rather than silently preferring one.
+
 ### 3.1 AFK (`auto`)
 
 `soma graph close` runs the node's declared probes and requires **≥1
--- a/src/cli/graph.ts
+++ b/src/cli/graph.ts
@@ -921,7 +921,19 @@ async function runClose(
   // cost a re-run of nothing rather than a re-run of `bun test`.
   let resolution: string | undefined;
   if (parsed.options.resolutionFile !== undefined) {
-    resolution = (await deps.readTextFile(parsed.options.resolutionFile)).trim();
+    let raw: string;
+    try {
+      raw = await deps.readTextFile(parsed.options.resolutionFile);
+    } catch (error) {
+      // Named rather than raised raw: the flag is required on every close now, so
+      // a mistyped path is the most common way this verb will fail, and `ENOENT`
+      // with a stack is a worse answer than the path that was not there.
+      throw new SomaCliError(
+        `soma graph close --resolution-file ${parsed.options.resolutionFile} could not be read: ${error instanceof Error ? error.message : String(error)}`,
+        1,
+      );
+    }
+    resolution = raw.trim();
     if (resolution.length === 0) {
       throw new SomaCliError(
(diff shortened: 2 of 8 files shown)


jcfischer:
## Review round 1 — local multi-lens

Sage's review consumer is still dormant (`G-1111 pending` — `tasks.code-review.*` envelopes are never claimed), so this is a **self-review**, same as #586. Recorded plainly: it is weaker evidence than an independent round, and the reason is infrastructure rather than judgement. That makes three merged PRs on this map with no independent review.

### Findings, all fixed in this round

**1 — duplication of a rule, the #582 shape one type along.** `assertClosable` and `renderCloseReceipt` were each spelling *"is there prose here"* with their own `?.trim() ?? ""`. The drift that would eventually produce: a receipt rendering a paragraph the gate did not count, or a gate counting whitespace the receipt renders as an empty heading. Now one `resolutionText`, two readers.

**2 — test coverage aimed at the wrong layer.** #556's complaint was that resolution-posting lived in **one CLI branch and nowhere else**. Proving the new rule only through `soma graph close` would reproduce exactly that, so `WorkGraph.close` refusing a prose-less receipt is now pinned directly at the seam.

**3 — ergonomics of a now-mandatory flag.** A mistyped `--resolution-file` path surfaced as a raw `ENOENT` with a stack. Since the flag is required on every close, that is the most common way this verb will now fail; the path is named instead.

### Checked and found clean

- The contract layer really does enforce it — `WorkGraph.close` calls `assertClosable` before `store.close`, so the rule reaches every consumer of the seam and not just the CLI. That was the PR's central claim, and it is now also a test rather than an assertion.
- The CLI's early refusal and the core's receipt check cannot disagree: `attestationFacts.proposal` is set exactly when `--proposal-comment` was passed and its comment read.
- The exempt path renders no empty `## Resolution` heading.
- Lint baseline unchanged; none of the touched files carry errors (`test/cli-graph.test.ts:80` is pre-existing and outside this diff).

`bun test` 2365 pass / 0 fail · `bunx tsc --noEmit` clean · both CI checks green.

### Noted, not fixed

`assertClosable` takes the receipt from its caller, so a hand-built receipt could assert `attestationFacts.proposal` and skip the prose. That is the pre-existing trust model of the whole receipt type — the same is true of evidence pointers — and the CLI derives those facts rather than accepting them. Not a new hole; worth knowing it is not a defence either.

PR #595, opened 2026-08-10
fix(adapters): standalone-binary hook assets, without breaking pi (#531)

### LINKED ISSUE #531: Standalone soma binary + participant-grade install path (c-005 pack needs it)
## Need

The c-005 pilot pack installs a 14-skill Soma home on **no-terminal participant Macs** (fresh machines, no bun, no repo checkout). `soma install codex --apply --skills …` is exactly the right projection tool (skills + hooks), but `soma` today is a dev wrapper (`cd <repo> && bun run src/cli.ts`) — not distributable.

## Ask

1. A **standalone compiled binary** (`bun build --compile`, darwin-arm64 at minimum) that carries `soma install codex` without repo/bun.
2. Installable by a facilitator script (copy to a PATH dir is fine — no installer polish needed).

## Context

- Course pack + facilitator script: `jcfischer/ai-consulting` — supertag-course/courses/c-005-deine-persoenliche-ki/pack/ (script prefers a soma binary when present, falls back to direct copy into ~/.agents/skills as the pilot stopgap).
- Wayfinder map boundary: the course map references this work, doesn't own it (jcfischer/ai-consulting#1, Out of scope).
- Timeline pressure: Pilot #0 Session 0 in Q4 2026; Daniela Patient-#0 install this week uses the stopgap path.

jcfischer:
Closes #531.

Four commits. The first two give **codex and claude-code** a working `--apply` path from a `bun --compile` binary; the third repairs the collateral damage the second one caused to pi; the fourth is a review follow-up.

## Scope, precisely

**`soma install codex --apply` and `soma install claude-code --apply` work from a compiled binary. `soma install grok --apply` still does not** — it remains on the runtime read and fails with ENOENT under `/$bunfs`. That exclusion is deliberate and commented in `src/adapters/grok/hooks/assets.ts`: `grok-hook-verbs.mjs` is dual-use — `../adapter.ts` imports it as a real module for `GROK_PRE_TOOL_USE_VERB` *and* it ships verbatim to the user's machine. Bun resolves a specifier once, so a text import of the same path strips the named exports and breaks module load for anything that loads soma, hooks included. Untangling that means moving the verb constant into a `.ts` module both sides can source from — not a drive-by.

This still closes #531, whose Ask names `soma install codex` on no-terminal participant Macs (c-005 pilot). claude-code came along for free. Grok was never in scope.

## Why these ship together

`a5285e6` switched the codex + claude-code hook assets to `with { type: "text" }` so bun inlines them at bundle time — without it a compiled binary resolves `import.meta.url` to `/$bunfs/root` and dies with ENOENT before writing a single file.

That change also broke every pi session:

```
Failed to load extension "~/.pi/agent/extensions/soma-algorithm.ts":
__SOMA_CLAUDE_HOOK_EVENT_HANDLERS__ is not defined
```

pi loads extensions through **jiti**, not bun. jiti — like node — honours only `type: "json"` import attributes and silently **strips** the rest, so the target gets evaluated as a real ES module. `hook-runner.mjs` then died on its unreplaced projection placeholder before the extension could register anything.

Merging the two halves separately would put a knowingly-broken state on `main` in between, so they are one PR.

## How claude-code ended up in pi's import graph

Nothing pointed at it deliberately. The generated pi extension calls `checkSomaPolicy`, and the chain ran:

```
agent/extensions/soma-algorithm.ts
  -> src/policy-audit.ts
  -> src/projection-private-roots.ts
  -> src/install-spec-registry.ts     <- eagerly imports EVERY adapter's installer
  -> src/adapters/claude-code/install.ts
  -> src/adapters/claude-code/hooks.ts   <- with { type: "text" }
```

A policy check was pulling in six installers to read a `privateRoots` field that four of them define.

## The fix

Break the chain rather than revert the text imports, so the compiled-binary path stays intact.

- **`src/adapters/private-roots.ts`** (new) — leaf module owning the four substrate default homes and all five private-root builders. Imports only `node:*` and two types. Install specs import *from* it, never the reverse.
- **`src/projection-private-roots.ts`** — a `Record<InstallSubstrate, ...>` table instead of a registry walk. Keyed by substrate so a new substrate is a type error until it declares its roots; registry order preserved because callers compare positionally.
- **`test/pi-dev-extension-import-graph.test.ts`** (new) — walks the real generated extension import graph for any non-`json` import attribute. Includes a vacuity guard and a **negative control** seeded at `install-spec-registry` that must still flag `hooks.ts` — so the assertion cannot pass by going blind.

## Verification

- **jiti probe** (the exact call pi makes, `jiti.import(path, { default: true })`) on `~/.pi/agent/extensions/soma-algorithm.ts`: `FAILED __SOMA_CLAUDE_HOOK_EVENT_HANDLERS__ is not defined` → `OK`.
- **Differential probe** of the new private-roots table against a verbatim copy of the pre-fix registry walk: equivalent across 9 option shapes (default, `homeDir`, each of the 6 substrates, and an unregistered string). This caught a real diff — `anthropic-cowork` declared its builder inline with no substrate self-filter, so a naive extraction changed behaviour in 4 of those cases.
- **Compiled binary** writes the full codex projection — hooks, `hooks.json`, `config.toml`, `memories/soma/*`, rules, `AGENTS.md` — with the `.mjs` assets intact.
- `bun run typecheck` clean; `bun test` **2378 pass / 0 fail**; the touched files lint clean (the pre-existing `src/` lint errors are unchanged).

## Not covered

An actual interactive `pi` startup was not verified from the authoring session — the non-interactive probe blocked on a provider. The jiti probe exercises the identical loader call, so the defect is addressed, but a real pi start is the last mile.

### DIFF
--- a/src/adapters/anthropic-cowork.ts
+++ b/src/adapters/anthropic-cowork.ts
@@ -22,7 +22,9 @@ export const ANTHROPIC_COWORK_POLICY_PATH = "soma/policy.md";
 export const ANTHROPIC_COWORK_MEMORY_SNAPSHOT_PATH = "soma/memory-snapshot.md";
 export const ANTHROPIC_COWORK_ACTIVE_VSA_PATH = activeVsaProjectionPath("anthropic-cowork");
 export const ANTHROPIC_COWORK_CAPTURE_README_PATH = "capture/README.md";
-export const ANTHROPIC_COWORK_DEFAULT_HOME = ".anthropic-cowork";
+// Owned by the leaf `adapters/private-roots` module (which must not import an
+// adapter) and re-exported here so existing importers are unchanged.
+export { ANTHROPIC_COWORK_DEFAULT_HOME } from "./private-roots";
 export const ANTHROPIC_COWORK_SKILLS_ROOT_PATH = "skills";
 
 const COWORK_README_HEADING = "# Soma Anthropic Cowork Projection";
--- a/src/adapters/anthropic-cowork/install.ts
+++ b/src/adapters/anthropic-cowork/install.ts
@@ -1,9 +1,9 @@
-import { homedir } from "node:os";
 import { readFile } from "node:fs/promises";
 import { join, resolve } from "node:path";
 import { isEnoent } from "../../fs-errors";
 import { skillsLoaderUnder, type SubstrateInstallSpec, type UninstallContext } from "../../install-spec";
 import { projectVsaSkillBundleFiles } from "../../vsa-skill-installer";
+import { anthropicCoworkProjectionPrivateRoots } from "../private-roots";
 import {
   ANTHROPIC_COWORK_ACTIVE_VSA_PATH,
   ANTHROPIC_COWORK_ACTIVE_VSA_MARKER,
@@ -98,15 +98,7 @@ export const anthropicCoworkInstallSpec: SubstrateInstallSpec<"anthropic-cowork"
     destinationDir: (substrateHome) => resolve(substrateHome, "skills", "VSA"),
   },
   privateRoots: {
-    projection: (options) => {
-      const homeDir = resolve(options?.homeDir ?? homedir());
-      const substrateHome = resolve(options?.substrateHome ?? join(homeDir, ANTHROPIC_COWORK_DEFAULT_HOME));
-      return [
-        resolve(substrateHome, "soma"),
-        resolve(substrateHome, "capture"),
-        resolve(substrateHome, "skills/VSA"),
-      ];
-    },
+    projection: anthropicCoworkProjectionPrivateRoots,
   },
   uninstall: {
     kind: "implemented",
--- a/src/adapters/claude-code/hooks.ts
+++ b/src/adapters/claude-code/hooks.ts
@@ -1,10 +1,26 @@
-import { readFileSync } from "node:fs";
 import { chmod, mkdir, readFile, rm, stat, writeFile } from "node:fs/promises";
 import { dirname, resolve } from "node:path";
 import { resolveBunExecutable } from "../../bun-probe";
 import { isEnoent } from "../../fs-errors";
 import { renderFeedbackHookHelper } from "../shared/feedback-helper";
 import { isClaudeCodeInstallOptions } from "./install-options";
+
+// Hook assets are inlined at bundle time (`with { type: "text" }`) so the
+// compiled binary can project them; a runtime readFileSync against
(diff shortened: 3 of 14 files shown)


jcfischer (commented):

2 finding(s): 1 blocker, 1 suggestion.


### Maintainability

- **[suggestion]** `src/adapters/private-roots.ts:35` — **Deduplicate root-builder boilerplate**
  `"if (options.substrate !== undefined && options.substrate !== \"codex\") return []; const home = resolve(options.homeDir ?? homedir());"` is repeated in the Codex, Grok, and pi.dev builders. Future changes to substrate filtering or home resolution require synchronized edits.
  Fix: Extract a private helper that accepts the substrate, default home, and relative root segments.

### HonestOracle

- **[blocker]** `PR description:3` — **Compiled-install claim excludes Grok**
  It says “`soma install <substrate> --apply` work[s] from a `bun --compile` binary,” while the added Grok comment states “`soma install grok --apply` does not work from a compiled binary.” This misleads compiled-binary Grok users into a failing install path.


---

jcfischer (commented):

### EcosystemCompliance

- **[important]** `src/projection-private-roots.ts:34` — **Centralize private roots outside adapters**
  `"Record<InstallSubstrate, ...>"` and `"adding a substrate is a type error until it declares its private roots here"` make private-root facts a central table, while DD-4 assigns private projection roots to each adapter’s install facts. A new adapter can declare `privateRoots` correctly yet have policy handling omit them until this separate registry is updated.
  Fix: Expose a loader-safe private-roots leaf from each adapter and derive this lookup from adapter-owned exports.

### Maintainability

- **[suggestion]** `test/install.test.ts:36` — **Share copied-asset test loader**
  `const hookEntryDir = await mkdtemp(...)` plus `pathToFileURL(...)` duplicates the workaround in `policy.test.ts` (`const policyMarkerDir = await mkdtemp(...)` and `pathToFileURL(...)`). Future changes to the copied-module loading pattern must be kept in sync across both suites.
  Fix: Extract a test helper that copies an asset and imports it, with optional dependency assets.

### HonestOracle

- **[blocker]** `src/adapters/codex/hooks/assets.ts:7` — **No-Bun pilot still gets broken hooks**
  The comment says the pack “ships a compiled soma to people who have no bun and no repo,” but installation freezes `resolveBunExecutable()` into hook configs. With no Bun on PATH, the compiled Soma executable passes the shallow `--version` probe and is recorded as `bunPath`; later hook invocations run the Soma CLI with an `.mjs` path and fail as an unknown command, silently leaving the projected hooks nonfunctional.


---

Issue #618, opened 2026-08-12
Orienteer: a constraint that arrives after decisions close — reopen or supersede?


jcfischer:
## Question

A constraint arrives **after** decisions have already closed, and it invalidates
some of them. What happens?

Today: nothing. `soma graph <frontier|node|claim|add|close|audit|decisions>` has
no `reopen`, no `supersede`, no `invalidate`. Closing is one-way, and
`Workflows/WalkTheMap.md:74` ("Update or delete nodes the decision invalidated")
names an act with no verb behind it — and covers only invalidation by a
*decision*, never by a late constraint.

Raised by external feedback on 0.15.0 (see the companion prevention issue). The
user wanted to "rerun the whole thing with the new parameter". Better elicitation
prevents most of this class, but not all of it: a constraint can genuinely arrive
late, and the map needs an answer that is not "start over".

## Re-running everything is the wrong shape

Worth stating because it is the request that will keep arriving:

- **Most decisions are unaffected.** On a six-decision map, a budget constraint
  might invalidate one. Re-running six to fix one discards good work.
- **It burns the resource the constraint is usually about.** The user whose
  constraint was *no budget* was asking to spend six times the tokens.
- **It has no defined relationship to the receipts**, which are the record. A
  re-run either destroys them or forks a parallel history.

The narrow question — *which closed decisions does this constraint invalidate?* —
is already cheap to answer: `soma graph decisions <root>` projects every closed
node's gist, so reading them against a new constraint is a small, bounded
judgment. The gap is what to **do** with the ones that fail.

## The fork

**Branch A — reopen.** The affected node reopens, is re-resolved, and closes
again. A second receipt appends; `decisions --write` projects the latest.

- Keeps **one decision, one node** — load-bearing for the map-as-index, since
  `references/map.md` states a decision lives in exactly one place.
- Costs: a receipt is no longer final. The close gate's guarantee becomes "this
  was true when written" rather than "this is the answer". `decisions --write`
  must pick the latest receipt rather than the only one. Anything that read the
  first answer has no signal that it changed.

**Branch B — supersede.** The closed node stays closed with its receipt intact; a
new node supersedes it, linked both ways, and the decision index marks the
superseded line.

- Keeps **receipts immutable**, which is the one guarantee the close gate
  provides.
- Costs: two nodes now describe one decision, cutting against map-as-index. The
  superseding edge is a new edge type the DAG rules must account for. The index
  must render supersession without becoming a changelog.

[...]

Issue #602, opened 2026-08-11
Anti-criteria are write-only: declared by doctrine, unreachable by verify, invisible to show, ungated by advance


jcfischer:
Found while running an E1 close of map #495 through the harness.

## The gap

`RunAlgorithm.md` clause 7 is doctrine: *"Preserve anti-criteria with `Anti:` criteria when something must not happen."* The harness accepts them and then never reads them again.

Declared on `algorithm new`:

```
--criterion "Anti:No child node is reopened or edited to make the map closable"
```

`isAntiCriterion` routes it to `run.antiCriteria` (`src/cli/algorithm.ts:441`), and it lands correctly in the run file:

```json
"antiCriteria": [
  { "id": "Anti", "text": "No child node is reopened or edited to make the map closable", "status": "open" }
]
```

Then:

- **`algorithm verify` cannot reach it** — `--criterion-id Anti` → `Algorithm criterion not found: Anti`. The lookup searches `criteria` only, so an anti-criterion has no path from `open` to `passed`.
- **`algorithm show` does not render it.** The run above prints C1–C4 and nothing else, so an operator reading the run sees no trace of the constraint they declared.
- **`algorithm advance` does not gate on it.** The deterministic gate rejects missing verification for `criteria`; `antiCriteria` are not consulted, so a run advances to complete with every anti-criterion still `open`.

Every reference in the tree is a write or a passthrough:

```
src/algorithm.ts:132,166      construction + id-uniqueness validation
src/algorithm-store.ts:191    legacy migration passthrough
src/cli/algorithm.ts:394,442,450   the writers
src/types.ts:450,498          the type
```

No reader. The field is **write-only**.

## Why it matters

An anti-criterion is the one criterion type that exists to stop something, and it is the only one nothing checks. The failure is silent in both directions: declaring one produces no error, and never satisfying one produces no refusal — so a run that violated its own stated constraint completes indistinguishably from one that honoured it. Doctrine tells the operator to write them; the harness makes writing them a no-op.

Same family as #581 (invocation evidence recorded without executing the target) and #594 (thinking floors met by 0% of runs at E3+): **a declared floor with no enforcement path**, where the doctrine's confidence and the runtime's behaviour have quietly separated.

## What a fix has to decide

Not merely "wire it up" — an anti-criterion's verification is inverted, and that is a design question:

[...]

PR #619, opened 2026-08-13
docs(orienteer): give the map a home for a constraint, and make charting ask (#617)

### LINKED ISSUE #617: Orienteer: a map has no home for a constraint, and charting never asks for one
## The report

External feedback on soma 0.15.0, from Robert Chuvala:

> I installed soma 0.15.0 and woke up to some refreshingly thought out [maps].
> One issue I have with it is running an idea through it, seeing that I missed
> something that might change things (I have no budget to buy anything I am
> broke) and soma came back with BIG answers and I wanted to rerun the whole
> thing with the new parameter but that doesn't seem to be an option.
>
> […] it recommended I spend money but that's more on me. I worked around it,
> but I am used to reading results, seeing that I totally missed telling it a
> parameter my agents already know.

They charted a map, the map produced decisions that assumed a budget they do not
have, and there was no way to feed the missing parameter back in.

## Why this happened

**The map has no home for a constraint.** `references/map.md` gives the body five
sections — Destination, Notes, Decisions so far, Not yet specified, Out of scope
— and "I have no budget" is none of them:

- not **fog**: it is not a question awaiting an answer,
- not **out of scope**: it is not past the destination,
- not a **decision**: nobody decided it, it is simply true,
- so it lands in **Notes** as an unstructured "standing preference", which
  nothing reads and nothing checks a decision against.

**Charting never asks for one.** `Workflows/ChartTheMap.md` step 1 grills for the
destination and step 2 grills breadth-first for the frontier. There is no pass
that asks what is *fixed regardless of route*. The map did not mishandle Robert's
constraint — it never elicited it.

**The sharpest part of the report is "a parameter my agents already know."** Their
agents knew. The map did not. Orienteer charts from the conversation in front of
it and never consults the principal's own recorded constraints — which is a
strange gap for Soma specifically, since that is the compartment Soma exists to
hold. A principal's Purpose commitments and recorded challenges routinely encode
exactly this class of constraint (capacity, budget, time, non-negotiables). A
charting pass that read them would have raised budget without being told to.

## What to build

1. **A `## Constraints` section in the map body template** (`references/map.md`),
   sitting with Destination as part of what fixes the effort's shape: true
   regardless of which node you are on, and something every node's answer must
   respect. Distinct from Notes, which is orientation for the session; a
   constraint is a predicate on the *answers*.
2. **A constraint pass in `ChartTheMap.md`**, before or alongside naming the
   destination. Ask what is fixed: budget, time, tooling that cannot change,
   things that must not break, capabilities not available.
3. **Seed it from the principal's own store.** Charting should consult the
   Purpose compartment and the principal's recorded constraints rather than only
   the conversation, and propose what it finds for confirmation — never silently
   assume it.
4. **Nodes inherit the constraint.** A node body written under a constrained map
   should carry the constraints into its options, so a cost dimension is visible
   in the choices rather than discovered afterwards.

Robert was generous about the recommendation to spend money being "more on me".
It is not entirely — a map with a Constraints section would also have made cost
visible where the options are offered.

## Scope

This is the **prevention** half, and it is deliberately separated from the
late-arriving-constraint case, which needs a design decision first (see the
companion issue). Everything here is skill-and-template work: no new verb, no
runtime change, no change to the close gate.

jcfischer:
Closes #617.

## What happened

External feedback on 0.15.0 (Robert Chuvala): a charted map produced decisions
that assumed a budget the user did not have, and there was no way to feed the
missing parameter back in. The sharpest line in the report was *"a parameter my
agents already know"* — their agents knew; the map did not.

The map did not mishandle the constraint. It never elicited it, and had nowhere
to put it if it had: "no budget" is not fog (not a question awaiting an answer),
not out of scope (not past the destination), and not a decision (nobody decided
it) — so it lands in **Notes**, which nothing checks an answer against.

## What this changes

**`references/map.md`** — a `## Constraints` section in the body template,
sitting with Destination because the two together fix the effort's shape. The
test for what belongs there is whether an **answer could violate it**, which is
what separates it from the four sections around it. Notes orient *the session*;
a constraint is a predicate on the *answers*.

**`Workflows/ChartTheMap.md`** — step 1 now asks what is fixed regardless of
route, and seeds the question from the principal's own store rather than only
the conversation in front of it:

- `soma memory recall "budget capacity constraint deadline"` (read-only)
- `~/.soma/profile/purpose.md`, which carries this class under Commitments

Whatever it finds is **proposed for confirmation, never assumed** — a silently
assumed constraint is worse than one never elicited, because the map then
carries a predicate on every answer nobody agreed to. Where there is no Soma
home (orienteer runs in any repository) the pass degrades to the question,
asked.

**`Workflows/WalkTheMap.md`** — the map load reads Constraints and treats them as
binding on the answer being written; where a constraint touches the options,
that *dimension* is made visible in them (cost under a budget constraint) so it
applies where the choice is offered rather than after; a constraint the work
surfaces is recorded on the map, not buried in a receipt.

## Two decisions worth flagging

**It does not bind, and says so.** A Constraints section that nothing reads is
structurally identical to the Notes section it was created to escape — the
failure #602 names for anti-criteria. Rather than paper over that, `map.md`
states it outright: no verb reads the section, the close gate does not test
against it, `decisions --write` does not project it. It binds the way the
Destination binds — by being read — so the guidance is to write constraints a
reader can actually check an option against ("no budget for paid services" is
checkable, "keep it cheap" is not).

**Nodes do not inherit the constraint** (departing from #617's item 4).
Copying constraints into node bodies is the two-authoritative-homes problem the
map-as-index rule exists to prevent: amend a constraint and every copy goes
stale. Constraints live on the map and only on the map; what a node carries is
the dimension, not the text.

## Scope

Prevention only, and deliberately no runtime change: no new verb, no schema
change, no change to the close gate. The late-arriving-constraint case — what
happens to decisions already closed — stays with #618, which needs the
reopen-vs-supersede fork decided first.

## Verification

- `bun test test/install.test.ts` — 91 pass, 0 fail. No files added or removed,
  so the per-substrate projection counts (orienteer ×6) are unchanged.
- `bun run check-skill-version` — ok.
- Grepped for other places enumerating the map body sections
  (`"Destination, Notes"`, `"Destination and Notes"`): none outside the files
  changed here.

Not verified: this lands in live homes only after `soma install --apply`
reprojection — the change is to the source skill, not to any installed copy.

### DIFF
--- a/src/skills/orienteer/SKILL.md
+++ b/src/skills/orienteer/SKILL.md
@@ -103,7 +103,7 @@ Load only what the task routes to.
 | `Workflows/WalkTheMap.md` | Resolving a node on an existing map |
 | `references/map.md` | Writing the map body, or creating a node — anatomy, constraints, `autonomy` vs `kind`, the four work kinds |
 | `references/closing.md` | Closing a node — checkpoints, probes, the probe registry, AFK vs HITL receipts, attestation |
-| `references/fog.md` | Deciding whether something is fog, a node, or out of scope |
+| `references/fog.md` | Deciding whether something is fog, a node, or out of scope — or a constraint instead (`map.md`) |
 
 ## Lineage
 
--- a/src/skills/orienteer/SKILL.md
+++ b/src/skills/orienteer/SKILL.md
@@ -101,7 +101,7 @@ Load only what the task routes to.
 | --- | --- |
 | `Workflows/ChartTheMap.md` | Charting a new map from a loose idea |
 | `Workflows/WalkTheMap.md` | Resolving a node on an existing map |
-| `references/map.md` | Writing the map body, or creating a node — anatomy, `autonomy` vs `kind`, the four work kinds |
+| `references/map.md` | Writing the map body, or creating a node — anatomy, constraints, `autonomy` vs `kind`, the four work kinds |
 | `references/closing.md` | Closing a node — checkpoints, probes, the probe registry, AFK vs HITL receipts, attestation |
 | `references/fog.md` | Deciding whether something is fog, a node, or out of scope |
 
--- a/src/skills/orienteer/Workflows/ChartTheMap.md
+++ b/src/skills/orienteer/Workflows/ChartTheMap.md
@@ -3,7 +3,7 @@
 The user arrived with a loose idea. Charting is **one session's work and it
 hand-resolves nothing** — you end with a map and its first nodes, not answers.
 
-## 1. Name the destination
+## 1. Name the destination, and what constrains the way there
 
 Run a `/grilling` and `/domain-modeling` session to pin down what this map is
 finding its way to — the spec, decision, or change. One or two lines that a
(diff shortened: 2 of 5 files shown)


jcfischer:
## Review applied — findings 1–5 folded in (`11738d2`)

Reviewed this branch against a real Soma home rather than reading the diff, which is how finding 1 surfaced.

**1. The seeding pass leaked its source.** Running the prescribed `soma memory recall` on a live home returns `principal-challenges` — ADHD patterns, procrastination, therapy, "innere Grenzen", in German — none of it a constraint on any effort. `propose, never assume` did put a human in the loop, but it asked them to confirm a *constraint* while what was actually in front of them was a *disclosure*, headed for an issue body as public as its repo. Charting now translates before it asks: the constraint, never the source, and no recall output pasted into a map.

**2. Recall banners were ignored.** Results carry a verification banner with an age (the top hit was 38 days old, flagged `verify against …`). The text said nothing about it, so a weeks-old note could be proposed as a live constraint. Now: respect the banner, and vary the query — the hard-coded one is a starting point, and whole-file term scoring returns near-misses beside the hits (it also matched an unrelated game-dev note on "build budget").

**3. The router misrouted the very question this PR is about.** `SKILL.md` sent *"is this fog, a node, or out of scope?"* to `fog.md` — exactly the classification a constraint fails — while the answer had just moved to `map.md`. Fixed in the routing table, and `fog.md` now says plainly that what is true regardless of route is a constraint, not fog.

**4. `see below` pointed at nothing.** The template is copied into an issue, where the prose below it does not exist. It names the file now, like every other comment in that block.

**5. An unconstrained map had no defined shape.** `- none identified`, so a later session can tell an unconstrained effort from a question nobody asked.

### Cleared, not fixed

- Inserting a section above Decisions cannot break `decisions --write` — the splice is marker-based (`src/work-graph.ts:1020`), not positional.
- No files added or removed, so per-substrate projection counts are unchanged. `bun test test/install.test.ts` — 91 pass, 0 fail, re-run after the fixes.

### Split out

A sixth finding was real but pre-existing and off-subject: the map template ships without the `soma:decisions` markers, so **every** map fails its first `decisions --write`. Filed as #621 rather than widened into this PR.

PR #622, opened 2026-08-14
fix(orienteer): the map template ships the decisions markers (#621)

### LINKED ISSUE #621: Every map charted from the template fails its first `decisions --write` — the template ships without the markers
Found while reviewing #619. Pre-existing, unrelated to that PR's subject.

## The defect

`soma graph decisions <root> --write` refuses on any map charted from the
current body template, because the template does not contain the markers the
splice requires.

- `references/map.md:37` — the `## Decisions so far` section carries a prose
  comment and a placeholder line, and **no** `<!-- soma:decisions:begin -->` /
  `<!-- soma:decisions:end -->`.
- `src/work-graph.ts:1020` — `spliceSection` returns `undefined` when the
  markers are absent: *"the caller refuses rather than guessing where a
  decisions list belongs in prose it does not own."*
- `src/cli/graph.ts:1382` — `runDecisions` turns that into
  `Map <id> has no decisions markers — nothing to write into.`

So the verb is correct and its refusal is deliberate. The template is what is
wrong: it produces maps the verb cannot write to.

## Why it is worth fixing rather than living with

`decisions --write` is not an occasional verb — the skill's fast path and
`WalkTheMap.md` step 4 both run it after **every** close, so the failure lands
on the first node anyone ever closes, on every map.

The error message is actionable (it prints the two lines to add), which is
presumably why this has survived: each walker hand-patches their own map and
moves on. That is a per-map tax on a one-time template fix.

## The doctrine is mis-worded too

`Workflows/WalkTheMap.md:70` reads:

> Never hand-edit the span between the decisions markers; if the map predates
> them, add the two marker lines once and the verb owns the span from then on.

*"if the map predates them"* reads as legacy handling for old maps. It actually
describes **every** map, including one charted this morning — the markers have
no source that ever emits them.

## Fix

Put the two marker lines in the `## Decisions so far` section of the template in
`references/map.md`, so a charted map is writable by the verb that owns that
section. Then reword `WalkTheMap.md:70`, since after the template fix
*"predates them"* becomes true rather than misleading.

Worth deciding as part of it: whether `soma graph add` should inject the markers
when creating a root labelled `orienteer:map`, so a map is writable regardless of
which template its author copied. That is a runtime change and a larger question
than the template fix, which stands on its own.

jcfischer:
Closes #621.

## The defect

`soma graph decisions <root> --write` refused on **every** map charted from the
body template, because the template did not carry the markers the splice
requires.

- `references/map.md` — `## Decisions so far` had a prose comment and a
  placeholder line, and neither `<!-- soma:decisions:begin -->` nor its `end`.
- `src/work-graph.ts:1020` — `spliceSection` returns `undefined` without them,
  deliberately: *"the caller refuses rather than guessing where a decisions list
  belongs in prose it does not own."*
- `src/cli/graph.ts:1382` — that becomes `Map <id> has no decisions markers`.

The verb is right. The template was wrong, and it produced maps the verb could
not write to — on the first node anyone closed, on every map, since the fast
path and `WalkTheMap.md` step 4 both run `decisions --write` after every close.

It survived because the error prints the two lines to add, so each walker
hand-patched their own map and moved on: a per-map tax standing in for a
one-time template fix.

## The fix

**`references/map.md`** — both markers in `## Decisions so far`, with the
placeholder line *inside* the span so the first write clears it, and a comment
saying the verb owns that span.

**`Workflows/WalkTheMap.md`** — the old wording, *"if the map predates them, add
the two marker lines once"*, read as legacy handling for old maps when it
actually described every map, including one charted that morning. Now that the
template carries them, the sentence says what is genuinely left over: maps
charted before this, or from a hand-rolled body.

**`test/work-graph.test.ts`** — a regression guard that runs the real
`spliceSection` over the real template file, rather than asserting the marker
strings appear in it. The template and the splice are one contract living in two
files, and each half was correct in isolation — which is precisely why no
existing test could see the defect. Asserting them against each other is what
closes that gap.

## Not adopted

Injecting the markers at `soma graph add` time, floated in the issue. It would
have to append them to a body whose shape it does not know — the same guess
`spliceSection` refuses to make, and refuses for the same reason. A hand-rolled
map body stays the author's responsibility, with an actionable error when they
get it wrong.

## Verification

- `bun test test/work-graph.test.ts` — 40 pass, 0 fail, including the new guard.
  It fails by construction if the markers leave the template: `spliceSection`
  returns `undefined` and `expect(spliced).toBeDefined()` trips.
- Docs-only besides the test; no file added or removed, so per-substrate
  projection counts are unchanged.

### DIFF
--- a/src/skills/orienteer/Workflows/WalkTheMap.md
+++ b/src/skills/orienteer/Workflows/WalkTheMap.md
@@ -67,9 +67,11 @@ refusal is easier to avoid than to fix.
    let a map reader judge relevance without opening the node; the detail stays
    in the resolution.
 2. `soma graph decisions <root> --write` — the map's decision index re-derives
-   from receipts. Never hand-edit the span between the decisions markers; if the
-   map predates them, add the two marker lines once and the verb owns the span
-   from then on.
+   from receipts. Never hand-edit the span between the decisions markers. The
+   body template carries them, so a map charted from it is writable; a map
+   charted before they were in the template — or from a hand-rolled body — has
+   none, and the verb refuses rather than guessing where the index belongs. Its
+   error prints the two lines: add them once, and it owns the span from then on.
 
 For a HITL node wanting a second opinion, the two-phase `--propose` →
 ratification → `--proposal-comment` sequence stands, and the proposal body is the
--- a/src/skills/orienteer/references/map.md
+++ b/src/skills/orienteer/references/map.md
@@ -37,9 +37,13 @@ before choosing a node.>
 ## Decisions so far
 
 <!-- the index — one line per closed node: enough to judge relevance, then zoom
-     the link for the detail the node holds -->
+     the link for the detail the node holds. `soma graph decisions --write` owns
+     the span between the markers and refuses without them: keep both lines, and
+     never hand-edit between them -->
 
+<!-- soma:decisions:begin -->
 - [<closed node title>](link) — <one-line gist of the answer>
+<!-- soma:decisions:end -->
 
 ## Not yet specified
 
--- a/test/work-graph.test.ts
+++ b/test/work-graph.test.ts
@@ -25,6 +25,7 @@ import {
   type Probe,
   type WorkGraphNode,
 } from "../src/index";
+import { readFile } from "node:fs/promises";
 import { walkFakeSubtree } from "./fixtures/work-graph-fixtures";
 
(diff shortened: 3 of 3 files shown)


jcfischer (commented):

No findings. Sage approves.



---

PR #632, opened 2026-08-16
fix(release): configure Arc token at XDG path

jcfischer:
Fixes the v0.16.1 publishing workflow after Arc adopted its XDG config layout.

The workflow now writes and checks the prod source at Arc’s XDG config root under the runner home, .config/metafactory/arc.

Verified with an isolated current-Arc publish dry-run using a fixture token; it resolved @metafactory/soma v0.16.1 without the previous authentication failure.

### DIFF
--- a/.github/workflows/publish-soma.yml
+++ b/.github/workflows/publish-soma.yml
@@ -48,8 +48,13 @@ jobs:
             exit 1
           fi
 
-          mkdir -p ~/.config/metafactory
-          cat > ~/.config/metafactory/sources.yaml <<YAML
+          # Arc 0.37+ stores its configuration under the XDG app directory.
+          # Keep this in sync with Arc's default ~/.config/metafactory/arc root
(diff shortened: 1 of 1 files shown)


jcfischer (commented):

2 finding(s): 2 important.


### HonestOracle

- **[important]** `PR description:3` — **Dry-run claim lacks reproducible evidence**
  The description says, "Verified with an isolated current-Arc publish dry-run," but provides no command, Arc revision, or artifact. The workflow clones Arc’s moving default branch, so this does not establish the behavior CI will execute.

- **[important]** `.github/workflows/publish-soma.yml:53` — **Authentication claim uses a surrogate check**
  The comment claims the path is used "so the publish process reads the configured token," while the new pre-flight only runs `grep` against the written YAML. It never invokes Arc to prove that this Arc revision resolves that path or accepts the token.


---

PR #639, opened 2026-08-19
Skills canonical in ~/.soma: per-harness discovery, no duplicate catalog

### LINKED ISSUE #638: Skills canonical in ~/.soma; each projection implements the harness's own discovery instead of an eager catalog
## Problem

Skills should be canonical in `~/.soma/skills`, and each substrate projection should
implement *discovery* the way that substrate actually discovers skills. Today it works
the other way round: every adapter projects the same eager **text catalog**, and the
substrate's own discovery mechanism is either unused or duplicated.

On a real Claude Code home (`~/.claude`, 2026-08-19) this cost **~9,000 tokens of every
session** for zero capability:

- `rules/soma/SKILLS.md` — 36,702 bytes listing 115 skills
- Claude Code's own `Skill` tool already injected name + description for all 121 skills
  in `~/.claude/skills` — another ~12,300 tokens

Two lists, same skills, ~21.3k tokens/session. Worse, they **disagree on names**: the
catalog lists `red-team`, `systems-thinking`, `first-principles` (the `.soma` dir names),
while the Skill tool registers whatever `~/.claude/skills/<dir>` is called — `RedTeam`,
`SystemsThinking`, `FirstPrinciples`. Following the catalog produces an unresolvable
`Skill(...)` call. Deleting `SKILL.md` locally fixes it until the next
`soma install claude-code --apply` writes it back.

## Why the current shape produces this

Three mechanisms disagree about what projection means:

1. **`renderSkills`** (`src/adapters/shared/index.ts:226`) — the eager catalog. Emitted
   unconditionally by *every* adapter, and twice by some:
   - claude-code: `rules/soma/SKILLS.md` (`adapters/claude-code.ts:201`, builder `:230`)
   - cursor: `.cursor/rules/soma/SKILLS.md`
   - codex: `.codex/soma/skills.md` **and** `memories/soma/skills.md`
   - pi-dev: `.pi/extensions/soma-core/skills.md` **and** `agent/soma/skills.md`
   - anthropic-cowork: `soma/skills.md`
2. **`buildPortableSkillFiles`** (`adapters/shared/index.ts:96`) — returns
   `{path, content}`, i.e. it **copies** each skill's files into the substrate and
   rewrites paths through `rewriteSubstrateProjectionContent`. A copy drifts from its
   source on every edit to either side.
3. **`src/skill-projection.ts`** (#354 slice 1) — **symlinks** `~/.soma/skills/<Name>`
   into the substrate loader, explicitly so "edits to the source propagate with no
   re-sync". Its own header notes it "does not yet wire those callers".

So soma has a linking primitive that is right, a copying primitive that contradicts it,
and a catalog that duplicates whatever the harness already does natively.

Measured consequence of (2) on this machine: of 115 skills, **89 existed in
`~/.claude/skills` as independent copies rather than symlinks, and 50 of those pairs had
already drifted.** In every drifted pair the `.soma` side was newer; the substrate copies
carried stale PAI-era boilerplate. One (`push`) had lost its frontmatter entirely and so
never registered as a skill at all.

Separately, `soma install claude-code --dry-run` on this home plans exactly one
skill-related substrate file — `rules/soma/SKILLS.md` — and **no invocable
`skills/<name>/` directory at all**. That is ADR 0002's problem §1 ("No invocable-skill
projection") still live in the install path. (Possibly related: #612.)

## Proposed direction

Invert the default: **the canonical skill lives in `~/.soma/skills/<name>`; each adapter
declares how its harness discovers skills, and the catalog becomes the fallback, not the
rule.**

- Give `SubstrateInstallSpec` a skill-discovery mode, e.g.
  `skillDiscovery: { kind: "native-loader"; dir: string; link: true } | { kind: "catalog"; path: string }`.
- **native-loader**: symlink each canonical skill dir into the substrate's loader via the
  `src/skill-projection.ts` primitive, and **do not** emit `renderSkills`. The harness's
  own discovery is the catalog. Confirmed applicable: claude-code (auto-discovers
  `~/.claude/skills/*/SKILL.md`) and pi-dev (registers `skillPaths: [PI_SKILLS_HOME]`,
  `adapters/pi-dev/adapter.ts:373`).
- **catalog**: keep `renderSkills` where the harness has no loader and the catalog *is*
  the discovery mechanism rather than a duplicate of one.
- Which mode cursor, codex, grok and anthropic-cowork belong in is part of this issue —
  each needs its harness's actual discovery checked, not assumed.
- Retire the copying arm of `buildPortableSkillFiles`, or reduce it to the substrates that
  genuinely cannot follow a symlink.
- Whichever mode a substrate uses, the **name the harness registers must be the name the
  projection advertises**. Today they diverge silently.

## Acceptance

- [ ] `SubstrateInstallSpec` declares a skill-discovery mode per substrate; no adapter
      emits both a native loader projection and an eager catalog for the same skills.
- [ ] claude-code projects `~/.claude/skills/<name>` as symlinks to `~/.soma/skills/<name>`
      and stops writing `rules/soma/SKILLS.md`; `soma install --apply` does not recreate it.
- [ ] `soma uninstall claude-code` removes the projected symlinks and leaves
      `~/.soma/skills` untouched.
- [ ] `soma doctor` reports a substrate skill slot occupied by a real directory where a
      symlink is expected (the drift case above), rather than silently tolerating it.
- [ ] A skill edited in `~/.soma/skills/<name>` is live in the substrate with no
      reprojection step.
- [ ] Names advertised by the projection resolve in the harness's own invocation syntax.
- [ ] Substrates with no native loader keep the catalog, and a test asserts which mode
      each substrate is in.

## Notes

- Extends ADR 0002 (`docs/adr/0002-official-skill-collection-and-projection-primitive.md`),
  which proposed the symlink primitive but left the catalog unconditional.
- Adjacent: #386 (arc→soma projection delegation), #385 (promote official skills),
  #383 (customization overlay), #612 (zero bundled skills installed).
- The VSA installer (`vsa-skill-installer.ts`) is deliberately out of scope — per ADR 0002
  it owns drift-protected, edit-preserving projection that the symlink primitive cannot
  replace.
- Evidence for the numbers above: `~/work/skill-audit-2026-08-19.md` on the reporter's
  machine (Algorithm run `skill-audit-2026-08-19`).

jcfischer:
Closes #638 (except the doctor check — see *Not in this PR*).

`~/.soma/skills` becomes the curated set. Each substrate declares how its harness
discovers skills, and a substrate whose loader *is* that mechanism gets the whole
registry as symlinks instead of a text catalog listing names it cannot resolve.

## Why

Every adapter projected an eager catalog via `renderSkills`, regardless of whether
the harness already discovered skills on its own. On a real Claude Code home:

| | tokens/session |
|---|---:|
| `rules/soma/SKILLS.md` (115 skills) | ~9,000 |
| Claude Code's own Skill tool listing | ~12,300 |

Two lists, same skills. Worse, they disagreed on names: the catalog listed
`red-team`, `systems-thinking` (the `~/.soma` dir names) while the harness
registers the loader dir — `RedTeam`, `SystemsThinking`. Following the catalog
produced an unresolvable `Skill()` call.

Underneath that, three mechanisms disagreed about what projecting a skill means:
`renderSkills` listed them, `buildPortableSkillFiles` **copied** them with paths
rewritten, and `src/skill-projection.ts` **symlinked** them. On that same home, 89
skills existed as copies rather than symlinks and **50 had already drifted** — one
(`push`) had lost its frontmatter entirely and so never registered as a skill.

## What changed

**`SkillsDiscoveryMode` on `SubstrateInstallSpec`**, beside the existing
`skillsLoading`. The two are deliberately distinct: `skillsLoading` is *when a
discovered skill's body enters context*; `skillsDiscovery` is *whether the harness
advertises what it holds at all*.

- `loader` — the harness lists its own loader contents; Soma emits no catalog and
  projects the full registry into the loader.
- `catalog` — the projected list *is* discovery, not a duplicate of one; the
  loader stays opt-in via `--skills`.

Only **claude-code** is `loader`. The other five declare `catalog`, documenting
today's behaviour. An extra list is wasteful; a missing one is a capability
regression — so no substrate flips until its native discovery is verified. A test
pins all six, so flipping one is a diff, not a drift.

The catalog joins `ACTIVE_VSA` / `MEMORY` / `COMMUNICATION` as a conditional
projection file. Because `rules/soma` is an owned subtree, a `SKILLS.md` from an
older install is **reconciled away** rather than surviving as a stale
always-loaded copy — and the install plan no longer declares a file it never
writes, which was failing the greenfield "planned but never landed" check.

claude-code also stops emitting portable skill **copies** into its loader; the
manifest reconcile removes ones an earlier install left, so an existing home
migrates to symlinks on the next `--apply`. VSA keeps its dedicated
drift-protected installer (ADR 0002) and is excluded throughout.

## Two things a reviewer should push on

**The helpers take the discovery mode as a parameter instead of calling
`installSpecFor`.** Two reasons converged: spec modules import their adapter's
projection module for the rules-file list, so a registry lookup there closes an
import cycle — and inlining the check against a `const` lets TypeScript narrow it
to a literal, at which point ESLint correctly calls the condition dead code. The
indirection is what keeps the coupling live. If there's a cleaner way to break
that cycle, I'd take it.

**Wholesale projection is gated on `skillsDiscovery === "loader"`, not applied to
every substrate.** cursor and pi-dev project into *owned subtrees*, and
`reconcileOwnedSubtrees` would strip a symlink that isn't in the projected file
set — so wholesale linking there would silently delete itself on the next
install. Gating on `loader` keeps that invariant intact and makes the blast radius
exactly the substrate whose behaviour is verified here.

## Verification

- `bun test` — **2460 pass, 0 fail** (baseline 2438); 22 new tests (verified on head `c3c2946`)
- `tsc --noEmit` clean
- `eslint` — 62 errors, **byte-identical to `main`**; none added
- End-to-end on temp homes, not just units:
  - fresh `install claude-code --apply` writes no `SKILLS.md` and symlinks the whole registry
  - a planted stale `SKILLS.md` is reconciled away on reinstall
  - a home installed on `main` (real-dir copies) migrates to symlinks on reinstall
  - editing `~/.soma/skills/<name>/SKILL.md` is live in the loader with no reproject
  - `cursor` still gets its catalog and keeps its opt-in loader
  - `VSA` still projected by its own installer, not symlinked over

## Not in this PR

The #638 doctor check — "a loader slot holding a real directory where a symlink
belongs" — is deliberately deferred. It is now mostly *unreachable* for
claude-code: install no longer creates such directories and migrates the ones it
finds. What remains is a hand-made copy shadowing a registry skill, which is worth
a finding but is a different check from the one the issue described, and is better
written against the post-migration world than the pre-migration one.


src/skill-projection.ts
```diff
@@ -580,3 +585,36 @@ async function removeLink(linkPath: string, force: boolean): Promise<"removed" |
...
+    entries = await readdir(root, { withFileTypes: true });
+  } catch {
+    return [];
+  }
+  const dirs: string[] = [];
+  for (const entry of entries) {
+    // Symlinks count: a registry slot filled by `project-skill` from an external
+    // source dir is as canonical as one authored in place.
+    if (!entry.isDirectory() && !entry.isSymbolicLink()) continue;
+    if (entry.name === VSA_SKILL_NAME) continue;
```
jcfischer: **[blocker/correctness]** VSA is excluded on the dir name only, so a registry skill whose frontmatter says `name: VSA` slips through and breaks the whole install.

`linkSkill` names the loader slot from `readSkillName` (frontmatter first, dir basename as fallback), so a dir called anything with `name: VSA` inside resolves to the `VSA` slot — where the dedicated installer has already written a real directory. `ensureSymlink` then refuses it and the install aborts:

```
$ soma install claude-code --apply --home-dir $TMP
Refusing to replace non-symlink path at .../.claude/skills/VSA (not a Soma-created symlink). Pass force to overwrite it.
```

That is not a hypothetical: `projectableSkills` (adapters/shared/index.ts:64) guards this exact case with two keys and says why — *"matched on BOTH frontmatter name and canonical dir basename so a renamed SKILL.md cannot slip its dedicated projection back into the generic loop."* This function needs the same two-key guard: read the frontmatter name and skip when either it or the basename is `VSA`.

Note the blast radius is the whole install, not just the one skill — one odd registry entry and nothing projects.

src/cli/substrate-lifecycle.ts
```diff
@@ -475,22 +475,43 @@ export async function runSubstrateLifecycleCli(parsed: ParsedSubstrateLifecycleA
...
+ */
+async function installSkillDirs(
+  substrate: InstallSubstrate,
+  selected: string[],
+  options: SomaInstallOptions,
+): Promise<string[]> {
+  const somaHome = options.somaHome ?? defaultSomaHomePath(options.homeDir);
+  if (selected.length > 0) return selected.map((name) => resolveJoin(somaHome, "skills", name));
+  if (installSpecFor(substrate).skillsDiscovery !== "loader") return [];
+  return listRegistrySkillDirs(somaHome);
```
jcfischer: **[major/correctness]** `reproject` and `upgrade` never reach this — so the verb people use to resync a home does not resync skills.

`runSubstrateLifecycleCli` returns at the `reproject || upgrade` branch well before `installSkillDirs` is called, so only `install --apply` projects. Verified:

```
$ soma install claude-code --apply --home-dir $TMP   # ok
$ mkdir .soma/skills/NewSkill && ...                  # add a skill to the registry
$ soma reproject claude-code --home-dir $TMP
$ ls .claude/skills/NewSkill                          # absent
```

That was defensible when skills were an explicit `--skills` opt-in. Once the registry is the curated set that projects wholesale, "I added a skill to `~/.soma/skills`, now resync" is the main thing a user reprojects FOR, and it silently does nothing.

src/cli/substrate-lifecycle.ts
```diff
@@ -475,22 +475,43 @@ export async function runSubstrateLifecycleCli(parsed: ParsedSubstrateLifecycleA
...
+ * The skill source dirs this install should project.
+ *
+ * soma#638: `~/.soma/skills` is the curated set, so a substrate whose loader IS
+ * its discovery mechanism gets ALL of it by default — with no catalog to name an
+ * unprojected skill, anything left out of the loader is simply unreachable.
+ * `--skills` still narrows that to an explicit subset. A `catalog` substrate is
+ * unchanged: its catalog already advertises the whole registry, so its loader
+ * stays opt-in.
+ */
+async function installSkillDirs(
```
jcfischer: **[major/correctness]** The dry-run plan no longer describes what `--apply` will do.

The plan branch still only mentions skills when `--skills` was passed explicitly, so a plain `soma install claude-code` prints a plan that is silent about the N symlinks the apply will create. Verified: zero mentions of "Skills to project" on a home with 5 registry skills.

This matters more than usual here because the greenfield test in this very PR exists to enforce "the plan promises exactly what lands" — and the skill projection is now the largest thing `--apply` does that the plan does not name.

src/adapters/claude-code.ts
```diff
@@ -273,7 +307,16 @@ export function projectClaudeCodeHome(input: ProjectionInput): Projection {
   // kept), same shape as codex/grok. The `skills/` dir is SHARED (principal-authored + PAI-migrated
   // skills), so it is NOT an owned subtree; removals round-trip via the install
   // manifest (installClaudeCodeHomeProjection), not the owned-subtree reconcile.
-  const portableSkillFiles = buildPortableSkillFiles(input.profile.skills, input.bundledSkillNames, "claude-code");
+  // soma#638: a `loader` substrate takes its skills as symlinks to the canonical
+  // ~/.soma/skills entries (projected by install), not as copies. Emitting copies
+  // too would put a real directory in the slot the symlink needs, and — since a
+  // copy is rewritten for the substrate and then never resynced — would drift
```
jcfischer: **[major/consistency]** This is the dead-constant pattern the PR fixes elsewhere, left in place here.

`CLAUDE_CODE_SKILLS_DISCOVERY` is a `const` initialised to `"loader"`, so TypeScript narrows this to a literal comparison and the `buildPortableSkillFiles(...)` branch is unreachable. `projectedRulesFiles` was extracted into a function specifically so the same comparison would stay live — and its JSDoc argues the case. The two should be consistent: either route this through a helper that takes the mode as a parameter, or drop the ternary and call out that a loader substrate projects no copies.

As written a reader sees a live conditional where the compiler sees a constant, which is exactly the failure mode the other helper documents.

src/adapters/shared/index.ts
```diff
@@ -229,6 +231,51 @@ export function renderSkills(input: ProjectionInput): string {
...
+ * and because its rules dir is an owned subtree, a catalog left by a previous
+ * install is reconciled away rather than surviving as a stale second list.
+ *
+ * `discovery` is passed in rather than read from the install-spec registry: the
+ * spec modules import their adapter's projection module (for the rules-file
+ * list), so a registry lookup here would close an import cycle. Each adapter
+ * therefore exports one `SkillsDiscoveryMode` constant that both its projection
+ * and its install spec read — the same shape `CLAUDE_CODE_RULES_FILES` already
+ * uses.
+ */
```
jcfischer: **[nit/quality]** Two JSDoc blocks stacked here — the first one belongs to `skillCatalogFile`, which now has none.

The `skillCatalogFile` doc ("The eager skill catalog for one substrate...", including the import-cycle rationale) sits immediately above `projectedRulesFiles`, so tooling attributes it to the wrong function and the actual `skillCatalogFile` below is undocumented. Move it back down.

Same block: `Inlining it against a \\`const\\` mode` has literal backslash-escaped backticks that will render verbatim.

src/skill-projection.ts
```diff
@@ -580,3 +585,36 @@ async function removeLink(linkPath: string, force: boolean): Promise<"removed" |
...
+    const dir = join(root, entry.name);
+    if (!(await stat(join(dir, SKILL_MD)).catch(() => undefined))) continue;
```
jcfischer: **[nit/duplication]** `stat(...).catch(() => undefined)` re-implements `pathExists` from `src/fs-utils.ts:7`, which is already used by `adapters/skill-stub-doctor.ts` for the same kind of check. Use it — the truthiness dance on a `Stats|undefined` is also less obvious than a boolean.

Minor and adjacent: the `stat` per entry is awaited serially. Harmless at 5 skills, less so at the ~115 a real registry holds, and the repo does 4-wide concurrency for comparable scans elsewhere.

jcfischer (commented):
Lenses applied: CodeQuality, Architecture, Security (path/symlink handling), Performance, Duplication. 6 findings.

Self-review of my own PR, so weighted toward looking for what I got wrong rather than confirming what I got right. The direction holds up — the two-list problem is genuinely fixed and the migration path from real-dir copies to symlinks works on a home installed from `main`. The findings are all in the second commit, where wholesale projection went in with less scrutiny than the discovery-mode work in the first.

**Blocker**

- `listRegistrySkillDirs` excludes VSA on the dir name only. A registry skill whose frontmatter says `name: VSA` resolves to the `VSA` loader slot, hits the dedicated installer's real directory, and aborts the *entire* install. `projectableSkills` guards this same case with two keys and documents why; this function needs the same.

**Major**

- `reproject` / `upgrade` return before skill projection, so the verb for resyncing a home does not resync skills. Defensible when skills were `--skills` opt-in; wrong now that the registry projects wholesale.
- The dry-run plan is silent about the symlinks `--apply` will create — awkward in a PR whose own greenfield test enforces "the plan promises what lands".
- `CLAUDE_CODE_SKILLS_DISCOVERY === "loader"` in `projectClaudeCodeHome` is the dead-constant pattern `projectedRulesFiles` was extracted to avoid, left in place one file over.

**Nits**

- Orphaned JSDoc: `skillCatalogFile`'s doc sits above `projectedRulesFiles`; escaped backticks leaked into the same block.
- `stat().catch()` re-implements `pathExists` from `src/fs-utils.ts`; the scan is also serial where the repo does 4-wide elsewhere.

**What I checked and did not find fault with**

Symlink target resolution and the `assertSafeSkillName` guard (no traversal via a crafted registry dir name); the owned-subtree reasoning for gating wholesale projection on `loader` (cursor/pi-dev would have their symlinks reconciled away, so the gate is load-bearing, not timidity); manifest reconcile actually removing pre-existing copies; VSA's dedicated installer surviving intact.

verdict: blockers=1 majors=3 nits=2 — recommend: request-changes

(posted as comment-review — GitHub blocks request-changes on your own PR)

jcfischer:
## Re-review — head `3a5b075`

Three passes now. Each round found defects the previous round's *fix* introduced, which is worth stating plainly rather than presenting this as a clean convergence.

**Round 2 — all six original findings fixed.** Blocker (two-key VSA guard), reproject gap, silent plan, dead-constant ternary, orphaned JSDoc, `pathExists`. Verified by probe, covered by four regression tests.

**Round 3 — two defects introduced by the round-2 fix.**

- Adding the frontmatter-name check meant `readSkillName` ran during the scan, and it *throws* on an unsafe name — so one malformed registry skill aborted the run, including the **read-only plan**. A plan describes the registry; it must not die on account of contents it is only describing. Same shape as the blocker it was fixing.
- The plan printed registry **dir** basenames while the loader slot is named from **frontmatter**, so a dir `dir-name` holding `name: FrontmatterName` was announced as one and created as the other — breaking the promise-what-lands contract that fix cited as its own rationale.

Both fixed: `listRegistrySkillDirs` → `scanRegistrySkills`, which returns each resolved slot name and collects malformed entries into `unprojectable` instead of throwing. Callers render them under "Skipped (not projectable)" — silently dropping a curated skill is the failure this whole feature exists to prevent.

**Round 4 — two more, one a genuine regression from this PR.**

- **`uninstall` no longer round-tripped the skills it projected.** Skills used to reach the loader as copies recorded in the portable-skill manifest, which uninstall consumed. Symlinks are created by the projection primitive, outside any manifest — so uninstall had nothing to consume and orphaned them. This is #638's own AC3, and it was broken *by* this PR. `removeProjectedSkillLinks` now identifies them by target (a loader symlink pointing into `~/.soma/skills` is unambiguously ours, true even with a missing or stale manifest); real dirs and foreign symlinks are left alone; the registry is never touched. Verified a principal-authored skill sharing the loader dir survives untouched.
- Two registry dirs declaring the same frontmatter name resolved to one slot, both linked, last write winning — silently leaving one curated skill out of the harness. First by dir order keeps the slot, the rest are reported.

**Probed and found correct:** `--skills` with a typo still errors loudly rather than being silently skipped by the scan (deliberate — the principal named that dir); a SKILL.md with no `name:` falls back to the dir basename; a catalog substrate's plan stays silent about skills it will not project; VSA's dedicated installer keeps its slot throughout.

**Verification:** 2457 pass / 0 fail (baseline 2438), 19 new tests. `tsc --noEmit` clean. `eslint` 62 errors, byte-identical to `main`.

I'd want a second pair of eyes before merge. Three rounds of self-review each finding fresh defects in the previous fix is a signal about how much of this surface is load-bearing, not a sign it has converged.

jcfischer:
## Re-review round 4 — head `c3c2946`

Two findings, plus one I caused mid-review.

**Q3 — a dangling registry symlink vanished silently.** No "Skills to project" line, no "Skipped" line, nothing. A plain dir without a `SKILL.md` staying quiet is right (scratch folders are not skills), but a symlink whose target is gone was deliberately pointed somewhere and now resolves to nothing — silence leaves a curated entry that never reaches the harness with no signal why. Now reported; plain dirs still quiet.

**Q4 — `install grok --skills X` orphaned its symlink on uninstall.** Confirmed **pre-existing on `main`**, so not a regression from this PR, but the same defect class as the claude-code one fixed last round and grok has the same SHARED skills dir. Fixed rather than left as a known orphan.

**The import cycle that fix walked into.** Wiring grok to `removeProjectedSkillLinks` closed `grok/install.ts → skill-projection.ts → install-spec-registry.ts → grok/install.ts`, which TDZ-errors at module init:

```
ReferenceError: Cannot access 'grokInstallSpec' before initialization
  at src/install-spec-registry.ts:16:9
```

`claude-code/install.ts` had the same import and worked **only because the registry initialises it first** — latent, not sound. The helper now lives in `src/projected-skill-links.ts`, which imports nothing from the registry, and takes `loaderDir` rather than a substrate; a spec already computes its own via `skillsLoaderUnder()`, so the lookup was never needed. No install spec imports `skill-projection` any more.

Worth noting where that landed: because the Claude Code policy guard loads this worktree, the broken import took the editing session down and needed a human to run `git checkout`. That is #640, reproducing inside the PR that prompted me to file it.

**Probed and correct this round:** the `sep` guard holds — a loader symlink into `~/.soma/skills-backup/` survives uninstall while one into `~/.soma/skills/` is removed; repeat `--apply` is idempotent (`unchanged`, not `replaced`); principal-authored real dirs in the loader survive uninstall; `--skills` with a typo still errors loudly instead of being silently skipped by the scan; a `SKILL.md` with no `name:` falls back to the dir basename.

**Verification:** 2460 pass / 0 fail (baseline 2438), 22 new tests. `tsc --noEmit` clean. `eslint` 62 errors, byte-identical to `main`.

Four rounds, four rounds with findings — though the curve has flattened: rounds 2 and 3 found defects in the previous fix, round 4 found one pre-existing bug, one silent-drop, and one latent fragility that had been sitting in `main` unnoticed. I still want a second reviewer before merge.

jcfischer:
## Independent review by Sage — head `c3c2946`

Run offline via `sage review the-metafactory/soma#639` (its own lens pipeline, not the CodeReview SKILL.md I used for my four self-review passes). Verdict: **changes-requested — 21 findings (5 important, 13 suggestion, 3 nit, 0 blockers)**. Sage's `--post` did not reach the forge (`posted=false`, `github_review_id: 0`), so I am relaying its output verbatim below.

---


- **[suggestion]** `src/adapters/grok/install.ts:259` — **grok postRemove recomputes the loader dir instead of asking the spec**
  `loaderDir: skillsLoaderUnder()(substrateHome)` re-derives a path the spec already owns, while claude-code's equivalent reads `claudeCodeInstallSpec.skillsLoaderDir(context.substrateHome)`. If grok's `skillsLoaderDir` ever takes an argument or moves, uninstall silently scans the wrong directory and orphans the links again.
  _Subject was added after Sage's previous review._
  Fix: Use `grokInstallSpec.skillsLoaderDir(substrateHome)`, matching the claude-code call site.

- **[suggestion]** `src/adapters/grok/install.ts:254` — **Orphan-link cleanup added to only two of six substrates**
  The grok comment says the orphan symlink "Pre-dates the wholesale-projection work but is the same defect", yet `installSkills` honours `--skills` for every substrate (`if (selected.length > 0) return { dirs: selected.map(...) }`) and only claude-code and grok gained `removeProjectedSkillLinks`. The PR argues cursor and pi-dev are covered by owned-subtree reconcile; codex and anthropic-cowork are not addressed either way.
  _Subject was added after Sage's previous review._
  Fix: State in the PR (or fix) whether codex/anthropic-cowork loaders are owned subtrees; if not, add the same postRemove call.

- **[nit]** `test/skill-projection.test.ts:581` — **Test suites named after review rounds, not behaviour**
  `describe("soma#638 review fixes")`, `"re-review fixes"`, `"third-pass fixes"`, `"fourth-pass fixes"` name the process that produced them; a reader cannot tell from the suite name what area is covered or where a new case belongs.
  _Subject was added after Sage's previous review._
  Fix: Rename by subject, e.g. "registry scan: slot-name resolution", "uninstall: projected loader links".

### Security

- **[suggestion]** `src/cli/substrate-lifecycle.ts:537` — **--skills names are unvalidated path segments**
  `dirs: selected.map((name) => resolveJoin(somaHome, "skills", name))` with "An explicit --skills list is taken at face value" lets `--skills ../../../tmp/evil` resolve outside the registry and symlink arbitrary content into the harness skill loader, where it becomes agent-loadable instructions. The scan path already rejects traversal (a frontmatter `name: ../escape` lands in `unprojectable`), so the two entry points disagree on the same invariant.
  _Subject was added after Sage's previous review._
  Fix: Reject selected names containing a path separator or `..`, or assert the resolved dir stays under `<somaHome>/skills` before returning it.

- **[suggestion]** `src/projected-skill-links.ts:40` — **Uninstall deletes loader symlinks by target, not ownership**
  `A loader symlink pointing into the soma skills registry is unambiguously ours` is a heuristic, not a record — a principal-authored `~/.claude/skills/mine -> ~/.soma/skills/mine` is deleted by `soma uninstall`, in a directory the adapter comment documents as `SHARED (principal-authored + PAI-migrated skills)` whose removals `round-trip via the install manifest`. The tests cover a user-owned real dir and a link outside the registry, but not a user-authored link into it.
  _Subject was added after Sage's previous review._
  Fix: Record projected links in the install manifest and delete only recorded paths, falling back to the target heuristic when the manifest is absent.

- **[suggestion]** `src/skill-projection.ts:624` — **External registry symlinks reach agent context with no reproject gate**
  `Symlinks count: a registry slot filled by project-skill from an external source dir is as canonical as one authored in place` plus wholesale loader linking means a registry entry pointing at a cloned third-party repo has its SKILL.md live in every session — the PR verifies `editing ~/.soma/skills/<name>/SKILL.md is live in the loader with no reproject`. A `git pull` in that source now changes agent instructions with no principal action, where the copy path previously required an explicit install.
  _Subject was added after Sage's previous review._
  Fix: Note the exposure in the #638 doctor check: report registry entries whose symlink target resolves outside `~/.soma`.

### Architecture

- **[important]** `src/cli/substrate-lifecycle.ts:546` — **Wholesale skill projection lives only in the CLI layer**
  `installSkills` gates on `installSpecFor(substrate).skillsDiscovery !== "loader"` inside `src/cli/`, so only `runSubstrateLifecycleCli` links the registry. The exported library install (`installSomaForClaudeCode`, used by `test/claude-code-install.test.ts`) now emits neither `rules/soma/SKILLS.md` nor loader symlinks, so a programmatic claude-code install yields a home with no discoverable skills at all — strictly worse than before this PR.
  _Subject was added after Sage's previous review._
  Fix: Move the registry→loader projection into the install layer (`runInstall` / the claude-code install spec's post-projection steps) and leave the CLI to format its result.

- **[important]** `src/install-spec.ts:69` — **New public type's docs adopt `harness`, a banned alias**
  CONTEXT.md §substrate: "Do not use `harness`, `host`, `runtime`, `platform`, `agent runtime` in Soma docs or code. `substrate` is the only word." The new `SkillsDiscoveryMode` JSDoc says "the harness advertises the contents of" and "names the harness cannot resolve", and the term repeats in `projected-skill-links.ts`, `skill-projection.ts`, `substrate-lifecycle.ts` and the test names — this is the doc comment every future adapter author will copy.
  Fix: s/harness/substrate/ across the new comments; also "Soma emits NO catalog" → "projects no catalog" (CONTEXT.md §project kills `emit`).

- **[suggestion]** `src/install-spec.ts:84` — **`loader` conflates discovery capability with owned-subtree safety**
  `skillsDiscovery` is documented as "whether the harness advertises what it holds at all", yet it also gates wholesale symlinking, which is only safe when `skillsLoaderDir` sits outside `ownedSubtrees` — a precondition stated in the PR body and enforced nowhere on `SubstrateInstallSpec`. Setting `skillsDiscovery: "loader"` on cursor or pi-dev would have `reconcileOwnedSubtrees` delete the links the same install just created.
  Fix: Assert (in the six-substrate pinning test or at spec load) that a `loader` substrate's `skillsLoaderDir` is not inside any `ownedSubtrees` entry.

### Performance

- **[suggestion]** `src/cli/substrate-lifecycle.ts:547` — **Every registry SKILL.md is read twice per apply**
  `scanRegistrySkills` resolves each slot name by reading the full SKILL.md, then the result is narrowed to `dirs: scan.skills.map((skill) => skill.dir)` — dropping the names, so `linkSkill` re-reads every file to recompute the same string. ~120 registry skills means ~240 whole-file reads per `--apply`/`reproject` where 120 would do.
  _Subject was added after Sage's previous review._
  Fix: Thread `scan.skills` ({dir, name}) through `projectInstallSkills` into `projectSkills` so `linkSkill` can take a pre-resolved name.

- **[suggestion]** `src/cli/substrate-lifecycle.ts:512` — **Whole registry linked serially, now on every reproject too**
  `const projected = await projectInstallSkills(parsed.substrate, dirs, parsed.options)` takes the default skill count from 0 to the entire registry, and `projectSkills` links with `for (const dir of skillDirs) await linkSkill(...)` — a serial readFile + lstat + symlink chain per skill, ~1k serialized syscalls. `withProjectedSkills` puts that on the reproject/upgrade path as well.
  _Subject was added after Sage's previous review._
  Fix: Batch `linkSkill` with a bounded concurrency (e.g. 8) inside `projectSkills`; keep the existing rollback set shared across the batch.

- **[suggestion]** `src/adapters/shared/index.ts:256` — **Catalog refresh builds a projection that can never match**
  `files.filter((path) => !(path === catalogPath && discovery === "loader"))` means claude-code emits no SKILLS.md, but `projectSkills` still ends in `refreshSkillCatalogs`, which does a second `loadSomaHome` plus a full `buildSubstrateHomeProjection` (rendering the ~120-entry catalog string) purely to search for a file that is no longer in the bundle. Wasted on every install and reproject — and it also trips the "no SKILLS catalog file found" stderr warning.
  Fix: Skip the refresh when the substrate's `skillsDiscovery === "loader"`.

### Maintainability

- **[suggestion]** `src/skill-projection.ts:628` — **scanRegistrySkills does four jobs in one function**
  One body covers readdir + VSA/name filtering, per-entry classification (`const scanned: ScanRow[] = await Promise.all(...)`), slot-clash dedupe via `claimedBy`, and two sorts — with a locally-declared `type ScanRow` union to hold it together. The clash rule (`loader slot "${row.name}" is already claimed by`) is the part most likely to change and is hardest to see under the scan.
  Fix: Extract `classifyRegistryEntry(root, entry): Promise<ScanRow>` and `resolveSlotClashes(rows): RegistrySkillScan`; leave `scanRegistrySkills` as readdir → map → resolve.

- **[nit]** `test/skill-projection.test.ts:489` — **authorSkill helper redeclared in four describe blocks**
  `async function authorSkill(homeDir, dir, frontmatterName)` with the identical `---\nname: ${frontmatterName}\n---\n# ${dir}\n` body is copy-pasted into "review fixes", "re-review fixes", "third-pass fixes", plus a two-arg variant in "the registry is the curated set".
  Fix: Hoist one `authorSkill(homeDir, dir, frontmatterName = dir)` to module scope and delete the four copies.

- **[nit]** `test/skill-projection.test.ts:486` — **Describe blocks named by review round, not behaviour**
  `soma#638 review fixes`, `soma#638 re-review fixes`, `soma#638 third-pass fixes`, `soma#638 fourth-pass fixes` name the PR's review history rather than what is under test, so a future reader cannot tell where a new skill-scan case belongs.
  Fix: Regroup by subject: "registry scan", "plan output", "uninstall link cleanup".

### HonestOracle

- **[important]** `src/install-spec.ts:70` — **Harness discovery of symlinked skills asserted, never probed**
  `loader` is defined as "the harness advertises the contents of {@link SubstrateInstallSpec.skillsLoaderDir} itself", and on that basis the catalog — the only fallback list — is deleted; but every verification bullet is a filesystem assertion on a temp home ("symlinks the whole registry", `isSymbolicLink()`), with no evidence that a real Claude Code session resolves a symlinked `skills/<name>` as a registered skill. If it does not, skills go from double-listed to invisible with no signal.

- **[important]** `test/skill-projection.test.ts:489` — **Reported test delta does not match the tests this diff adds**
  The description claims "`bun test` — **2448 pass, 0 fail** (baseline 2438); 10 new tests", but the diff adds 22 `test(...)` cases (19 here, 3 in claude-code-install.test.ts) and deletes none — so the cited run predates roughly half of them and does not evidence that this diff is green.

- **[suggestion]** `src/adapters/claude-code/install.ts:91` — **Migration-from-copies claim unqualified and untested**
  "a home installed on `main` (real-dir copies) migrates to symlinks on reinstall" is asserted from one manual temp-home run, while this diff's own comment says the copy round-trip preserves "user-edited files" and another test comment says "ensureSymlink refuses to replace a non-symlink and the whole install aborts" — the intersection of the two is never engaged or pinned by a test, on a home the PR says already holds 50 drifted copies.
  _Subject was added after Sage's previous review._

- **[suggestion]** `src/cli/substrate-lifecycle.ts:527` — **"names are loader slot names" is false on the --skills path**
  The doc states unconditionally that "`names` are the LOADER SLOT names, resolved from frontmatter — not the dir basenames", yet the branch immediately below returns `names: selected` (dir basenames), so `--skills dir-name` prints exactly the promise-a-path-apply-never-creates failure the comment claims to prevent; the accompanying test covers only the no-flag path.
  _Subject was added after Sage's previous review._

- **[suggestion]** `src/projected-skill-links.ts:38` — **"Unambiguously ours" overclaims target-based ownership**
  "A loader symlink pointing into the soma skills registry is unambiguously ours" — identification is by target, not provenance, so a principal's own hand-made link into `~/.soma/skills` is removed on uninstall; the diff's survival test covers only a real directory (`UserOwned`), not a user link into the registry.
  _Subject was added after Sage's previous review._


---

jcfischer:
Correcting one of Sage's findings against myself before anything else: the PR description claimed `2448 pass … 10 new tests`, written at `e2846dd` and never updated. Head `c3c2946` is 2460 pass and 22 new tests. The cited run predated roughly half the tests in the diff, so it did not evidence that this diff is green. Description now updated to the head numbers.

Two of Sage's findings do not survive checking, and I would rather say so than accept them silently:

- **`--skills` path traversal (security, suggestion)** — `parseSkillNames` already rejects path-like names at `substrate-lifecycle.ts:356`. Probed: `--skills ../../evil` → `--skills takes skill names, not paths (got "../../evil")`, nothing linked. Sage read `installSkills` in isolation and missed the CLI-layer validation. The doc comment there is still misleading and I will reword it.
- **Catalog refresh builds a projection that can never match (performance)** — `refreshSkillCatalogs` already `continue`s on `skillsDiscovery === "loader"` before `findCatalogFile`, so neither `buildSubstrateHomeProjection` nor the stderr warning runs. Only the single `loadSomaHome` at the top is unconditional — real, but a fraction of what was described.

PR #654, opened 2026-08-24
fix(skills): reconcile migrated symlink slots safely

### LINKED ISSUE #652: skills: install aborts with ENOTDIR when a reconciled loader slot is a symlink
## Symptoms

`soma install claude-code --apply` aborts the **entire** install with:

```
ENOTDIR: not a directory, rmdir '/Users/fischer/.claude/skills/the-algorithm'
```

After manually removing that one symlink and re-running, it aborts again on the next path:

```
ENOTDIR: not a directory, rmdir '/Users/fischer/.claude/skills/orienteer'
```

Each abort leaves the home half-projected (earlier steps applied, later skills unlinked).

## Trigger

A loader slot path that reconcile wants to remove/recreate is a **symlink**, not a real
directory. Observed in two distinct code paths:

1. **Portable-skill manifest consumption** (`src/adapters/shared/portable-skill-manifest.ts`,
   imports `rm`/`rmdir`). A July-era `install-manifest.json` recorded
   `skills/the-algorithm/**`, `skills/Memory/**`, `skills/migrate-pai-purpose/**` as COPIES.
   Those slots were since replaced by symlinks; consuming the manifest still treats the
   recorded dir as a directory.
2. **Bundled-skill owned-subtree reconcile** (`src/projection-reconcile.ts`).
   `removeEmptyDirs()` ends with `rmdir(root)` and tolerates only ENOENT/ENOTEMPTY
   (lines ~46–71); a symlinked root raises ENOTDIR, which propagates and kills the run.
   `src/skills/{the-algorithm,orienteer}` are the bundled set, so their loader slots hit this.

## Why symlinks exist there (legitimately)

#639/#643 moved claude-code to loader-mode whole-registry projection: registry skills land
as symlinks into the harness loader. Homes installed before that change hold manifests and
copies from the copy era; the migration path (copy → symlink) is exactly the shape that
crashes.

## Reproduction sketch (unit-level)

1. Project a substrate whose manifest records `skills/foo/**`.
2. Replace `<loader>/skills/foo` with a symlink pointing anywhere.
3. Run install `--apply`.
4. Observe ENOTDIR abort; no skill beyond the failure point is linked.

(Real-world reproduction: macOS 26, APFS case-insensitive, soma 0.18.0, Aug 22 2026 —
two consecutive aborted runs; workaround was manually `rm`-ing each offending symlink,
after which the same install completed cleanly.)

## Fix direction

- Classify the slot with `lstat` **before** removal: symlink → `unlink(linkPath)`;
  real dir → existing rmdir flow. The data-loss reasoning already present in
  `ensureSymlink` (symlink replacement loses nothing) applies here verbatim.
- Defensively map ENOTDIR to tolerated at the `rmdir` sites as well.
- Regression test: manifest-recorded path + bundled owned-subtree root realized as
  symlinks; assert install completes and the symlink is replaced by the intended shape.

## Acceptance criteria

- [ ] Install completes when any reconciled slot is a symlink (manifest path and bundled path).
- [ ] No partial-application cliff: one bad slot cannot prevent the remaining skills from projecting.
- [ ] Regression tests cover both reconcile paths.

Observed against soma 0.18.0 (fe6be2a); the cited code paths are unchanged on current main.

jcfischer:
## Summary
- prevent copy-era manifest cleanup from following a migrated skill-slot symlink into its registry target
- revalidate the slot symlink identity before unlinking and preserve shared loader roots
- tolerate benign ENOTDIR cleanup races and cover manifest, owned-subtree, and full Claude install migration paths

Closes #652

## Verification
- `bun test test/portable-skill-manifest.test.ts test/projection-reconcile.test.ts test/claude-code-install.test.ts` — 51 pass
- `bun run typecheck`
- `bunx eslint src/adapters/shared/portable-skill-manifest.ts src/projection-reconcile.ts test/portable-skill-manifest.test.ts test/projection-reconcile.test.ts test/claude-code-install.test.ts`
- Full `bun test` was run: 2491 pass; one existing unrelated failure reports an arc manifest/package version mismatch (0.17.0 vs 0.18.0).

### DIFF
--- a/src/adapters/shared/portable-skill-manifest.ts
+++ b/src/adapters/shared/portable-skill-manifest.ts
@@ -1,5 +1,5 @@
 import { createHash } from "node:crypto";
-import { mkdir, readFile, rm, rmdir, writeFile } from "node:fs/promises";
+import { lstat, mkdir, readFile, rm, rmdir, unlink, writeFile } from "node:fs/promises";
 import { dirname, join, resolve, sep } from "node:path";
 import { isEnoent } from "../../fs-errors";
 
@@ -115,6 +115,29 @@ function isInsideRoot(root: string, target: string): boolean {
   return target !== root && target.startsWith(rootPrefix);
 }
 
+interface SymlinkIdentity {
+  path: string;
+  dev: number;
+  ino: number;
+}
+
+/** Find a directory symlink above a manifest-recorded file without following it. */
+async function symlinkAncestor(root: string, target: string): Promise<SymlinkIdentity | undefined> {
+  for (let dir = dirname(target); isInsideRoot(root, dir); dir = dirname(dir)) {
+    try {
+      const entry = await lstat(dir);
+      if (entry.isSymbolicLink()) return { path: dir, dev: entry.dev, ino: entry.ino };
+    } catch (error) {
+      const code = (error as NodeJS.ErrnoException).code;
+      // A missing path or a non-directory ancestor can be the result of another
+      // manifest entry already removing the same migrated slot. Keep walking so
+      // a higher symlink ancestor can still be identified.
+      if (code !== "ENOENT" && code !== "ENOTDIR") throw error;
+    }
+  }
+  return undefined;
+}
+
 export async function readPortableSkillManifest(
   somaHome: string,
   substrate: PortableSkillManifestSubstrate,
@@ -147,6 +170,34 @@ async function removeListedProjectionFiles(
   for (const file of files) {
     const target = resolve(substrateHome, file.path);
     if (!isInsideRoot(substrateHome, target)) continue;
+
+    // Copy-era manifests name files below skill directories. Loader-mode
+    // migration may have replaced such a directory with a symlink. Reading the
+    // recorded file first would follow that link and rm() would delete registry
+    // content. Remove the slot link itself before touching any recorded child.
+    const linkedDir = await symlinkAncestor(substrateHome, target);
+    if (linkedDir !== undefined) {
+      // The substrate's top-level directories are shared surfaces, not
+      // manifest-owned skill slots. If one is itself a symlink, fail open for
+      // this entry rather than unlinking the whole shared loader root.
+      if (dirname(linkedDir.path) === substrateHome) continue;
+      try {
+        // Narrow the lstat→unlink race: only unlink the same symlink inode we
+        // classified. A swapped regular file or replacement link is user state,
+        // so this manifest entry fails open instead of deleting it.
+        const current = await lstat(linkedDir.path);
+        if (!current.isSymbolicLink() || current.dev !== linkedDir.dev || current.ino !== linkedDir.ino) continue;
+        await unlink(linkedDir.path);
+        removed.push(linkedDir.path);
+      } catch (error) {
+        if (!isEnoent(error)) throw error;
+      }
+      for (let dir = dirname(linkedDir.path); isInsideRoot(substrateHome, dir); dir = dirname(dir)) {
+        candidateDirs.add(dir);
+      }
+      continue;
+    }
+
     let content: string;
     try {
       content = await readFile(target, "utf8");
@@ -173,7 +224,7 @@ async function removeListedProjectionFiles(
       // (EACCES/EPERM/EBUSY/…) is a real filesystem fault and must surface,
       // not be silently masked as a keep.
       const code = (error as { code?: string } | null)?.code;
-      if (code !== "ENOTEMPTY" && code !== "ENOENT") throw error;
+      if (code !== "ENOTEMPTY" && code !== "ENOENT" && code !== "ENOTDIR") throw error;
     }
   }
   return removed;
--- a/src/projection-reconcile.ts
+++ b/src/projection-reconcile.ts
@@ -66,7 +66,7 @@ async function removeEmptyDirs(root: string, excludedAbs: readonly string[] = []
     await rmdir(root);
   } catch (error) {
     const code = (error as NodeJS.ErrnoException).code;
-    if (code !== "ENOENT" && code !== "ENOTEMPTY") throw error;
+    if (code !== "ENOENT" && code !== "ENOTEMPTY" && code !== "ENOTDIR") throw error;
   }
 }
 
--- a/test/portable-skill-manifest.test.ts
+++ b/test/portable-skill-manifest.test.ts
@@ -1,6 +1,6 @@
 import { expect, test } from "bun:test";
 import { createHash } from "node:crypto";
-import { mkdtemp, mkdir, readFile, rm, stat, writeFile } from "node:fs/promises";
+import { lstat, mkdtemp, mkdir, readFile, rm, stat, symlink, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { dirname, join, resolve } from "node:path";
 import {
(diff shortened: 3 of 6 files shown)


src/adapters/shared/portable-skill-manifest.ts
```diff
@@ -115,6 +115,29 @@ function isInsideRoot(root: string, target: string): boolean {
...
+  ino: number;
+}
+
+/** Find a directory symlink above a manifest-recorded file without following it. */
+async function symlinkAncestor(root: string, target: string): Promise<SymlinkIdentity | undefined> {
+  for (let dir = dirname(target); isInsideRoot(root, dir); dir = dirname(dir)) {
+    try {
+      const entry = await lstat(dir);
+      if (entry.isSymbolicLink()) return { path: dir, dev: entry.dev, ino: entry.ino };
+    } catch (error) {
```
jcfischer, line 130: **[suggestion] ENOTDIR swallowed in symlinkAncestor skips real permission errors**

The catch block silently ignores EACCES and EPERM alongside ENOENT/ENOTDIR: `if (code !== "ENOENT" && code !== "ENOTDIR") throw error;`. The comment says the walk is for 'missing path or non-directory ancestor,' but permission-denied on a parent directory is neither — it means the process genuinely cannot inspect that path, and continuing may produce a wrong result (no symlink found when one exists above the blocked directory). The isInsideRoot guard still bounds the walk, but the caller gets `undefined` instead of the error that would let it fail appropriately.

Fix: Add EACCES and EPERM to the throw conditions: `if (code !== "ENOENT" && code !== "ENOTDIR" && code !== "EACCES" && code !== "EPERM") throw error;` — or, if the intent is to truly only skip missing/non-directory errors, restrict the comment and code to match.

src/adapters/shared/portable-skill-manifest.ts
```diff
@@ -147,6 +170,34 @@ async function removeListedProjectionFiles(
...
+        if (!current.isSymbolicLink() || current.dev !== linkedDir.dev || current.ino !== linkedDir.ino) continue;
+        await unlink(linkedDir.path);
+        removed.push(linkedDir.path);
+      } catch (error) {
+        if (!isEnoent(error)) throw error;
+      }
+      for (let dir = dirname(linkedDir.path); isInsideRoot(substrateHome, dir); dir = dirname(dir)) {
+        candidateDirs.add(dir);
+      }
+      continue;
```
jcfischer, line 198: **[suggestion] lstat→unlink inode guard fails closed on EACCES**

The guard `await lstat(linkedDir.path)` inside the try block re-throws all errors except isEnoent. If the symlink sits inside a directory where the process now lacks search permission, lstat throws EACCES, which propagates out of the try/catch and aborts the entire reconcile loop. The prior `symlinkAncestor` call already identified this path as a symlink, so failing here is conservative but may be unnecessarily harsh — a permission race on the symlink itself (not its target) is unlikely to matter for correctness.

Fix: Consider catching EACCES/EPERM from the guard lstat and treating it like an inode mismatch (continue). Optional, not required for correctness.

src/adapters/shared/portable-skill-manifest.ts
```diff
@@ -173,7 +224,7 @@ async function removeListedProjectionFiles(
       // (EACCES/EPERM/EBUSY/…) is a real filesystem fault and must surface,
       // not be silently masked as a keep.
       const code = (error as { code?: string } | null)?.code;
-      if (code !== "ENOTEMPTY" && code !== "ENOENT") throw error;
+      if (code !== "ENOTEMPTY" && code !== "ENOENT" && code !== "ENOTDIR") throw error;
```
jcfischer, line 227: **[nit] ENOTDIR justification missing in comment**

The catch for rmdir now tolerates ENOTDIR alongside ENOTEMPTY and ENOENT: `if (code !== "ENOTEMPTY" && code !== "ENOENT" && code !== "ENOTDIR") throw error;`. The existing comment above explains why ENOTEMPTY and ENOENT are safe to suppress — a sibling manifest entry may have already removed the directory — but doesn't mention ENOTDIR. A reader sees three tolerated codes with a comment that only addresses two.

Fix: Extend the comment to note that ENOTDIR can arise when the directory path was replaced by a non-directory (e.g., a symlink) between the stat and the rmdir call.

src/projection-reconcile.ts
```diff
@@ -66,7 +66,7 @@ async function removeEmptyDirs(root: string, excludedAbs: readonly string[] = []
     await rmdir(root);
   } catch (error) {
     const code = (error as NodeJS.ErrnoException).code;
-    if (code !== "ENOENT" && code !== "ENOTEMPTY") throw error;
+    if (code !== "ENOENT" && code !== "ENOTEMPTY" && code !== "ENOTDIR") throw error;
```
jcfischer, line 69: **[nit] Same ENOTDIR tolerance lacks rationale**

The `removeEmptyDirs` function in projection-reconcile.ts adds ENOTDIR tolerance with no comment: `if (code !== "ENOENT" && code !== "ENOTEMPTY" && code !== "ENOTDIR") throw error;`. This is the same pattern as the portable-skill-manifest change, but there's no comment explaining when a directory path would become a non-directory during cleanup.

Fix: Add a brief comment noting the race condition (a file or symlink replaces the directory between readdir and rmdir).

test/portable-skill-manifest.test.ts
```diff
@@ -163,6 +163,60 @@ test("reconcile removes only files the current projection dropped, keeping the m
...
 });
 
+test("reconcile removes a migrated symlink slot without touching its target (#652)", async () => {
+  await withDirs(async (somaHome, substrateHome) => {
+    const files = [
+      { path: "skills/the-algorithm/SKILL.md", content: "algorithm body\n" },
+      { path: "skills/the-algorithm/Workflows/Run.md", content: "workflow body\n" },
+      { path: "skills/gone/SKILL.md", content: "later stale skill\n" },
+    ];
+    await project(substrateHome, files);
```
jcfischer, line 173: **[nit] Test uses gone() without definition in diff**

The test at line 183 calls `await gone(join(substrateHome, "skills/gone"))` with `expect(...).toBe(true)`, but the diff doesn't show a `gone` helper being defined or imported. The existing test file likely has it, and it may compile fine, but the diff alone cannot confirm the helper exists.

Fix: Verify the `gone` helper is defined or imported elsewhere in the test file. If it doesn't exist, the test will fail at runtime with a ReferenceError.

src/adapters/shared/portable-skill-manifest.ts
```diff
@@ -147,6 +170,34 @@ async function removeListedProjectionFiles(
   for (const file of files) {
     const target = resolve(substrateHome, file.path);
     if (!isInsideRoot(substrateHome, target)) continue;
+
+    // Copy-era manifests name files below skill directories. Loader-mode
+    // migration may have replaced such a directory with a symlink. Reading the
```
jcfischer, line 175: **[suggestion] Extract symlink-slot removal into a named helper**

The block from `const linkedDir = await symlinkAncestor(...)` through the `continue` at line 189 nests ~25 lines inside the loop body and adds a second early-exit path. Extracting it into a `removeSlotSymlink` helper would make the loop read as a two-case dispatch (symlink vs. file) instead of interleaving both concerns.

Fix: Extract lines ~175-189 into `async function removeSlotSymlink(substrateHome: string, target: string, candidateDirs: Set<string>): Promise<string | undefined>` that returns the removed path or undefined, then call it as: `const slotRemoved = await removeSlotSymlink(...); if (slotRemoved !== undefined) { removed.push(slotRemoved); continue; }`

jcfischer (commented):

6 finding(s): 3 suggestion, 3 nit.

Round impact: behavior 1, checks 3, prose 2; unclassified impact 0; prior-round surface 0; restated from earlier rounds 0.


### CodeQuality

- **[suggestion]** `src/adapters/shared/portable-skill-manifest.ts:130` — **ENOTDIR swallowed in symlinkAncestor skips real permission errors**
  The catch block silently ignores EACCES and EPERM alongside ENOENT/ENOTDIR: `if (code !== "ENOENT" && code !== "ENOTDIR") throw error;`. The comment says the walk is for 'missing path or non-directory ancestor,' but permission-denied on a parent directory is neither — it means the process genuinely cannot inspect that path, and continuing may produce a wrong result (no symlink found when one exists above the blocked directory). The isInsideRoot guard still bounds the walk, but the caller gets `undefined` instead of the error that would let it fail appropriately.
  Fix: Add EACCES and EPERM to the throw conditions: `if (code !== "ENOENT" && code !== "ENOTDIR" && code !== "EACCES" && code !== "EPERM") throw error;` — or, if the intent is to truly only skip missing/non-directory errors, restrict the comment and code to match.

- **[suggestion]** `src/adapters/shared/portable-skill-manifest.ts:198` — **lstat→unlink inode guard fails closed on EACCES**
  The guard `await lstat(linkedDir.path)` inside the try block re-throws all errors except isEnoent. If the symlink sits inside a directory where the process now lacks search permission, lstat throws EACCES, which propagates out of the try/catch and aborts the entire reconcile loop. The prior `symlinkAncestor` call already identified this path as a symlink, so failing here is conservative but may be unnecessarily harsh — a permission race on the symlink itself (not its target) is unlikely to matter for correctness.
  Fix: Consider catching EACCES/EPERM from the guard lstat and treating it like an inode mismatch (continue). Optional, not required for correctness.

- **[nit]** `src/adapters/shared/portable-skill-manifest.ts:227` — **ENOTDIR justification missing in comment**
  The catch for rmdir now tolerates ENOTDIR alongside ENOTEMPTY and ENOENT: `if (code !== "ENOTEMPTY" && code !== "ENOENT" && code !== "ENOTDIR") throw error;`. The existing comment above explains why ENOTEMPTY and ENOENT are safe to suppress — a sibling manifest entry may have already removed the directory — but doesn't mention ENOTDIR. A reader sees three tolerated codes with a comment that only addresses two.
  Fix: Extend the comment to note that ENOTDIR can arise when the directory path was replaced by a non-directory (e.g., a symlink) between the stat and the rmdir call.

- **[nit]** `src/projection-reconcile.ts:69` — **Same ENOTDIR tolerance lacks rationale**
  The `removeEmptyDirs` function in projection-reconcile.ts adds ENOTDIR tolerance with no comment: `if (code !== "ENOENT" && code !== "ENOTEMPTY" && code !== "ENOTDIR") throw error;`. This is the same pattern as the portable-skill-manifest change, but there's no comment explaining when a directory path would become a non-directory during cleanup.
  Fix: Add a brief comment noting the race condition (a file or symlink replaces the directory between readdir and rmdir).

- **[nit]** `test/portable-skill-manifest.test.ts:173` — **Test uses gone() without definition in diff**
  The test at line 183 calls `await gone(join(substrateHome, "skills/gone"))` with `expect(...).toBe(true)`, but the diff doesn't show a `gone` helper being defined or imported. The existing test file likely has it, and it may compile fine, but the diff alone cannot confirm the helper exists.
  Fix: Verify the `gone` helper is defined or imported elsewhere in the test file. If it doesn't exist, the test will fail at runtime with a ReferenceError.

### Maintainability

- **[suggestion]** `src/adapters/shared/portable-skill-manifest.ts:175` — **Extract symlink-slot removal into a named helper**
  The block from `const linkedDir = await symlinkAncestor(...)` through the `continue` at line 189 nests ~25 lines inside the loop body and adds a second early-exit path. Extracting it into a `removeSlotSymlink` helper would make the loop read as a two-case dispatch (symlink vs. file) instead of interleaving both concerns.
  Fix: Extract lines ~175-189 into `async function removeSlotSymlink(substrateHome: string, target: string, candidateDirs: Set<string>): Promise<string | undefined>` that returns the removed path or undefined, then call it as: `const slotRemoved = await removeSlotSymlink(...); if (slotRemoved !== undefined) { removed.push(slotRemoved); continue; }`


---

jcfischer:
Thanks, Sage. I verified the review against the PR head.

- The EACCES/EPERM observation in `symlinkAncestor` is inverted: the existing condition already rethrows both; only ENOENT and ENOTDIR are tolerated. The revalidation lstat likewise fails closed on EACCES/EPERM, which is intentional.
- `gone()` is defined at the test-file seam before this regression.
- The ENOTDIR cases are deliberate cleanup races. I will retain the compact current shape rather than extract a helper without a behavior change.

Verdict is `commented` with zero blockers/important findings, so this review is an effective pass.

PR #659, opened 2026-08-24
fix(skills): recase stale projected symlinks

### LINKED ISSUE #653: skills: on case-insensitive volumes, equal-target links with stale casing are reported unchanged instead of recased
## Summary

On case-insensitive filesystems (macOS APFS default), `ensureSymlink`
(`src/skill-projection.ts:190–227`) reports `"unchanged"` for a loader slot whose **stored
name casing differs** from the requested slot name, as long as the resolved symlink target
matches. The stale casing is never normalized.

## Concrete failure chain (observed on a real home)

1. Registry frontmatter renamed 68 skills (`name: Council` → `name: council`, etc.),
   changing every derived loader slot name.
2. Re-running `soma install dsh --apply` / `claude-code --apply`: for each renamed skill,
   `lstat('<loader>/council')` resolves onto the pre-existing `Council` entry
   (case-insensitive lookup), `readlink` target equals the desired target →
   **"unchanged"**. Stored entry keeps its old casing; no conforming slot is ever created.
3. Plan-vs-landed diff showed 40 of 111 planned slots absent while the installer exited
   clean with no warnings.
4. An external cleanup then removed the stale-cased links (they *were* stale by name);
   result: those 40 skills silently lost ALL loader presence. Nothing in soma's output
   ever signaled the divergence.

## Why it matters

- Any consumer keying on the literal slot/dir name sees non-conforming names long after
  the registry conformed.
- Combined with any external tidy-up of "old-looking" links, skills become unreachable
  with zero signal — precisely the failure class `scanRegistrySkills`'s unprojectable
  reporting exists to prevent.

## Fix direction

In `ensureSymlink`, after the readlink-target equality check: if the stored entry's actual
basename (via `readdir(parentDir)`, since lstat folds case) case-folds to the requested
leaf but differs from it, take the `"replaced"` branch instead — `rm` + `symlink` under the
requested casing. On case-sensitive volumes the fold-equality test simply never fires for
genuinely different names, so the change is safe cross-platform. Alternative: detect
case-insensitivity once and gate the recase; always-on normalization is simpler and
matches the promise-what-lands contract.

Second-order gap worth folding in: neither the install plan nor the result surfaced
"slot exists under a different casing" as a reportable condition, unlike the existing
unprojectable/slot-collision reporting.

## Reproduction

On a case-insensitive volume:

1. `ln -s <dir> <loader>/Council`
2. Run install expecting slot `council` targeting the same resolved path.
3. Result: `"unchanged"`; `ls <loader>` shows `Council`, no `council`.

## Acceptance criteria

- [ ] Equal-target, differing-casing slots are recased to the requested slot name (result `"replaced"`).
- [ ] Install plan/result can express the recase (promise-what-lands holds on case-insensitive homes).
- [ ] Test coverage for the case-folding branch (CI on case-sensitive fs may need a helper-level test or a tmpfs toggle).

Observed against soma 0.18.0 (fe6be2a); `ensureSymlink` unchanged on current main.
Related: #639, #643.

jcfischer:
Closes #653

## Summary
- normalize equal-target symlinks whose stored slot casing differs on case-insensitive filesystems
- expose the prior spelling in dry-run and apply output
- cover the case-folding decision without requiring a case-insensitive CI volume

## Verification
- `bun test test/skill-projection.test.ts`
- `bun run typecheck`
- `bunx eslint src/skill-projection.ts src/cli/skill-projection-cli.ts test/skill-projection.test.ts`
- `bun test` (existing unrelated failures: arc-manifest version mismatch, one Codex executor timeout, and two 5s timeout flakes)

### DIFF
--- a/src/cli/skill-projection-cli.ts
+++ b/src/cli/skill-projection-cli.ts
@@ -157,7 +157,8 @@ function formatPlan(verb: string, plan: SkillProjectionPlan, marker: string): st
         `- ${link.scope === "registry" ? "registry" : link.substrate} ${marker} ${link.path}` +
         // Name the shape only when it is not the symlink default, so existing
         // on-demand output is unchanged (soma#542).
-        (link.kind === "stub" ? " (stub: frontmatter only)" : ""),
+        (link.kind === "stub" ? " (stub: frontmatter only)" : "") +
+        (link.recasedFrom ? ` (recase from ${link.recasedFrom})` : ""),
     ),
     "",
     `Catalog refresh: ${plan.catalogRefresh.join(", ")}`,
@@ -173,7 +174,8 @@ function formatProjectionResult(title: string, result: SkillProjectionResult): s
     "",
     "Links:",
     ...result.links.map((link) =>
-      `- ${link.scope === "registry" ? "registry" : link.substrate}: ${link.status} ${link.path}`,
+      `- ${link.scope === "registry" ? "registry" : link.substrate}: ${link.status} ${link.path}` +
+        (link.recasedFrom ? ` (recased from ${link.recasedFrom})` : ""),
     ),
     "",
     "Catalog:",
--- a/src/skill-projection.ts
+++ b/src/skill-projection.ts
@@ -81,6 +81,8 @@ export interface SkillLink {
   path: string;
   target?: string;
   status: SkillLinkStatus;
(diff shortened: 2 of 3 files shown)


jcfischer:
Maintainer accepted the current coverage approach. Targeted projection tests, changed-file lint, and typecheck pass; the prior full-suite run had unrelated baseline failures/timeouts.

PR #661, opened 2026-08-25
feat(runtime): freeze enforcement artifact

jcfischer:
## Summary
- stage, hash-validate, and atomically activate the guarded runtime under `~/.soma/runtime/`
- route Claude Code, Codex, and Grok hooks through `runtime/current`; add runtime status/rollback and doctor recovery guidance
- require structurally valid close receipts for bridged completion and test guarded private-root agreement

Closes #657

## Verification
- `bun test` (2531 passed)
- `bun run typecheck`
- `git diff --check`

## Deferred
- Per-invocation artifact hashing, network/Git freshness checks, automatic rollback, tracker reopening, and adapter redesign remain out of W1 scope.

Issue #657, opened 2026-08-24
Implement the W1 frozen-artifact and completion contract


jcfischer:
## Question

Implement the ratified W1 frozen-artifact contract from [the W1 contract decision](https://github.com/the-metafactory/soma/issues/648), without widening it into supply-chain verification or an adapter redesign.

### Scope

- Stage an immutable, versioned enforcement artifact at `~/.soma/runtime/<hash>/`;
  load-check it; atomically point guarded substrate hooks at it; retain the preceding
  artifact; and provide an explicit local switch back.
- Stop guarded Claude Code, Codex, and Grok hooks from resolving policy through an
  editable Soma checkout. A bad/missing configured artifact still fails closed and
  names the recovery action.
- Make graph consumers distinguish a receipt-backed **Soma-complete** close from a
  receipt-less tracker close; expose the latter as an exception without an automatic
  reopener.
- Add the cross-substrate private-root drift test: install spec, projected hook
  config, and policy evaluation must resolve identical roots.
- Add doctor/install reporting for the active enforcement artifact and its recovery
  path.

### Constraints

- No per-invocation hashing, network/Git freshness checks, automatic rollback,
  tracker-reopening writer, or adapter architecture redesign.
- Preserve fail-closed policy decisions and Pi.dev's loader-safe import boundary.
- If implementation exposes a decision not settled by the contract, create a
  scaffold node beneath this node; do not decide it implicitly.

### What closing looks like

1. A deliberately broken editable checkout does not affect a guarded operation
   running through the installed artifact; a policy deny still denies.
2. A missing/unloadable configured artifact fails closed and reports how to recover.
3. New artifact activation is load-checked and atomic; the explicit prior-artifact
   switch restores guard behavior.
4. A receipt-less GitHub auto-close is not treated as Soma-complete.
5. All substrate private-root sources agree.
6. `bun test` and `bunx tsc --noEmit` pass.

PR #668, opened 2026-08-25
fix(claude-code): the policy guard runs a pinned runtime, not the worktree

### LINKED ISSUE #640: Policy guard loads the soma worktree, so editing soma can brick the session that edits it
## What happened

While editing soma itself (PR #639), I renamed an export in `src/skill-projection.ts` and had not yet updated its importer in `src/cli/substrate-lifecycle.ts` — a broken-import window of a few seconds that every refactor passes through.

Every subsequent tool call died:

```
Soma runtime policy inspection failed closed:
SyntaxError: Export named 'listRegistrySkillDirs' not found in module
  '/Users/fischer/work/mf/soma/src/skill-projection.ts'
```

Including the tools needed to repair it. `Read`, `Edit`, `Write` and `Bash` are all matched by the guard, so the session could not open the file, could not fix the import, and could not even run `git checkout`. Recovery required a human to run a shell command outside the agent.

## Mechanism

`soma-policy-guard.mjs` is fail-closed by explicit design — a documented SECURITY INVARIANT, and correct as such. The problem is not fail-closed; it is **what it fails closed on**.

`runInspect` shells out with the working tree as cwd:

```js
return spawnSync(config.bunPath, args, {
  cwd: config.trustedSomaRepo,   // /Users/fischer/work/mf/soma
  ...
});
```

`soma-policy-guard.config.json`:

```json
{ "somaHome": "...", "trustedSomaRepo": "/Users/fischer/work/mf/soma", "bunPath": "..." }
```

So every guarded tool call type-loads soma's **live, mid-edit source**. Editing soma is therefore editing the thing that authorises editing. Any transient syntax error, broken import, or half-applied rename in `src/` — the normal intermediate states of a refactor — takes the whole session down until a human intervenes.

`PreToolUse` matcher is `Bash|Read|Edit|Write|MultiEdit|NotebookEdit`, i.e. every tool that could repair the fault. There is no bypass env var, and the deny message names the error but not a recovery path.

## Why it matters beyond the inconvenience

Two second-order effects, both worse than the lost time:

- **It makes soma uniquely dangerous to work on with an agent.** Every other repo can be refactored freely; its own is the one place where an intermediate state can brick the tooling.
- **It creates pressure to weaken fail-closed**, which would be the wrong lesson. Fail-closed is right. The bug is that a *policy decision* is coupled to the *edit surface*.

## Proposed direction

Decouple the guard's runtime from the working tree, preserving fail-closed exactly as it is.

- Point `trustedSomaRepo` at an **installed** soma — a built/pinned artifact (e.g. under `~/.soma/runtime/`) refreshed by `soma install claude-code`, not the git worktree. A broken working tree then cannot affect the guard, and the guard keeps enforcing the last known-good policy.
- Keep fail-closed for every decision path. This changes *which* source is loaded, not what happens when loading fails.
- Distinguish the two failures in the deny message: "policy denied this action" vs "the guard could not load" are different operator problems. The second should name the recovery (`soma install claude-code --apply`, or the pinned-runtime path to inspect).
- Consider a `soma doctor` check that the guard's configured runtime loads at all, so a broken one is caught at doctor time rather than at the next tool call.

## Acceptance

- [ ] The guard resolves its runtime from an installed artifact, not the repo working tree.
- [ ] A deliberately broken `src/` in the soma worktree does NOT block tool calls in a session editing that worktree.
- [ ] Fail-closed behaviour is unchanged for every policy decision (test asserts a deny still denies).
- [ ] Guard-load failure is reported distinctly from policy denial, and the message names a recovery step.
- [ ] `soma doctor` reports an unloadable guard runtime.

## Notes

- Same shape presumably applies to the codex `codex-hook-entry.mjs` contract the header references, and to any other substrate whose guard shells into `trustedSomaRepo`.
- Encountered on 2026-08-19 while working #638 / PR #639.

jcfischer:
Closes #640.

## The bug

`soma-policy-guard.config.json` carried `trustedSomaRepo` = the git working tree, and `runInspect` spawned `bun src/cli.ts` with that as cwd. So every guarded tool call type-loaded soma's **live, mid-edit source**. Any transient syntax error, broken import, or half-applied rename — the normal intermediate states of a refactor — denied `Bash|Read|Edit|Write|MultiEdit|NotebookEdit`, i.e. every tool that could repair the fault. Recorded four times.

Fail-closed is right and is not relaxed here. The bug is that a *policy decision* was coupled to the *edit surface*.

## The fix

`soma install claude-code` builds a self-contained CLI bundle and freezes its absolute path into the hook configs:

```text
<somaHome>/runtime/
├── soma-cli.mjs     # bun build src/cli.ts --target=bun  (1.4 MB, ~91 ms)
└── runtime.json     # { version, entry, builtFrom, sourceEntry }
```

- **The build is the load check.** `bun build` followed by a `--version` probe. A tree that cannot produce a loadable runtime fails the *build*; the previously pinned runtime stays in place and keeps enforcing the last known-good policy rather than publishing rubble.
- **Publish is a `rename`**, so a hook reading mid-install sees the old runtime or the new one, never a truncated file.
- **No timestamp in `runtime.json`** — install stays byte-idempotent (asserted).
- **Bundle built once per process**, so a multi-substrate install and the test suite pay for one build, not N.
- **Backwards compatible:** a config with no `runtimeEntry` falls back to `trustedSomaRepo`/`src/cli.ts` exactly as before. No install breaks by omission, and none is forced.

Guard, mode classifier, precompact and the lifecycle hook-runner all resolve through the same `somaRuntime(config)` helper — they share one config, so leaving three of them pointed at the worktree would have been incoherent.

### Deny messages now carry a failure class

"Policy denied this action" and "the guard could not run" are different operator problems, and only the second has a recovery. The second is now prefixed `Soma policy guard UNAVAILABLE (fail-closed — this is not a policy denial)` and names `soma install claude-code --apply`. A rule that fired still carries the rule's own reason. Both still deny.

### `soma doctor`

Follows whatever the installed config points at and reports two states:

| finding | severity | means |
|---|---|---|
| `claude-code-policy-guard-runtime-unloadable` | `error` | every guarded tool call is being denied right now |
| `claude-code-policy-guard-runtime-unpinned` | `warning` | still loading the working tree (soma#640 exposure) |

### `runtime/` is gitignored in the home

Joins `memory/INDEX.md` in the snapshot **generated files** block — 1.4 MB of rebuildable output that changes with every version bump has no business in snapshot churn. (Adopters already ignore compiled skill binaries by hand for the same reason.)

## Acceptance criteria

- [x] The guard resolves its runtime from an installed artifact, not the repo working tree.
- [x] A deliberately broken `src/` in the soma worktree does NOT block tool calls in a session editing that worktree.
- [x] Fail-closed behaviour is unchanged for every policy decision (test asserts a deny still denies).
- [x] Guard-load failure is reported distinctly from policy denial, and the message names a recovery step.
- [x] `soma doctor` reports an unloadable guard runtime.

## Verification

`test/claude-code-policy-guard-runtime.test.ts` — 9 tests, including a **reproduction** of the original failure on the unpinned path (a legacy config against a mid-rename tree denies `ls -la`), so the two paths are asserted side by side.

End-to-end against a real scratch install:

```
== benign, pinned + BROKEN worktree:
{"continue":true}
== exfil, pinned + BROKEN worktree:
{"...","permissionDecision":"deny","permissionDecisionReason":"Runtime policy denied this action: credential-file-egress."}
== benign, UNPINNED + BROKEN worktree (the reported bug):
{"...","permissionDecision":"deny","permissionDecisionReason":"Soma policy guard UNAVAILABLE (fail-closed — this is not a policy denial): ... run `soma install claude-code --apply` ..."}
```

`soma doctor --substrate claude-code`: reports `-unloadable` (error) on a broken entry, silent after reinstall.

Gates: `bun test` 2551 pass / 0 fail · `tsc --noEmit` clean · `eslint` 72 problems with **and** without the change (verified by `git stash` comparison — this PR adds none).

## Not in scope

Per the issue's Notes: the codex `codex-hook-entry.mjs` and grok hook entries still spawn `trustedSomaRepo` and carry the same exposure. The pinned runtime is deliberately substrate-neutral (it lives under the Soma home, not a substrate home), so adopting it there is a small follow-up. The claude-code feedback-capture helper also still shells `bun run soma` through the repo's `package.json` scripts — fail-open background capture, a different shape, left alone.

https://claude.ai/code/session_017HHB5DjnCNUWVt35AiBRj2


jcfischer:
Closing as superseded by #661 (issue #657), which landed the same fix with broader scope. Not a rebase candidate: the `CONFLICTING` state is the two designs colliding, and merging would leave soma with two runtime-pinning mechanisms.

**Verified against a scratch install off `01f05ce`**, not read from source:

```json
"trustedSomaRepo": "<somaHome>/runtime/claude-code/current"
```

resolving to `runtime/artifacts/33e37e11…` with its own `src/cli.ts` inside. `install.ts:159`: *"Guarded substrates execute only the immutable artifact, never this editable checkout."*

Where main's version is broader:

| | this PR | #661 on main |
|---|---|---|
| substrates | `claude-code` only — this PR's own body flagged codex/grok as still exposed | `claude-code`, `codex`, `grok` |
| artifact | one `bun build` bundle + `runtime.json` | content-addressed snapshot per source hash |
| recovery | reinstall | `soma runtime status` / `rollback --substrate` |
| hardening | — | symlink refusal, read-only payloads, unreferenced pruning |
| tests | 9 | 13 |

Its `soma doctor` contribution is also already covered: `onboarding.ts:311` emits `runtime-artifact-unloadable` / `-missing` at `error` severity with a recovery action.

**What was worth keeping is in #673:**

1. The deny-message failure class. Main names a recovery but does not separate "the guard could not run" from "policy denied you" — an operator has to infer it from a bun stack trace. This PR's `UNAVAILABLE (fail-closed — this is not a policy denial)` framing is ported.
2. The end-to-end reproduction test. Main's 13 tests cover staging mechanics; none puts a guarded tool call against a broken source checkout, so the actual #640 symptom was untested. #673 asserts the allow (pinned) and deny (unpinned) paths side by side.

The branch stays on the remote if anyone wants to read the alternative design; #640 will close with #673.

PR #669, opened 2026-08-26
fix(graph): redact all probe observations

### LINKED ISSUE #666: command probes publish the operator's home path into close receipts
Found while fixing #662 (review finding B2), deliberately scoped out of that branch.

## Problem

`describeCommand` (`src/work-graph-probes.ts`, ~line 282) builds a `command` probe's `observed` text by joining the subprocess's stdout and stderr through `boundObserved`, with **no home-path redaction**:

```ts
const tail = boundObserved([outcome.stdout, outcome.stderr].filter(p => p.trim().length > 0).join("\n"), limit);
return tail.length === 0 ? `exit ${outcome.exitCode}` : `exit ${outcome.exitCode}: ${tail}`;
```

`observed` is published into a close receipt, which lands as a comment on a tracker whose visibility soma cannot know. Any command whose output names an absolute path under `$HOME` therefore publishes the operator's account name. Measured example shape:

```
exit 1: ls: /Users/<account>/.tmp-nonexistent-dir: No such file or directory (expected exit 0)
```

This predates #662; that branch fixed the equivalent hole on the **git** probe path only.

## Why it is arguably worse than the git case

A `command` probe runs an adopter-declared shell command whose output is unbounded in shape — unlike git, whose error vocabulary is small and predictable. Anything the declared command prints goes to the tracker.

## The fix is now one call

#662 added `redactHome` in `src/work-graph.ts` (a sibling of `collapseHome`, which is prefix-only and therefore inert on a path embedded mid-sentence — see that function's docblock for the measured counterexample). Applying it to the `describeCommand` tail should be a one-line change plus tests.

## Scope note

A sweep during #662 found exactly two places subprocess output reaches `observed`: `describeCommand` (this issue) and `gitUnreachable` (fixed in #662). Lines that pass `stdout` through to extract a sha do not publish it.

## Acceptance

- [ ] A `command` probe whose output names a path under `$HOME` publishes it home-collapsed.
- [ ] A real-subprocess test asserts the raw home prefix is absent from the published string (not a stub — a stubbed outcome cannot catch this class).
- [ ] The sweep is re-run and any further publication sites are covered or recorded.

## Related, worth deciding first

#662's implementer raised a design question this issue is a symptom of: every probe branch currently assembles its own `observed` string and re-decides redaction, and the same defect has now recurred twice at that seam. A single "publish this" chokepoint that every branch routes through would make the question unanswerable-per-branch. Larger than this fix; worth deciding before patching site three.

jcfischer:
Fixes #666.

## Summary
- Redact every `ProbeResult.observed` at its result-construction boundary.
- Redact command output before tail bounding, so truncation cannot split a home path.
- Cover real command output plus absent/invalid probe-registry refusals.

## Verification
- `bun test` (2547 pass)
- `bun run typecheck`
- `bunx eslint src/work-graph-probes.ts test/work-graph-probes.test.ts`

The repository-wide lint command currently reports unrelated pre-existing errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
+## [0.18.2] - 2026-08-26
+
+### Fixed
+
+- **DSH graph closure is available through a bounded host tool.** `soma_graph`
+  now exposes `close` with resolution prose or a workspace-relative resolution
+  file; traversal and symlink escapes are refused before the existing graph
+  closure gate runs. The tool also surfaces CLI stderr when a graph command
+  fails, while `decisions --write` remains limited to the derived map index.
+- **Portable-skill reconciliation preserves principal-replaced links.** A stale
+  manifest may remove a migrated loader slot only when that slot targets its
+  exact matching `~/.soma/skills/<name>` registry entry; links to other
+  principal-owned locations, including another registry skill, remain intact.
 
 ## [0.18.1] - 2026-08-25
 
--- a/arc-manifest.yaml
+++ b/arc-manifest.yaml
@@ -1,6 +1,6 @@
 schema: arc/v1
 name: soma
-version: 0.18.1
+version: 0.18.2
 type: tool
 tier: community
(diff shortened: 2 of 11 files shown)


jcfischer:
Superseded by #670, recreated from current `main` so this PR contains only the #666 change.

jcfischer:
Closing in favor of #670.

PR #670, opened 2026-08-26
fix(graph): redact all probe observations

### LINKED ISSUE #666: command probes publish the operator's home path into close receipts

jcfischer:
Fixes #666.

## Summary
- Route every `ProbeResult.observed` through a home-path redaction boundary.
- Redact command text before tail truncation, preventing a truncated home path from leaking.
- Add focused coverage for command output and probe-registry refusal text.

## Verification
- `bun test test/work-graph-probes.test.ts` (43 focused tests)
- `bun run typecheck`
- `git diff --check`

### DIFF
--- a/src/work-graph-probes.ts
+++ b/src/work-graph-probes.ts
@@ -245,6 +245,11 @@ function resolveDeps(options: ProbeRunnerOptions): ProbeRunnerDeps {
   };
 }
 
+/** The sole publication boundary for probe observations: close receipts use this value verbatim. */
+function publishObserved(observed: string): string {
+  return redactHome(observed);
+}
+
 function probed(
   probe: Probe,
   outcome: "pass" | "fail",
@@ -252,7 +257,7 @@ function probed(
   at: string,
   cwd?: string,
 ): ProbeResult {
-  return { probe, state: "probed", outcome, observed, at, ...(cwd === undefined ? {} : { cwd }) };
+  return { probe, state: "probed", outcome, observed: publishObserved(observed), at, ...(cwd === undefined ? {} : { cwd }) };
 }
 
 /**
@@ -279,7 +284,9 @@ function describeCommand(
       : `timed out after ${timeoutSec}s (killed)`;
   }
   const limit = passed ? OBSERVED_PASS_TAIL_LIMIT : OBSERVED_TAIL_LIMIT;
-  const tail = boundObserved([outcome.stdout, outcome.stderr].filter((part) => part.trim().length > 0).join("\n"), limit);
+  // Redact before bounding: a tail cut inside a home path would otherwise sever
+  // its prefix and make the later publication chokepoint unable to recognise it.
+  const tail = boundObserved(redactHome([outcome.stdout, outcome.stderr].filter((part) => part.trim().length > 0).join("\n")), limit);
   return tail.length === 0 ? `exit ${outcome.exitCode}` : `exit ${outcome.exitCode}: ${tail}`;
 }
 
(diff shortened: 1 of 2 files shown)


src/work-graph-probes.ts
```diff
@@ -245,14 +245,19 @@ function resolveDeps(options: ProbeRunnerOptions): ProbeRunnerDeps {
   };
 }
 
+/** The sole publication boundary for probe observations: close receipts use this value verbatim. */
```
jcfischer, line 248: **[suggestion] `publishObserved` is not the sole redaction boundary**

The function is named and commented as "The sole publication boundary for probe observations" and the `finish` comment says "`probed` is the sole publication chokepoint," yet `describeCommand` (line ~287) calls `redactHome` directly: `const tail = boundObserved(redactHome(...), limit)`. If `publishObserved` later gains logging, metrics, or a different sanitizer, `describeCommand` silently misses it.

Fix: Route `describeCommand`'s redaction through `publishObserved` and move the call before `boundObserved`. Since `redactHome` is idempotent on already-redacted paths, the double-redact from `probed` → `publishObserved` is harmless.

jcfischer (commented):

1 finding(s): 1 suggestion.

Round impact: behavior 1, checks 0, prose 0; unclassified impact 0; prior-round surface 0; restated from earlier rounds 0.


### Maintainability

- **[suggestion]** `src/work-graph-probes.ts:248` — **`publishObserved` is not the sole redaction boundary**
  The function is named and commented as "The sole publication boundary for probe observations" and the `finish` comment says "`probed` is the sole publication chokepoint," yet `describeCommand` (line ~287) calls `redactHome` directly: `const tail = boundObserved(redactHome(...), limit)`. If `publishObserved` later gains logging, metrics, or a different sanitizer, `describeCommand` silently misses it.
  Fix: Route `describeCommand`'s redaction through `publishObserved` and move the call before `boundObserved`. Since `redactHome` is idempotent on already-redacted paths, the double-redact from `probed` → `publishObserved` is harmless.


---

jcfischer:
Addressed the Sage suggestion in 20f76f5: command output now uses `publishObserved` before tail bounding, and the result constructor reuses the same idempotent sanitizer.

jcfischer (commented):

No findings. Sage approves.

Round impact: behavior 0, checks 0, prose 0; unclassified impact 0; prior-round surface 0; restated from earlier rounds 0.



---

Issue #543, opened 2026-08-06
Projection duplication: renderAssistantCore reaches the model twice per turn


jcfischer:
The core identity block is rendered into two projected files, and the substrate hook loads both into the same prompt. Every turn carries it twice, byte-identical.

Measured on pi.dev with a logging proxy in front of the model gateway (40 real requests captured):

- four sections duplicated verbatim
- **2,143 B, about 595 tokens, repeated per request**
- the profile file is 2,400 B total, of which 2,143 B is identical to the context file: 89% redundant, 257 B unique

## Root cause

`renderAssistantCore(input)` is called more than once per adapter and both outputs reach the model:

| adapter | calls |
|---|---|
| `src/adapters/claude-code.ts` | 2 |
| `src/adapters/cursor.ts` | 2 |
| `src/adapters/anthropic-cowork.ts` | 2 |
| `src/adapters/pi-dev/adapter.ts` | 3 |
| `src/adapters/codex/adapter.ts` | 3 |

For pi-dev, `adapter.ts:22` renders it into `instructions` (the context file) and `adapter.ts:574` renders it again into the profile file. The projected extension then concatenates both on every turn in `before_agent_start`, reading `profile.md` and `context.md` and interpolating both into one prompt string.

This is not a stale-file or reinstall artefact. It is the designed data flow.

## Scope

Not pi-specific: the claude-code projection has the same overlap between its context and profile files, and everything under `rules/` is auto-discovered, so that substrate pays for it too. Five adapters share the pattern.

## Why fix it beyond the token count

595 tokens is small next to a 42,000-token prompt, so cost alone is a weak argument. Two better reasons:

1. **Two sources of truth in one prompt.** If a projection is ever partially rewritten (soma doctor repairing one file, a hand edit, version skew between installs) the model receives two versions of the same content with no rule for which wins. The provenance header distinguishes managed from hand-replaced, but not managed-and-inconsistent.
2. **It shows the split has no reader.** Adapter comments say to use the profile file for fast orientation, but nothing reads it *instead of* the context. The hook loads both unconditionally, so the fast path is never taken and the file only adds weight.

## Suggested fix

1. **Stop concatenating both.** Inject the context file only in the hot path; leave the profile as an on-demand read. It is already reachable via `soma_context action=profile` and the `/soma` command, both of which read it explicitly.
2. **Make the profile file a pointer rather than a copy** — its unique 257 B plus a reference to the context file.
[...]

PR #673, opened 2026-08-29
fix(claude-code): name a guard failure as a guard failure, and pin the #640 symptom

### LINKED ISSUE #640: Policy guard loads the soma worktree, so editing soma can brick the session that edits it

jcfischer:
Closes #640. Supersedes #668.

## The fix already landed — this is what it left behind

#657 stopped the policy guard from loading the soma working tree. Verified against a scratch install off `01f05ce`:

```json
"trustedSomaRepo": "<somaHome>/runtime/claude-code/current"
```

which resolves to `runtime/artifacts/33e37e11…` and carries its own `src/cli.ts`. The guard runs entirely inside a frozen snapshot; `install.ts:159` says it outright — *"Guarded substrates execute only the immutable artifact, never this editable checkout."*

So #668 (a second, claude-code-only pinning mechanism) is closed rather than rebased. Two gaps survived it.

## 1. A guard that cannot run said so in the same register as a rule that fired

Denials from a broken guard read:

```
Soma runtime policy inspection failed closed: error: Cannot find module
'./does-not-exist' from '…/src/cli.ts'  Bun v1.3.6 (macOS arm64)
. Reinstall the active enforcement artifact with soma install claude-code --apply…
```

against a real policy denial:

```
Runtime policy denied this action: credential-file-egress.
```

Both deny — fail-closed is not relaxed anywhere in this PR — but only one is a judgement about the action, and only the other has a recovery. Every non-policy path (unreadable config, malformed hook input, non-zero exit, unparseable output) now carries:

```
Soma policy guard UNAVAILABLE (fail-closed — this is not a policy denial): …
Recover with: soma install claude-code --apply, or soma runtime rollback --substrate claude-code.
```

A rule that fired is untouched and still carries the rule's own reason.

## 2. Nothing tested the symptom

`runtime-artifact.test.ts` covers the mechanism well — 13 tests over hashing, sealing, activation, rollback, pruning. None of them puts a **guarded tool call against a broken source checkout**, which is what the issue was about. The regression could return under a fully green suite.

`test/claude-code-policy-guard-runtime.test.ts` (5 tests) asserts both halves side by side:

| test | asserts |
|---|---|
| install pins at the artifact | `trustedSomaRepo` is the substrate pointer, not `REPO_ROOT`, and holds its own `src/cli.ts` |
| **a broken checkout no longer blocks** | install from a copied source root, break it, `ls -la` still **allows** |
| a policy denial still denies | `credential-file-egress` denies, and is *not* dressed as UNAVAILABLE |
| an unrunnable guard | denies with the class + names the recovery |
| **the original failure reproduces** | pre-artifact config shape → `ls -la` **denies** |

The last two rows are the point: the same input, allow on the pinned path and deny on the unpinned one.

## Verification

- The unavailable-wording test **fails against main's message** and passes with this change (checked by reverting the hook and re-running) — it isn't a tautology.
- Full suite: **2575 pass, 0 fail** (160 files).
- No other test or source depended on the old wording.

## Left open deliberately

The codex and grok hook entries carry the same undifferentiated wording (`codex-hook-entry.mjs:315,374`, `grok-hook-entry.mjs:361`). Same one-line shape, but a separate substrate surface and a separate change.

https://claude.ai/code/session_017HHB5DjnCNUWVt35AiBRj2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   asserted the behaviour that motivated it: a guarded tool call against a broken
   source checkout. `test/claude-code-policy-guard-runtime.test.ts` now pins the
   allow (artifact) and deny (unpinned config) paths side by side. ([#640])
+- **A guard config that parses to a non-object fails closed instead of
+  crashing.** `JSON.parse` succeeds on every JSON scalar, so a config file of
+  literal `null` reached the field check and threw on `config.error` before any
+  denial was emitted — the one non-policy path the wording above did not in fact
+  cover. `readConfig` now rejects any parse result that is not an object. ([#640])
 
 ## [0.19.1] - 2026-09-05
 
@@ -102,6 +107,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   inspection. ([#658])
 
 [#315]: https://github.com/the-metafactory/soma/issues/315
+[#543]: https://github.com/the-metafactory/soma/issues/543
 [#640]: https://github.com/the-metafactory/soma/issues/640
 [#654]: https://github.com/the-metafactory/soma/issues/654
 [#658]: https://github.com/the-metafactory/soma/issues/658
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **A guard that cannot run says so, distinctly from a rule that fired.** Every
+  non-policy failure path in the Claude Code policy guard (unreadable config,
+  malformed hook input, non-zero inspection exit, unparseable output) now denies
+  with `Soma policy guard UNAVAILABLE (fail-closed — this is not a policy
+  denial)` and names its recovery. A rule that fired still carries the rule's own
+  reason. Both still deny; fail-closed is unchanged. ([#640])
+- **The soma#640 symptom is covered by regression tests.** `soma install` has
+  pinned guarded substrates to a frozen artifact since #657, but nothing
+  asserted the behaviour that motivated it: a guarded tool call against a broken
+  source checkout. `test/claude-code-policy-guard-runtime.test.ts` now pins the
+  allow (artifact) and deny (unpinned config) paths side by side. ([#640])
+
 ## [0.19.1] - 2026-09-05
 
 ### Fixed
@@ -88,6 +102,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   inspection. ([#658])
 
 [#315]: https://github.com/the-metafactory/soma/issues/315
+[#640]: https://github.com/the-metafactory/soma/issues/640
 [#654]: https://github.com/the-metafactory/soma/issues/654
 [#658]: https://github.com/the-metafactory/soma/issues/658
 [#659]: https://github.com/the-metafactory/soma/issues/659
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   asserted the behaviour that motivated it: a guarded tool call against a broken
   source checkout. `test/claude-code-policy-guard-runtime.test.ts` now pins the
   allow (artifact) and deny (unpinned config) paths side by side. ([#640])
-- **A guard config that parses to a non-object fails closed instead of
-  crashing.** `JSON.parse` succeeds on every JSON scalar, so a config file of
-  literal `null` reached the field check and threw on `config.error` before any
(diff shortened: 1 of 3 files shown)


jcfischer:
## Review + rebase pass

Rebased onto `main` (the PR was `CONFLICTING` since 0.19.1 landed — the `CHANGELOG` entries now sit under `[Unreleased]`, with main's 0.19.1/#543 section preserved below).

Sage, offline, `--lens-concurrency 3 --timeout 300`, three rounds. Both findings came from HonestOracle and both were against this PR's own CHANGELOG claim that "every non-policy failure path" denies with UNAVAILABLE. Both were real, and both were fixed by making the claim true rather than by weakening it:

1. **A config parsing to literal `null`** — `JSON.parse` succeeds on every JSON scalar, so the parse-error branch never saw it, and reading `config.error` off `null` threw before any decision was emitted. Probed: `null` throws; `[]`, `"str"`, `0`, `false`, `{bad` all already reached the field check.
2. **`bunPath: ""`** — satisfies `typeof x === "string"`, then throws inside Bun's spawn. Probed against the real hook, not a model of it: `""` exits 1 with empty stdout, while `"   "` and a nonexistent path both deny correctly.

Patching the second the way the first was patched would have invited a third, so `main` now picks the deny surface first and wraps the rest: any unexpected throw becomes an UNAVAILABLE denial instead of a stack trace. Enumerated paths keep their specific messages, and fail-closed is unchanged throughout.

Both regression tests were checked against the unfixed hook — they exit 1 there and 0 here, so neither is a tautology.

Full suite: **2584 pass, 0 fail** across 162 files. Sage round 3: approved, no findings.

Also defines the dangling `[#543]` link reference introduced by 0.19.1. Nineteen older refs are still dangling in historical release blocks and are left for a separate pass.

